PR #42, opened 2023-02-11
tRPCOptions extends `FetchHandlerRequestOptions`-ish

AlexErrant:
This allows the use of options like `responseMeta` or `createContext`.

### DIFF
--- a/packages/trpc-server/src/index.ts
+++ b/packages/trpc-server/src/index.ts
@@ -1,18 +1,20 @@
 import type { AnyRouter } from '@trpc/server'
+import type { FetchHandlerRequestOptions} from '@trpc/server/adapters/fetch'
 import { fetchRequestHandler } from '@trpc/server/adapters/fetch'
 import type { MiddlewareHandler } from 'hono'
 
-type tRPCOptions = {
-  endpoint?: string
-  router: AnyRouter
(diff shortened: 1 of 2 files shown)


packages/trpc-server/src/index.ts
```diff
@@ -1,18 +1,20 @@
...
 
-type tRPCOptions = {
-  endpoint?: string
-  router: AnyRouter
-}
+type tRPCOptions =
+  Omit<
+    FetchHandlerRequestOptions<AnyRouter>,
+    'req' | 'endpoint'
+  > & Partial<Pick<FetchHandlerRequestOptions<AnyRouter>, 'endpoint'>>
```
AlexErrant, line 10: Please note that this is babby's first utility-Typescript. I have no idea if this is good or not - I just tried as hard as I could to rely on pieces of `FetchHandlerRequestOptions`. This way, if `endpoint` or something should change names/types in the future, this will break/pass the type through as expected.

yusukebe:
Hi @AlexErrant !

Thank you for the PR. We want to accept it. Could you create the changeset?

```
yarn changeset
```

and git add & commit & push.

yusukebe:
Thanks! I'll merge it.

PR #52, opened 2023-02-19
Auth.js Middleware

ivoilic:
Hi!

This PR adds middleware for a [Auth.js](https://authjs.dev/) Server, it's still a WIP.

There's an issue with Jest and the @auth/core imports that's making it impossible to test currently. @yusukebe if you have suggestions please let me know.


yusukebe:
OK, I'll check it later!

yusukebe:
Hi @ivoilic 

This problem seems to be the Auth.js matter because the minimum project following will fail.

* https://github.com/yusukebe/jest-authjs

SS

balazsorban44:
This is not an Auth.js problem, but a config issue, see the solution explained here: https://github.com/nextauthjs/next-auth/issues/6822#issuecomment-1447347641

Here is the actual sub-module being published correctly: https://unpkg.com/browse/@auth/core@0.5.0/providers/github.js, https://unpkg.com/browse/@auth/core@0.5.0/package.json

yusukebe:
Hi @balazsorban44 !

Thanks for your comment.

> This is not an Auth.js problem, but a config issue

I'm sorry I doubted Auth.js!

@ivoilic 

Could you check it?

ivoilic:
@yusukebe I tried to follow @balazsorban44's (Thanks so much for taking the time to help with this!) example and made changes at the package level. This resolved the current issue but now I'm getting the error: `ReferenceError: exports is not defined`. I also tried making these changes at the root of the repo and was getting the same error with other imported packages. I'm not familiar enough with resolving these sort of packaging problems to make meaningful progress on this currently.

balazsorban44:
`ReferenceError: exports is not defined` means you have CJS code that is being interpreted as ESM, since the CommonJS globals `exports` or `module` are not available in ES modules.

yusukebe:
Hi @ivoilic !

To avoid the error, it would be good not to use `jest-environment-miniflare`. So, set `jest.config.js` as follows according to @balazsorban44's advice.

```ts
module.exports = {
  testMatch: ['**/test/**/*.+(ts|tsx|js)'],
  transform: {
    '^.+\\.(ts|tsx)$': ['ts-jest', { useESM: true }],
  },
  extensionsToTreatAsEsm: ['.ts'],
}
```

Then we cannot use `crypto`, so set it to global in the test code.

```ts
import crypto from 'crypto'

// ...

globalThis.crypto = crypto
```

And this may work.

We can't test without a GitHub token, so I would like to know if there is another good way to test that doesn't use third-party tokens.

Fnux8890:
does any one have a problem with 
```bash
node:internal/modules/esm/resolve:303
  return new ERR_PACKAGE_PATH_NOT_EXPORTED(
         ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/bruger/Documents/Github/Semester_project_trustworthy/hub_server/node_modules/@auth/core/package.json
```

neolectron:
Hello everyone, can I take over this branch and start a new PR or is @ivoilic still available to have a look ?

yusukebe:
We already have https://github.com/honojs/middleware/tree/main/packages/auth-js

PR #89, opened 2023-06-01
Sentry middleware: fix undefined DSN environment variable error

mllnd:
Sentry middleware currently expects that `SENTRY_DSN` or `NEXT_PUBLIC_SENTRY_DSN` are available in the request context. This causes the middleware to throw an undefined property error when given variables are not set.

This PR also bumps the Hono version due to changes in `MiddlewareHandler` type definition.

### DIFF
--- a/.changeset/kind-parents-play.md
+++ b/.changeset/kind-parents-play.md
@@ -1,0 +1,5 @@
+---
+'@hono/sentry': patch
+---
+
+Use optional chaining for determining Sentry DSN from environment (#89)
--- a/packages/sentry/deno_dist/index.ts
+++ b/packages/sentry/deno_dist/index.ts
@@ -41,7 +41,7 @@ export const sentry = (
       hasExecutionContext = false
     }
     const sentry = new Toucan({
-      dsn: c.env.SENTRY_DSN ?? c.env.NEXT_PUBLIC_SENTRY_DSN,
+      dsn: c.env?.SENTRY_DSN ?? c.env?.NEXT_PUBLIC_SENTRY_DSN,
       allowedHeaders: ['user-agent'],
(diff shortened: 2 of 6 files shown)


yusukebe:
Hi @mllnd !

Thank you for creating this PR. Looks good!

Could you do `yarn changeset` and edit&add the changese? This will be patch change.

mllnd:
Hey @yusukebe, thanks for the quick reply. `changeset` file is now present, as requested.

yusukebe:
@mllnd 

Great! Just one more thing. Could you run `denoify` in the sentry directory? It's necessary to generate files for Deno.

mllnd:
@yusukebe Should be all good now. 

Quick question: what would be the best way of using the Denoify'ed version of this middleware in Deno projects? The official documentation suggests we use the `npm:@hono/sentry` syntax for pulling it in or is there a better way?

yusukebe:
@mllnd 

Thanks!

Ah, currently, this Sentry middleware is hosted in this monorepo. Since some middleware are in a monorepo, it's difficult to distribute them from `deno.land/x`. This is because one GitHub repo can distribute only one module with `deno.land/x`. So, the only way to use Sentry middleware or others is to use the `npm:` specifier. I'm planning to avoid this and distribute each middleware from `deno.land/x`, but it will be a big challenge. 

The reason I want to support `deno.land/x` is because I want to use them on Deno Deploy. But, Deno Deploy will likely support the `npm:` specifier in the future (or maybe it already does?).

So if you're not running it on Deno Deploy, you can use the `npm:` specifier. But if you want to use it on Deno Deploy, you'll have to wait until it supports the `npm:` specifier.

PR #127, opened 2023-08-20
feat(qwik-city): adapt to v1.2.0 API and refine dependencies

shuymn:
Hi everyone,

The main focus of this PR is to adapt to the new `ServerRequestEvent` signature introduced in [qwik-city v1.2.0](https://github.com/BuilderIO/qwik/releases/tag/v1.2.0) by  https://github.com/BuilderIO/qwik/pull/4433

v1.1.5: https://github.com/BuilderIO/qwik/blob/v1.1.5/packages/qwik-city/middleware/request-handler/types.ts#L16-L24C2
v1.2.0: https://github.com/BuilderIO/qwik/blob/v1.2.0/packages/qwik-city/middleware/request-handler/types.ts#L31

To achieve this, the peerDependencies have been updated to target qwik-city v1.2.0 or higher. 

Additionally, the Hono version was previously pinned at 3.1.5, which was unnecessarily restrictive and forced middleware users to stick with that specific version. To provide more flexibility, the requirement has been changed to Hono v3.1.5 or higher.

I've also done some minor refactoring inspired by the implementation in [qwik-city's middleware](https://github.com/BuilderIO/qwik/blob/09e55521deb9cab5e954b7cbf55b3e26d49de799/packages/qwik-city/middleware/cloudflare-pages/index.ts). 

Note: The newly introduced `getClientConn` function may need to address use cases where the implementation varies depending on the deployment environment, such as whether it's running on Cloudflare Workers or Deno. This middleware doesn't currently address these scenarios, I think it may require some additional planning and consideration.

Thank you for developing this wonderful software; I'm looking forward to contributing more in the future.

### DIFF
--- a/.changeset/thirty-geese-live.md
+++ b/.changeset/thirty-geese-live.md
@@ -1,0 +1,5 @@
+---
+'@hono/qwik-city': patch
+---
+
+feat(qwik-city): adapt to qwik-city changes in v1.2.0
--- a/packages/qwik-city/package.json
+++ b/packages/qwik-city/package.json
@@ -29,12 +29,12 @@
   "peerDependencies": {
     "@builder.io/qwik": "^1.0.0",
     "@builder.io/qwik-city": "^1.0.0",
-    "hono": "3.1.5"
+    "hono": "^3.1.5"
   },
   "devDependencies": {
     "@builder.io/qwik": "^1.0.0",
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @shuymn!

Thanks for the PR. It looks so good that I have nothing more to add. Let's ship it!

PR #135, opened 2023-08-23
refactor(valibot-validator): update to latest valibot version

### LINKED ISSUE #110: We will change the `safeParse` API of Valibot
This is just a hint that we will change the `safeParse` API. `.error` will be deprecated and `.issues` will be added.

- https://github.com/fabian-hiller/valibot/blob/main/library/src/methods/safeParse/safeParse.ts
- https://github.com/honojs/middleware/blob/main/packages/valibot-validator/src/index.ts#L29

ZerNico:
closes #110 

Uses the exported `safeParse` result type.

### DIFF
--- a/.changeset/fuzzy-spoons-crash.md
+++ b/.changeset/fuzzy-spoons-crash.md
@@ -1,0 +1,5 @@
+---
+'@hono/valibot-validator': minor
+---
+
+update valibot version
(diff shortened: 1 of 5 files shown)


yusukebe:
@ZerNico 

Cool! Thanks for a quick response.

PR #151, opened 2023-09-06
Clerk Middleware

octoper:
This PR adds a middleware that will help easily integrate user authentication and user management by [Clerk](https://clerk.com) to your Hono application.

### DIFF
--- a/packages/clerk-auth/README.md
+++ b/packages/clerk-auth/README.md
@@ -7,7 +7,7 @@ This middleware can be used to inject the active Clerk session into the request 
 ## Installation
 
 ```plain
-npm i hono @hono/clerk-auth
+npm i hono @hono/clerk-auth @clerk/backend
 ```
 
 ## Configuration
--- a/packages/clerk-auth/README.md
+++ b/packages/clerk-auth/README.md
@@ -1,0 +1,83 @@
+# Clerk middleware for Hono
+
+This is a [Clerk](https://clerk.com) third-party middleware for [Hono](https://github.com/honojs/hono).
+
+This middleware can be used to inject the active Clerk session into the request context.
+
+## Installation
+
+```plain
+npm i hono @hono/clerk-auth
+```
+
+## Configuration
+
+Before starting using the middleware you must set the following enviroment variables:
+
+```plain
+CLERK_SECRET_KEY=<You-secret-key>
+CLERK_PUBLISHABLE_KEY=<Your-publishable-key>
+```
+
+## How to Use
+
+```ts
+import { clerkMiddleware, getAuth } from '@hono/clerk-auth'
+import { Hono } from 'hono'
+
+const app = new Hono()
+
+app.use('*', clerkMiddleware())
+app.get('/', (c) => {
+  const auth = getAuth(c)
+
+  if (!auth?.userId) {
+    return c.json({
+      message: 'You are not logged in.'
+    })
+  }
+
+  return c.json({
+    message: 'You are logged in!',
+    userId: auth.userId
+  })
+})
+
+export default app
+```
+
+## Accessing instance of Backend API client
+
+```ts
+import { clerkMiddleware, getAuth } from '@hono/clerk-auth'
+import { Hono } from 'hono'
+
+const app = new Hono()
+
+app.use('*', clerkMiddleware())
+app.get('/', async (c) => {
+  const clerk = c.get('clerk')
+
+  try {
+    const user = await clerkClient.users.getUser('user_id_....')
+
+    return c.json({
+      user,
+    })
+  } catch (e) {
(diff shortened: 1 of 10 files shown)


packages/clerk-auth/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Installation
+
+```plain
+npm i hono @hono/clerk-auth
+```
+
+## Configuration
+
+Before starting using the middleware you must set the following enviroment variables:
```
yusukebe, line 15: `enviroment` => `environment`

packages/clerk-auth/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/honojs/middleware.git"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "devDependencies": {
+    "hono": "^2.7.2"
```
yusukebe, line 31: You have to use the latest version of Hono.

packages/clerk-auth/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "dependencies": {
+    "@clerk/backend": "^0.28.1"
```
yusukebe, line 34: Perhaps, it might be better to make `@clerk/backend` as `peerDependencies`. And fix the install command as the following:

```plain
npm i hono @hono/clerk-auth @clerk/backend
```

yusukebe:
Hi @octoper !

This, supporting Clerk auth awesome! I've leaved some comments and I'd like you to do the following two things:

* Writing tests - It seems to be difficult because it's treating the external service, but you have to write at least a minimal test.
* Adding an example - It takes time to understand how to use this middleware. If users can see an example project, it will be a big help.

Anyway, great work! Thanks.

octoper:
Hi @yusukebe!

Thanks for the feedback! I will try to implement some tests and do the changes you asked!

octoper:
I have implemented some tests and I will try to have an example repo in the next few days.

yusukebe:
Hi @octoper,

Awesome work!

One thing I'd like you to do in this PR is to create the CI script as follows:

https://github.com/honojs/middleware/tree/main/.github/workflows

Once this is done, we can merge this into the main branch and ship it!

octoper:
Hi @yusukebe,
It's ready!

packages/clerk-auth/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const getAuth = (c: Context) => {
+  return c.get('clerkAuth')
+}
+
+export const clerkMiddleware = (options?: ClerkOptions): MiddlewareHandler => {
+  return async (c, next) => {
+    const { secretKey, publishableKey, apiUrl, apiVersion, ...rest } = options || {
+      secretKey: c.env.CLERK_SECRET_KEY || '',
```
yusukebe, line 23: I realized that we need to use the `env` helper to support other runtimes, not only Cloudflare:

https://hono.dev/helpers/adapter#env

We can write it as follows:

```ts
    const {
      CLERK_PUBLISHABLE_KEY,
      CLERK_SECRET_KEY,
      CLERK_API_URL,
      CLERK_API_VERSION,
      CLERK_FRONTEND_API,
    } = env<Env>(c)
```

yusukebe (commented):
Hi @octoper!

Thanks! It's almost done, but two things remain. Please check the `env` thing, and could you run the `prettier` command to prettify the code:

```text
npx prettier --write ./src
```

yusukebe (commented):
Hi @octoper!

Thanks! It's almost done, but two things remain. Please check the `env` thing, and could you run the `prettier` command to prettify the code:

```text
npx prettier --write ./src
```

yusukebe:
Hi @octoper !

I've updated the `env` helper thing by myself. Then all done!

Awesome work! Thanks a lot!

kevbook:
Would also love one for auth0 if possible!

packages/clerk-auth/package.json
nikosdouvlis: Hello @yusukebe :) Was going through this PR and I'd like to ask if there were any specific reasons you prefer `npm i hono @hono/clerk-auth @clerk/backend` instead of `npm i hono @hono/clerk-auth` and making `@clerk/backend` a direct dependency?

Please feel free to ignore my question completely as I'm only asking out of curiosity. Thank you!

yusukebe: Hi @nikosdouvlis 

I didn't have any significant reasons. This is better for importing `@clerk/backend` in other places but does not seem necessary.

PR #153, opened 2023-09-09
[zod-openapi] Merge subapps' definitions into main app

### LINKED ISSUE #152: [Zod OpenAPI] accept args in the `constructor`
It should receive args like a `strict: false`.

mikestopcontinues:
Three features are included in this PR:

1. You can pass init options to `OpenAPIHono` (fixes #152)
2. You can generate OpenAPI v3.1 spec
3. You can merge the spec definitions of nested `OpenAPIHono` routers

If you would prefer to handle these separately, they're in clean, distinct commits. 

New tests cover all new functionality. No old tests changed.

### DIFF
--- a/.changeset/quiet-berries-accept.md
+++ b/.changeset/quiet-berries-accept.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+Merge subapps' spec definitions into main app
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @mikestopcontinues !

Great work! This is what we want. I've reviewed the all code, but there is no problem. Let's ship it! Thanks!

PR #158, opened 2023-09-14
Add Typia validator

dworznik:
This is a validator using [Typia](https://github.com/samchon/typia). The code is based on `zod-validator`.


yusukebe:
Hi @dworznik !

Thank you! Looks good. One thing. Could you add a CI config? Like `zod-validator`:

https://github.com/honojs/middleware/blob/main/.github/workflows/ci-zod-validator.yml

yusukebe:
Thanks! Let's go with this.

PR #167, opened 2023-09-20
Zod-Openapi-Middleware: Export types to allow handlers and hooks to be declared in more places.

exsjabe:
Exported intermediate types representing the handler and the hook for a route so that handlers can be declared outside of the register method.

Like such

```ts
const route = createRoute({
  method: 'get',
  path: '/users/{id}',
  request: {
    params: ParamsSchema,
  },
  responses: {
    200: {
      content: {
        'application/json': {
          schema: UserSchema,
        },
      },
      description: 'Retrieve the user',
    },
  },
})

const handler: RouteHandler<typeof route> = (c) => {
 const { id } = c.req.valid('param')
  return c.jsonT({
    id,
    age: 20,
    name: 'Ultra-man',
  })
}

app.openapi(route, handler)

```

### DIFF
--- a/.changeset/rare-pillows-change.md
+++ b/.changeset/rare-pillows-change.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+Export types that allow for separate declaratins of route handlers and hooks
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @exsjabe 

This is a great improvement! Could you do `yarn changeset` ? It should be a "minor version" up. Thanks.

exsjabe:
Sure I'll get on it tomorrow 😁

yusukebe:
@exsjabe 

Thanks! I'll merge it now!

Issue #95, opened 2023-06-18
Swagger middleware


vincent-vade:
Hi,

Hone is awesome but is it possible to add swagger middleware ? 

Best regards 

Vincent

yusukebe:
Hi @vincent-vade!

Yes, creating a middleware for Swagger sounds like a good idea. Ideally, we would create an OpenAPI middleware, but we're still in the planning phase since we haven't started the API design yet. However, if you're talking about implementing just the Swagger UI, it might be simpler to accomplish.

If you're referring to the full suite of OpenAPI tools, we would need to start designing the middleware. For example, we would need to decide how to define schemas, whether it be with YAML, JSON, or a JS code base.

yusukebe:
"zod-to-openapi" looks good. I'll check it in more detail later.

ZerNico:
I think it'd be a good idea to also allow support for different validation libraries, so a modular design would be good. Maybe a general combination with the validation middlewares to just extract the schema from them and a mapper for validation lib to openapi schema

yusukebe:
@kiancyc

Is Hapi's generator this?:
https://github.com/hapi-swagger/hapi-swagger

@ZerNico

Yeah, it's great that we can support Zod, Valibot, Joi, or others. Hono's validator (though the design might not be the best) can support any validator. So, we have Zod, Valibot, and TypeBox validators. If we implement something like that, we can handle any validator, and it can generate OpenAPI docs.

yusukebe:
I've tried "[zod-to-openapi](https://github.com/asteasolutions/zod-to-openapi)" and a new feature for the validator.

Even though it's not released yet, I've made the zod validator capable of receiving the schema argument to validate response types (just types) as the third argument:

```ts
const apiRoutes = api.get('/users/:id', zValidator('param', ParamsSchema, UserSchema), (c) => {
  const { id } = c.req.valid('param');
  //
  return c.jsonT({
    id: '123',
    name: 'foo',
    age: 30,
  });
})
```

`UserSchema` will validate response types, which are rewritten in `jsonT()`. Perhaps it should validate actual values, but for now, it validates types only.

Next, I created the schema using zod-to-openapi:

```ts
import type { RouteConfig } from '@asteasolutions/zod-to-openapi'
import { extendZodWithOpenApi, OpenAPIRegistry } from '@asteasolutions/zod-to-openapi'
import { z } from 'zod'

extendZodWithOpenApi(z)

const registry = new OpenAPIRegistry()

[...]

PR #168, opened 2023-09-21
Added ability to display swagger ui

naporin0624:
Implementation to build swagger ui from OpenApiHono.

- Add swagger method to OpenApiHono
- Add SwaggerUI component
  - Create function to import css, js and html to build swagger ui from unpkg


naporin0624:
Will this solve the problem?
https://github.com/honojs/middleware/issues/95

yusukebe:
@naporin0624 

Super cooool! I love this feature.

I have a suggestion. How about making this feature an independent `swagger-ui` middleware? There will be other OpenAPI implementations besides this Zod OpenAPI. I want to use the swagger-ui for them too.

Is that possible? Or should I keep it in Zod OpenAPI? Thanks!

naporin0624:
@yusukebe 

> I want to use the swagger-ui for them too.

ok!
I want to move the SwaggerUI template to `@hono/swagger-ui`! The directory structure and package.json contents are based on other packages. Is there anything I should pay attention to?

## @hono/swagger-ui

```ts
import { swaggerUI } from "@hono/swagger-ui"

// Still thinking about it.
// I'm thinking of referring to https://github.com/scottie1984/swagger-ui-express as an existing implementation!
const swaggerOptions = {
   // accept open-api url, spec
   // accept customJs?
  // accept customCss?
}
app.use("/swagger-ui", swaggerUI(swaggerOptions))
```

## @hono/zod-openapi

I would like to add to OpenApiHono the ability to provide swaggerui as a method!

```ts
// @hono/zod-openapi depends internally on @hono/swagger-ui.
app.swagger("/swagger-ui, options)
```

yusukebe:
Hi @naporin0624 

Yes! I think the ways you said are great!

naporin0624:
@yusukebe 

I'm sorry! I am late.
@hono/swagger-ui provides swaggerUI middleware and SwaggerUI components.

@hono/zod-openapi provides only swagger method.

yarn.lock
naporin0624, line 1: I used npm by mistake and a lot of changes have been made, so I will fix it.

yusukebe:
@naporin0624 

Thanks. I'll take a look later.

yusukebe:
Hi @naporin0624 !

I have a simple question: is it possible to remove the dependency on `openapi3-ts`?

naporin0624:
@yusukebe 

I was accepting url and openapi spec as options in swagger ui and needed `openapi3-ts` for the spec type...
https://github.com/honojs/middleware/blob/cb36714f55e36f8979a5da98710b5424c21af5d6/packages/swagger-ui/src/index.tsx#L11

I tried to write it in devDependencies, but it was not added to node_modules when installing as a package.

DefinitelyTyped's react did this as well, so I used the same method.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/package.json

yusukebe:
Hi @naporin0624 

Thanks!

I've noticed that we should have a simpler API like the following:

```ts
app.get(
  '/ui',
  swaggerUI({
    url: '/doc'
  })
)
```

If we can realize this API, we don't need to include it in `@hono/zod-openapi` and we don't need to use `openapi3-ts`.

```js
window.onload = () => {
  window.ui = SwaggerUIBundle({
    url: ${url},
    dom_id: '#swagger-ui'
  })
}
```

The easiest way to implement this is to pass the `url`  directly to the `SwaggerUIBundle`. That's all we need to do to create a UI. I think it would be good to enable the `title`, `css`, `js`, etc. options you added to that.

And, creating a JSX component is a very interesting idea, but may not be necessary if that will be the entire page. In that case, it would be useful to use the [html](https://hono.dev/helpers/html) helper instead of JSX.

This is my opinion, but please consider!

naporin0624:
Thank you very much! @yusukebe

I think it is a very simple and good idea. My implementation had too many responsibilities.

I will make a small modification to swagger-ui to remove the dependency from zod-openapi, and make component small enough to provide only divs and scripts!

naporin0624:
@yusukebe
It became a very, very simple api. This one looks much more beautiful. Thank you very much.

I have prepared swaggerUI middleware and SwaggerUI component so that you can choose between these two. Indeed, `hono/html` was more suitable than `hono/jsx`.

```ts
app.get('/ui', swaggerUI({ url: '/doc' }));

export default app;.
```

or
 
````ts
app.get('/ui', (c) => {
  return c.html(html`)
    <html lang='en'>
      <head>
        <meta charset="utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1" />
        <meta name="description" content="Custom Swagger" />
        <title>Custom Swagger</title
        <script>
          // custom script
        </script>
        <style>
          /* custom style */
        </style>
      </head>
      ${SwaggerUI({ url: '/doc' })}
    </html>
  `);
});
```

packages/swagger-ui/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+// Use the middleware to serve Swagger UI at /ui
+app.get('/ui', swaggerUI({ url: '/doc' }))
+
+export default app
+```
+
+### Component Usage
+
+If you are using `hono/utils/html`, you can use the `SwaggerUI` component to render Swagger UI within your custom HTML structure. Here's an example:
```
yusukebe, line 33: `hono/utils/html` should be `hono/html`.

packages/swagger-ui/README.md
```diff
@@ -0,0 +1,81 @@
+# Swagger UI Middleware and Component for Hono
+
+This library, `@hono/swagger-ui`, provides a middleware and a component for integrating Swagger UI with Hono applications. Swagger UI is an interactive documentation interface for RESTful APIs, making it easier to understand and test API endpoints.
```
yusukebe, line 3: I think it would be better to specify that the API is compliant with the [OpenAPI Spec](https://swagger.io/specification/), not "RESTful APIs."

packages/swagger-ui/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  const asset = remoteAssets({ version: options?.ui?.version })
+
+  return html`
+    <div>
+      <div id="swagger-ui"></div>
+      ${asset.css.map((url) => html`<link rel="stylesheet" href="${url}" />`)}
+      ${asset.js.map((url) => html`<script src="${url}" crossorigin="anonymous"></script>`)}
+      <script>
+        window.onload = () => {
+          window.ui = SwaggerUIBundle({
```
yusukebe, line 24: What if we could pass options to the `SwaggerUIBundle`? For example, a user might want to customize like this:

```js
    window.onload = () => {
      window.ui = SwaggerUIBundle({
        url: '${options.url}',
        dom_id: '#swagger-ui',
        presets: [
          SwaggerUIBundle.presets.apis,
          SwaggerUIStandalonePreset
        ],
        layout: "StandaloneLayout",
      });
    };
```

You can refer to https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation/.

yusukebe:
Hi @naporin0624 

Good API! I like this.

I leaved some comments. Please check them!

packages/swagger-ui/README.md
naporin0624: Oh, sorry. I omitted a correction.

packages/swagger-ui/src/index.ts
naporin0624: thx!
I'll give it a try!

naporin0624:
@yusukebe

Sorry for stopping this PR.
Work has been very busy and I don't think I will be able to start this PR this month .....

yusukebe:
@naporin0624 

No need to hurry! If I have the time, I might be able to help.

yusukebe:
Hi @naporin0624 and @sor4chi!

If @naporin0624 is OK with it, @sor4chi, can you try to complete the remaining tasks? Like [this](https://github.com/honojs/middleware/pull/168#discussion_r1337807246). I know @naporin0624 doesn't have time, but @sor4chi have enough time to work on your code :)

packages/swagger-ui/src/index.ts
sor4chi: Hi, @yusukebe @naporin0624

I tried some and I figured out that `SwaggerUIStandalonePreset`, `SwaggerUIBundle.presets.apis`, etc… is a function.

I think it’s difficult to apply the configuration object to string html.

Do you have any plan or thought for this?

sor4chi: I suppose it could be configured as is with a string, but I believe there is a DX-like problem.

yusukebe: Hi @sor4chi,

I thought these should be strings like this:

```ts
type SwaggerUIOptions = {
  url: string;
  ui?: ResourceOptions;
  presets?: string[];
  layout?: string;
}
```

And the implementation is like this:

```ts
const optionsStrings = Object.entries(options)
  .map(([k, v]) => {
    return `${k}: '${Array.isArray(v) ? v.join(',') : v}'`;
  })
  .join(',\n');
```

> I believe there is a DX-like problem.

I think it's OK for them to be strings because the most important thing is that users can set their own settings.

naporin0624:
@yusukebe Yes, I would like to be able to use the same information! I would like to ask if possible. Thank you 😭

packages/swagger-ui/src/index.ts
sor4chi: https://www.npmjs.com/package/@types/swagger-ui-dist?activeTab=code
Are you planning to not support other options?

yusukebe: @sor4chi 

You found it well. I think it's better to support it.

sor4chi: I tried it out, but there is actually a place to write a callback function in the options of this library.
I think it would be difficult to fully support it, and some different approach may be needed.

yusukebe: @sor4chi

I think that `operationsSorter`, `requestInterceptor`, and `responseInterceptor` can be strings. They may not be used often, so it's not a big problem to write the functions as strings.

sor4chi: alright! thank you for your opinion!

PR #170, opened 2023-09-23
feat(zod-openapi): add support for defaultHook in initializer

msutkowski:
Adds support for a `defaultHook` to avoid needing to pass the same error reformatter to every route.

If we're okay with this approach, I can add docs to the README.

Another option would be to export the types for `.openapi` and allow someone to create an `applyDefaultHook` similar to what I've shown here: https://tsplay.dev/NnMyaw

### DIFF
--- a/.changeset/witty-dolls-raise.md
+++ b/.changeset/witty-dolls-raise.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+Add defaultHook as an option for OpenAPIHono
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @msutkowski !

Great feature! I think `defaultHook` is better than `applyDefaultHook`.

I'd like to merge this into the main. Please fix the conflicts and update the README. Thanks!

msutkowski:
> Hi @msutkowski !
> 
> Great feature! I think `defaultHook` is better than `applyDefaultHook`.
> 
> I'd like to merge this into the main. Please fix the conflicts and update the README. Thanks!

@yusukebe Updated :)

yusukebe:
Coool! Ship it now!

PR #182, opened 2023-09-28
fix: properly convert openapi path type to hono

ZerNico:
This fixes the conversion from the openapi path type.
It's also recursive to convert all params instead of just one.

E.g.:
`/user/{id}/posts/{postId}/comments/{commentId}` to `/user/:id/posts/:postId/comments/:commentId`


yusukebe:
Hi @ZerNico !

Thanks. This is a good fix! Ship it.

Issue #181, opened 2023-09-28
(zod-openapi) Return type of `handler` doesn't respect the schema strictness


masnormen:
Hi, I'm having a problem making sure the return type of a handler to be as strict as the defined zod schema in `createRoute`. Consider the following code:

```ts
const app = new OpenAPIHono().openapi(
  createRoute({
    summary: 'Test route',
    method: 'post',
    path: '/test',
    responses: {
      200: {
        description: 'test',
        content: {
          'application/json': {
            schema: z
              .object({
                name: z.string(),
              })
              .strict(), // <= a supposedly strict schema for { name: string; }
          },
        },
      },
    },
  }),
  (c) => {
    return c.jsonT({
      name: 'John Doe',
    });
  }
);
```

However, it seems that `zod-openapi` doesn't do a type parse/validation on return type. The following handler which returns an extra property/key also works, and doesn't throw an error:

```diff
(c) => {
  return c.jsonT({
    name: 'John Doe',
+    age: 12,
  });
}
```

This could make the API less safe/secure. For example after fetching user data from a DB table, the developer could accidentally return the user data without omitting the password fields first, because TypeScript doesn't complain about the extra "password" field.

I think this is related to the usage of `z.infer` in the `OutputType<R>` type, which I guess can be fixed by either:

- raising a runtime error (i.e. invoking a `z.parse()` on the response) (using an extra middleware?)
- or using advanced TypeScript typechecking tricks

Is there a workaround for making sure the type is exactly the same? Or is there something I'm missing? Anyway thank you for the amazing works!

yusukebe:
Hi! @masnormen 

You are correct, Hono's Validator does not validate the response. What Zod OpenAPI does now is very loose validation. I would like to think about this issue. Thanks.

masnormen:
Hi. I've found a solution for this issue and would like to propose a PR soon. Thanks for the response!

yusukebe:
Great! Waiting!

npearson72:
@masnormen what was your solution?

dirkluijk:
With the introduction of [codecs in Zod v4.1](https://zod.dev/codecs), the absence of this feature becomes an even bigger gap. Not only would you want to parse JSON, but also encode it again.

My proposal would be to introduce an `.encodeJson()` method that accepts the output type, where Hono Zod OpenAPI encodes the response using the same schema as defined in the route.

I’d be happy to contribute to this.

@yusukebe What are your current thoughts on the direction for solving this?

[...]

PR #184, opened 2023-09-29
feat(zod-openapi): implement Zod schema validation on response

masnormen:
Hi, here is my try to implement a fix for #181.

Because this (might be) a breaking change, and not all people might want to use the feature, I implemented them behind the flag `strictStatusCode?: boolean`, `strictResponse?: boolean` so we can enable the feature when initializing `OpenAPIHono`

- `strictStatusCode`: basically we read from `RouterConfig` about the list of possible status codes, and just return an error whenever the schema doesn't contain the status code.
- `strictResponse` basically matches the status code, get that status code's schema, then do `safeParseAsync` against the data.

Honestly I still don't know the best practice in Hono on how throw and/or modify the response (and if this is a minor or major bump). I hope the tests can make it more clear though. Thank you

### DIFF
--- a/.changeset/itchy-games-fail.md
+++ b/.changeset/itchy-games-fail.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+add response Zod schema validation on response
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -189,15 +189,15 @@ export class OpenAPIHono<
 > extends Hono<E, S, BasePath> {
   openAPIRegistry: OpenAPIRegistry
   defaultHook?: OpenAPIHonoOptions<E>['defaultHook']
-  strictResponseCode?: OpenAPIHonoOptions<E>['strictStatusCode']
-  strictResponseSchema?: OpenAPIHonoOptions<E>['strictResponse']
+  strictStatusCode?: OpenAPIHonoOptions<E>['strictStatusCode']
+  strictResponse?: OpenAPIHonoOptions<E>['strictResponse']
 
   constructor(init?: HonoInit<E>) {
     super(init)
     this.openAPIRegistry = new OpenAPIRegistry()
     this.defaultHook = init?.defaultHook
-    this.strictResponseCode = init?.strictStatusCode
-    this.strictResponseSchema = init?.strictResponse
+    this.strictStatusCode = init?.strictStatusCode
+    this.strictResponse = init?.strictResponse
   }
 
   openapi = <
@@ -262,7 +262,7 @@ export class OpenAPIHono<
       }
     }
 
-    if (this.strictResponseSchema) {
+    if (this.strictResponse) {
       const responseZodSchemaObject: Record<string, ZodType<any>> = {}
       for (const [statusCode, responseConfig] of Object.entries(route.responses)) {
         for (const mediaTypeObject of Object.values(responseConfig.content ?? {})) {
@@ -292,7 +292,7 @@ export class OpenAPIHono<
       }
     }
 
-    if (this.strictResponseCode) {
+    if (this.strictStatusCode) {
       validators.push(async (c, next) => {
         await next()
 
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @masnormen 

Thank for the PR! Please wait a little while, I'll look at it in detail later. This may sound difficult, but I'll see if we can't also validate the “types".

masnormen:
Hi, thank you, it's fine. Before this I've also tried exploring the typing but it seems TypeScript doesn't really support it...

https://stackoverflow.com/a/64266045
https://github.com/microsoft/TypeScript/issues/241

yusukebe:
Yeah. We often can't do what we want to do with TypeScript.

yusukebe:
Hi @masnormen 

You are right, type validations seem difficult, so let's not consider it this time.

This PR looks good to me, but I am curious about one point about returning an error. If the validation fails, it returns a 500 response, as shown below, but ideally it would be better to allow the user to customize it.

https://github.com/honojs/middleware/blob/edd032651dcb591eea97a3dc8d5c201373049d63/packages/zod-openapi/src/index.ts#L284-L286

https://github.com/honojs/middleware/blob/edd032651dcb591eea97a3dc8d5c201373049d63/packages/zod-openapi/src/index.ts#L300-L308

However, the API can be complex. What do you think about this?

masnormen:
@yusukebe We probably can make `.route()` take some kind of hook for that purpose.

But at this point there I think would be a need to make distinction between a `beforeHook` for the current implemented hook (for request validation) and a new `afterHook`/`responseHook` (for response validation.

Should we make OpenAPIHono take a `defaultAfterHook`, and make `.route()` take a fourth argument for `afterHook`? Or maybe overload the third argument to take an object like `{ beforeHook: ..., afterHook: ... }`? I saw Elysia use this type of distinction

yusukebe:
@masnormen 

> But at this point, I think there would be a need to make distinction between a `beforeHook` for the current implemented hook (for request validation) and a new `afterHook`/`responseHook` (for response validation.

Yeah. I think so. It's good!

> Should we make OpenAPIHono take a `defaultAfterHook`, and make `.route()` take a fourth argument for `afterHook`? Or maybe overload the third argument to take an object like `{ beforeHook: ..., afterHook: ... }`? I saw Elysia use this type of distinction

I think either is fine, but the second - `{ beforeHook: ..., afterHook: ... }` might be better. It prevents longer arguments and is easier to handle when the number of arguments increases.

masnormen:
Come to think of it, it's better if this before/after hook stuff is implemented in the main Hono first. But currently I don't have the resource to look into that

yusukebe:
@masnormen 

> Come to think of it, it's better if this before/after hook stuff is implemented in the main Hono first.

You are right, maybe we should rethink from the design of Hono's Validator. Or we could make “Response” validation only with Zod OpenAPI. Either way, we'd do it without too much haste.

mandado:
Is there some deadline to merge that PR ? I would like to using that in my project 😄

ewauq:
Any news about this PR? @yusukebe 

I would need this as well in my project. In fact, I thought it was already a thing when I started using the `zod-openapi` middleware because you can define the response schema, so it seemed obvious to me that it would validate it. However, I did not expect that it would not be a strict validation, which poses a security issue. 

I will continue to validate all my response schemas myself with `zod` while waiting for the functionality to be integrated. Thank you very much for your work!

yusukebe:
Ah, it's just an idea.

If we can allow a custom JSON serializer in `c.json()`, we may validate the JSON object in the Response:

```ts
app.get('/api', (c) => {
  return c.json(
    {
      message: 'Hi',
      ok: true,
    },
    jsonSerializerWithResponseValidator({
      schema: z.object({
        message: z.string(),
        ok: z.boolean(),
      }),
    })
  )
})
```

yusukebe:
Anyway, it's not good to parse the `Response` already created, like this PR. So, the best is to add a *hook* for `c.json()` to access the object before creating `Response`.

bennobuilder:
for me enforced types (as specified in the schema) on `c.json()` would already be good enough..

belgattitude:
Recent zod version 3.24 implements the standard schema spec. I'm wondering how does it impact work in this area ?  

https://github.com/standard-schema/standard-schema?tab=readme-ov-file#what-schema-libraries-implement-the-spec

Issue #185, opened 2023-10-01
raise error instead of reponse null and 400 (firebase auth)


shriharip:
Hi,
I was trying to make this work with deno runtime, but I guess might be an issue across all edge networks.

We need to set the middleware for all routes. Reason, on my home page if the user is signed in , (already has a valid authorization) then I need to show Logout Button. 

```
// set middleware
app.use('*', verifyFirebaseAuth(config))
```

Now, let say my root route.
```
app.get('/', (c) => {
 return c.text('Hello from home')
})
````

because the middle ware is called on each route invocation and we try to 

```
 const authorization = c.req.headers.get(config.AuthorizationHeaderKey) 
```

if the authorization is null, we are sending null response with 400. This results in home page also not being shown. So I think, this behavior is not proper or I might be missing something.

yusukebe:
Hi @shriharip 

Middleware such as Basic and Bearer authentication will throw an `HTTPException` when authentication fails.

https://hono.dev/api/exception#throw-httpexception

But, this Firebase Auth middleware only returns a 400 response. If Firebase Auth would throw `HTTPException`, would that solve your problem?

shriharip:
Hello @yusukebe ,

Thanks for your reply and the wonderful honojs . 
I agree that on auth having ```HTTPException``` is valid. However, in my case, I want to show my homepage on my ```/``` index route with some static images and static content. But since the middleware sends a response it is difficult to catch that. 

Yes, having this throw a ```HTTPException``` would be best solution for all those use cases and for me. Thanks.

yusukebe:
@shriharip Okay!

@Code-Hex :

Since [this PR](https://github.com/honojs/hono/pull/796) merged, Basic and Bearer authentication throws [`HTTPException`](https://hono.dev/api/exception#throw-httpexception) when authentication fails. So, can you please make Firebase Auth also throw `HTTPException` with reference to [Basic Auth middleware](https://github.com/honojs/middleware/blob/main/packages/firebase-auth/src/index.ts)?

Code-Hex:
@yusukebe @shriharip I created PR for this: https://github.com/honojs/middleware/pull/191

yusukebe:
This may be fixed by #191. 

@shriharip can you check it out?

shriharip:
Hello @yusukebe , thanks for the fix. 
However, this is still not working as at one place we still need to raise an error.

[...]

PR #191, opened 2023-10-05
fix(firebase-auth): throw HTTPException instead of reponse null

Code-Hex:
I made to throw HTTPException instead of response null 401.

See: https://github.com/honojs/middleware/issues/185

Now, we can achieve our own error handling like the below:
```ts
app.onError((e, c) => {
  if (e instanceof HTTPException && e.status === 401) {
    return c.json(
      {
        message: 'Custom error message!',
      },
      401
    )
  }
})
```

### DIFF
--- a/.changeset/make-http-exception.md
+++ b/.changeset/make-http-exception.md
@@ -1,0 +1,5 @@
+---
+'@hono/firebase-auth': patch
+---
+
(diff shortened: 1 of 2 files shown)


yusukebe:
Thanks @Code-Hex !

Let's go with it!

PR #192, opened 2023-10-05
test(zod-openapi): additional headers test for `Authorization`

callumfrance:
I was trying to figure out how Authorisation headers will work with this repository. I discovered that there is some interesting behaviour with the capitalisation, and as such have added a test into the package to note this behaviour as seen here.

Additionally, I have removed a duplicate dependency from the `package.json` as this is already listed as a full dependency. In most scenarios, having a package listed twice the the `package.json` file is erroneous - I am sure if this was intentional someone will let me know 🙂

### DIFF
--- a/packages/zod-openapi/package.json
+++ b/packages/zod-openapi/package.json
@@ -37,7 +37,6 @@
     "zod": "3.*"
   },
   "devDependencies": {
-    "@hono/zod-validator": "^0.1.9",
     "hono": "^3.7.2",
     "zod": "^3.22.1"
   },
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @callumfrance 

Thank your the PR. Indeed, it would be nice to have a test for `authorization` or `Authorization` in this project!

I think we don't need `route` and `routeWithAuth`, just one. We can remove `route` and rename `routeWithAuth` to `route`. What do you think?

callumfrance:
@yusukebe I have merged the two routes together

yusukebe:
@callumfrance 

Thanks! I'll merge it now!

PR #194, opened 2023-10-06
fix(firebase-auth): throw HTTPException for bad request

Code-Hex:
fix 2 for https://github.com/honojs/middleware/issues/185


jmsheff:
I just ran into this problem right at this exact moment and I was about to put out a PR for it. Thanks for being on top of it!

PR #197, opened 2023-10-06
add unknownutil validator

ryoppippi:
Hi!
I created a middleware of `unknownutil` (https://github.com/lambdalisue/deno-unknownutil)
This is kind of a zod or valibot, but so much faster and lighter!

### DIFF
--- a/.github/workflows/ci-unknwnutil-validator.yml
+++ b/.github/workflows/ci-unknwnutil-validator.yml
@@ -1,0 +1,25 @@
+name: ci-unknownutil-validator
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/unknownutil-validator/**'
+  pull_request:
+    branches: ['*']
+    paths:
+      - 'packages/unknownutil-validator/**'
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./packages/unknownutil-validator
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 18.x
+      - run: yarn install --frozen-lockfile
+      - run: yarn build
+      - run: yarn test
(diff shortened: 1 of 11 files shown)


yusukebe:
Hi @ryoppippi !

Thanks for the PR. I didn't know `unknownutil`!

I'm curious about one thing. It seems to me that `unknownutil` only validates **Types**. So, in practice, it can only be used for "JSON". `zod` and `valibot` will validate the actual value. Examples, length of string, whether it is an email or not.

```ts
z.string().length(5);
z.string().email();
```

But `unknownutil` only handles **Types**. Honesty, I'm not a fan of calling it a Validator, what do you think?

ryoppippi:
Hi @yusukebe 
Thank you for checking and reviewing my PR!!

As you mentioned, `unknownutil` validates only types. It is dead simple.
However, the bundle size is so tiny and it works really fast.

I'm a huge fan of `Hono RPC`.
Recently I've been using `unknownutil` with `Hono` for typing.
Although `unknownutil` lacks some features like `length` or `email`, It is quite useful for simple API. 

If you do not think this is a validator, I'll close this and create an independent repository.

ryoppippi:
Here is the author of `unknownutil` comment about whether it is a validator or not 
https://github.com/icebob/validator-benchmark/pull/7#issuecomment-1751182584

yusukebe:
Hi @ryoppippi,

Thanks for the explanation!

I understand well that `unknownutil` is not a data validator but a type validator. I also understand that you prefer to use `unknownutil`.

However, I would like to reject this PR, because most people imagine an actual data validator when they hear the word "Validator".

> If you do not think this is a validator, I'll close this and create an independent repository.

When you create your repository, you can add the project to this page!

https://hono.dev/middleware/third-party

ryoppippi:
@yusukebe 
So you can close this!!
Cheers

yusukebe:
@ryoppippi 

Okay! Cheers.

Issue #198, opened 2023-10-07
Hono / GraphQL / Pothos integration issues


armstrong-pv:
# Full repro, including tests located here: https://github.com/armstrong-pv/hono-pothos-issue-repro

First of all, I'm really enjoying using Hono - great job on speed and simplicity 👏. I initially started with plain REST, but then moved on to using GraphQL using the middleware you provide. [Pothos](https://github.com/hayes/pothos) is a type safe schema builder which I'd really like to use together with Hono, but I'm having some issues trying to get them to work together. 
1. Context Propagation
I can get the context from the pipeline to go to a root resolver, but not to go to individual resolvers, e.g. for mutations.
2. When I try to use `authScopes` for resolvers, the GraphQL processing returns a "Invalid value used as weak map key". I raised an issue in the Pothos repo here: https://github.com/hayes/pothos/issues/1064, and the author thinks it's related to propagating the context object, so if 1 is solved, then this should be solved also.
3.  It would be nice to opt out of the GraphQL error handling, e.g. by returning a plain old 401 http code, but everything gets wrapped in a GraphQL error. Not sure if there's anything that can be done about this.

I'm using Cloudflare Workers and it's not trivial to get everything up and running, so I have created a full repro of these issues, along with tests that demonstrate them at this repo: https://github.com/armstrong-pv/hono-pothos-issue-repro

There is a readme which covers setting everything up and running the tests: https://github.com/armstrong-pv/hono-pothos-issue-repro/blob/main/README.md

Many thanks!

metrue:
Thanks @armstrong-pv , I will take a look at it.

PR #201, opened 2023-10-15
fix(graphql-server): enhance type safety of RootResolvers and Options

koralle:
related #198 (I'm not familiar with Pothos, so I'm not sure...)

## Summary

This commit introduces more specific type annotations to the following modules:
* `RootResolvers`
* `Options`

Specifically, the same type parameters used in the definition of `Context` class were given to them:

```ts
<E extends Env = any, P extends string = any, I extends Input = {}>
```

See: https://github.com/honojs/hono/blob/main/src/context.ts

## Diff

To confirm, change `bun_test/index.test.ts` as follows (not included in the commit)

```ts
describe('graphql-server', () => {
  // Construct a schema, using GraphQL schema language
  const schema = buildSchema(`
  type Query {
    hello: String
  }
`)

  // My motivation is to development a GraphQL server on Cloudflare Workers with D1.
  type Bindings = {
    DB: string
  }
  
  type GraphQLContext = Context<{ Bindings: Bindings }, '/graphql'>

  // The root provides a resolver function for each API endpoint
  const rootValue = (ctx?: GraphQLContext) => {
    return {
      hello: () => {
        return 'Hello world!'
      },
    }
  }

  const app = new Hono()

  app.use(
    '/graphql',
    graphqlServer({
      schema,
      rootResolver: rootValue,
    })
  )

  // ...


})
```

### Before

スクリーンショット 2023-10-15 13 18 19

スクリーンショット 2023-10-15 13 18 46

スクリーンショット 2023-10-15 13 19 22

### After

スクリーンショット 2023-10-15 13 15 35

スクリーンショット 2023-10-15 13 17 00

スクリーンショット 2023-10-15 16 52 08

### DIFF
--- a/.changeset/old-readers-complain.md
+++ b/.changeset/old-readers-complain.md
@@ -1,0 +1,5 @@
+---
+'@hono/graphql-server': patch
+---
+
+add type safety to RootResolvers and Options
--- a/packages/graphql-server/src/index.ts
+++ b/packages/graphql-server/src/index.ts
@@ -17,26 +17,36 @@ import type {
   GraphQLFormattedError,
 } from 'graphql'
 
-import type { Context } from 'hono'
+import type { Context, Env, Input } from 'hono'
 import { parseBody } from './parse-body'
 
-export type RootResolver = (ctx?: Context) => Promise<unknown> | unknown
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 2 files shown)


koralle:
Sorry.
I've found something that needs to be fixed, so I'll close it for now.

yusukebe:
@koralle 

Thank you for the PR. Looks good! Ship it now.

PR #206, opened 2023-10-20
feat(zod-openapi): allows the response to be `Response`

### LINKED ISSUE #205: Response with `null` body yields a compile error, which should be allowed for 204
This is a continuation from https://github.com/honojs/hono/issues/1529#issuecomment-1745113645, which I still think is a problem on second thought.

This is the reproduction code I've shown in the issue above

```ts
import { OpenAPIHono, createRoute, z } from "@hono/zod-openapi";

const userSchema = z.object({ id: z.string() });
declare const getUserFromDatabase: (
  id: string,
) => z.infer<typeof userSchema> | undefined;

export default new OpenAPIHono()
  .openapi(
    createRoute({
      method: "get",
      path: "/users/{id}",
      responses: {
        200: {
          content: { "application/json": { schema: userSchema } },
          description: "Found",
        },
        404: { description: "Not found" },
      },
    }),
    (c) => {
      const user = getUserFromDatabase(c.req.param("id"));
      return user ? c.jsonT(user) : c.jsonT(null, 404);
    },
  )
  .openapi(
    createRoute({
      method: "put",
      path: "/users/{id}",
      responses: { 204: { description: "Successfull update" } },
    }),
    (c) => {
      // update user info here.
      return c.jsonT(null, 204)
    },
  );
```
 and `c.jsonT(null)` yields the following compile error.

```
Argument of type '(c: Context<Env, "/users/:id", {}>) => TypedResponse<{ id: string; }> | TypedResponse<null>' is not assignable to parameter of type 'Handler<Env, "/users/:id", {}, HandlerResponse<{} | { id: string; }>>'.
  Type 'TypedResponse<{ id: string; }> | TypedResponse<null>' is not assignable to type 'HandlerResponse<{} | { id: string; }>'.
    Type 'TypedResponse<null>' is not assignable to type 'HandlerResponse<{} | { id: string; }>'.
      Type 'TypedResponse<null>' is not assignable to type 'TypedResponse<{} | { id: string; }>'.
        Type 'null' is not assignable to type '{} | { id: string; }'.(2345)
```

Replacing `null` with `{}` eliminates the compile error but is problematic in case of 204 status response. You have to use `null` as a response body if the status is 204:

```ts
import { Hono } from "hono";

// TypeError: Response constructor: Invalid response status code 204
await new Hono().get("/", (c) => c.jsonT({}, 204)).request("/");
```

Returning either `null` or `{}` for 4xx/5xx responses might be a personal preference (I'd prefer the former), but I'd be happy if we can return `null` in Zod OpenAPI Hono because there are no choices for 204.

yusukebe:
With this PR, I made it allows the response type to be `Response` and not just `TypedResponse`.

Resolves #205

### DIFF
--- a/.changeset/tough-beds-speak.md
+++ b/.changeset/tough-beds-speak.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+feat: allows the response to be `Response` not just `TypedResponse`.
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -143,13 +143,17 @@ type Hook<T, E extends Env, P extends string, O> = (
         error: ZodError
       },
   c: Context<E, P>
-) => TypedResponse<O> | Promise<TypedResponse<T>> | void
+) => TypedResponse<O> | Promise<TypedResponse<T>> | Response | Promise<Response> | void
 
 type ConvertPathType<T extends string> = T extends `${infer Start}/{${infer Param}}${infer Rest}`
   ? `${Start}/:${Param}${ConvertPathType<Rest>}`
   : T
 
-type HandlerResponse<O> = TypedResponse<O> | Promise<TypedResponse<O>>
+type HandlerResponse<O> =
+  | TypedResponse<O>
+  | Promise<TypedResponse<O>>
+  | Response
+  | Promise<Response>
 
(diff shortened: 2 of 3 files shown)


yudai-nkt:
LGTM. I was actually looking at the code base to see if I can contribute, but you were so quick! 💯

yusukebe:
@yudai-nkt Okay! Ship it now.

PR #208, opened 2023-10-22
fix(zod-validator): make validation input optional when schema is optional

MonsterDeveloper:
Fixes types for optional schema by making input optional (`?`) when the schema is optional.

See honojs/hono#1583


yusukebe:
Hi @MonsterDeveloper !

Thanks. LGTM. Ship it now.

Issue #247, opened 2023-11-09
Error Encountered When Using @hono/firebase-auth Middleware


tukiminya:
I set up `@hono/firebase-auth` according to the [sample](https://github.com/honojs/middleware/tree/main/packages/firebase-auth#module-worker-syntax-recommend), but when I tried to make a request, I got the following error:

```
{
  message: 'failed to verify the requested firebase token',                                                                                                                                                                                                 
  err: FirebaseAuthError: Error fetching public keys for Google certs: TypeError: Cannot read properties of undefined (reading 'get')                                                                                                                       
      at FirebaseTokenVerifier.mapJwtErrorToAuthError (file:///C:/Users/tukimi/Documents/pj/hono/.wrangler/tmp/dev-CHmjq0/index.js:692:16)                                                                                                                
      at FirebaseTokenVerifier.decodeAndVerify (file:///C:/Users/tukimi/Documents/pj/hono/.wrangler/tmp/dev-CHmjq0/index.js:635:24)                                                                                                                       
      at async file:///C:/Users/tukimi/Documents/pj/hono/.wrangler/tmp/dev-CHmjq0/index.js:901:27                                                                                                                                                         
      at async file:///C:/Users/tukimi/Documents/pj/hono/.wrangler/tmp/dev-CHmjq0/index.js:1997:62                                                                                                                                                        
      at async jsonError (file:///C:/Users/tukimi/Documents/pj/hono/.wrangler/tmp/dev-CHmjq0/index.js:2652:12) {                                                                                                                                          
    errorInfo: {                                                                                                                                                                                                                                            
      code: 'auth/argument-error',                                                                                                                                                                                                                          
      message: "Error fetching public keys for Google certs: TypeError: Cannot read properties of undefined (reading 'get')"                                                                                                                                
    },                                                                                                                                                                                                                                                      
    codePrefix: 'auth'                                                                                                                                                                                                                                      
  }                                                                                                                                                                                                                                                         
}
[mf:inf] GET /test 500 Internal Server Error (14ms)
```

index.ts
```ts
import { Hono } from 'hono'
import { VerifyFirebaseAuthConfig, verifyFirebaseAuth } from '@hono/firebase-auth';

const firebaseConfig: VerifyFirebaseAuthConfig = {
  projectId: 'MY_PROJECT_ID'
}
const app = new Hono()

app.use('/test', verifyFirebaseAuth(firebaseConfig))
app.get('/test', (c) => c.text('Hi!'))

export default app
```

[...]

PR #248, opened 2023-11-11
fixed firebase-auth for #247

Code-Hex:
This fix comes as a set with this: https://github.com/Code-Hex/firebase-auth-cloudflare-workers/pull/10


yusukebe:
Hi @Code-Hex,

Wow, using Miniflare's `mf.get*` is cool! Looks good, I'm merging it now.

Thanks!

PR #251, opened 2023-11-13
Authentication middleware

monoald:
This PR adds a new middleware for social login, at the moment it has authentication with Facebook, Github, Google and LinkeIn, and more coming soon.


yusukebe:
Hi @monoald,

Awesome work!

One thing I'm wondering about: Have you considered separating the packages for each platform? This would mean creating:

* `@hono/facebook-auth`
* `@hono/github-auth`
* `@hono/google-auth`
* `@hono/linkedin-auth`

I think this approach, rather than combining everything into one package, is better. Separating the packages makes maintenance easier, and users can choose to install only the packages they want to use.

monoald:
Makes a lot of sense. I'm going to make the changes and create a new PR

yusukebe:
Cool!

PR #252, opened 2023-11-13
Facebook, Github, Google and Linkedin Auth

monoald:
Adding new middlewares for authentication.
Previous middleware split into different middlewares.


yusukebe:
@monoald

Super cool! It looks good at a glance. I'll review them later.

@Code-Hex, could you review this when you have time?

pilcrowOnPaper:
Any reasons to split them into different packages? There’s like 10 major OAuth providers. And since OAuth is standardized, it might better to provide a single OAuth package?

yusukebe:
@pilcrowOnPaper

I think that if they don't share many common logics, they should be separated to maintain better maintainability. If it's a single package, an update to one provider would require incrementing the version of the entire package, which includes all providers.

pilcrowOnPaper:
@yusukebe Speaking from experience, there's a lot shared logic since they all follow the OAuth 2.0 authorization code spec

yusukebe:
@pilcrowOnPaper

Thank you. If that's the case, we might need to consolidate them into a single package, like "OAuth Middleware" or "OAuth Providers Middleware".

However, the code in this PR appears to have limited common logic.

Also, this isn't just an implementation issue; it's about how we package these features - whether as a unified "OAuth" or as distinct "Google Auth, GitHub Auth, Facebook Auth, and LinkedIn Auth".

I would like to hear @monoald's opinion on this.

packages/facebook-auth/README.md
```diff
@@ -0,0 +1,162 @@
...
+		- `token` for quick access to the token, suitable when the app's authentication flow will be managed by the **client side**.
+		- `code%20token`:  A combination of both the code and access token received as parameters on the url.
+- `include_granted_scopes`:
+	- Type: `boolean`.
+	- `Optional`.
+	- To obtain the scopes for which the user has **granted permission**, set this value to `true`. This will allow you to implement a flow in cases where your app cannot access crucial permissions and needs to request them again, among other use cases.  Value `false` by default.
+- `fields`:
+	- Type: `string[]`.
+	- Fields you request from the Facebook API to be sent once the user has logged in. You can find a comprehensive reference for all the fields you can request on the [Facebook User Reference page](https://developers.facebook.com/docs/graph-api/reference/user/#fields).
+- `state`:
```
pilcrowOnPaper, line 94: `state` isn't a secret. It just needs to be different time. You defeat the purpose of using `state` if you use the same value every time. A new `state` should be generated and stored (usually as a cookie) every time the user is redirected to the provider's login page.

https://pilcrow.vercel.app/blog/oauth-guide#authorization-url

packages/google-auth/README.md
```diff
@@ -0,0 +1,177 @@
...
+	- Type: `string[]`.
+	- `Required`.
+	- Set of **permissions** to request the user's authorization to access your app for retrieving user information and performing actions on their behalf.<br /> Review all the scopes Google offers for utilizing their API on the [OAuth 2.0 Scopes page](https://developers.google.com/identity/protocols/oauth2/scopes). 
+	> If your app is not **verified** by Google, the accessible scopes for your app are significantly **limited**.
+- `response_type`:
+	- Type: `string`.
+	- `Required`.
+	- Choose the **OAuth flow** for the authentication process. The available options are:
+		- `code` for a more secure flow, applicable when the app's authentication flow is managed by the **server side**.
+		- `token` for quick access to the token, suitable when the app's authentication flow will be managed by the **client side**.
```
pilcrowOnPaper, line 85: Implicit grant type is not recommended, and is pretty much deprecated. I don't see a reason to implement this since Hono is a sever side library and it's intended for purely client side applications. If you were to implement this, `client_secret` should be optional

packages/google-auth/src/authFlow.ts
```diff
@@ -0,0 +1,153 @@
...
+      redirect_uri: this.redirect_uri,
+      client_id: this.client_id,
+      include_granted_scopes: this.include_granted_scopes,
+      scope: this.scope.join(' '),
+      state: this.state
+    })
+    const url = `https://accounts.google.com/o/oauth2/v2/auth?${parsedOptions}`
+
+    if (this.response_type === 'token') {
+      return html`
```
pilcrowOnPaper, line 75: Any reasons to redirect users client side, and not just use `Location` header?

packages/google-auth/src/googleAuth.ts
```diff
@@ -0,0 +1,57 @@
...
+      code: c.req.query('code'),
+      token: {
+        token: c.req.query('access_token') as string,
+        expires_in: Number(c.req.query('expires-in')) as number
+      }
+    })
+
+    // Avoid CSRF attack by checking state
+    if (c.req.url.includes('?') && c.req.query('state') !== options.state) {
+      throw new HTTPException(401, { message: 'Unauthorized, risk of CSRF attack.' })
```
pilcrowOnPaper, line 37: I'd just return a 400/401 and remove the error message (you usually want to make these vague)

packages/github-auth/src/githubAuth.ts
```diff
@@ -0,0 +1,38 @@
+import type { MiddlewareHandler } from 'hono'
+
+import { AuthFlow } from './authFlow'
+import type { GithubScope } from './types'
+
+export function githubAuth(options: {
```
pilcrowOnPaper, line 6: Every provider should implement `state` if it's available

packages/facebook-auth/src/authFlow.ts
```diff
@@ -0,0 +1,106 @@
...
+    const response = await fetch(`https://graph.facebook.com/v18.0/me?access_token=${this.token?.token}`)
+      .then(res => res.json()) as FacebookMeResponse
+
+    this.resError(response)
+
+    this.user = response
+  }
+
+  private resError(response: FacebookErrorResponse) {
+    if (response.error) throw new HTTPException(500, { message: response.error?.message })
```
pilcrowOnPaper, line 91: The error can be caused by an invalid authorization code as well, in which case it should be a 400 error response

packages/facebook-auth/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+    code: number
+    fbtrace_id: string
+  }
+}
+
+export type FacebookTokenResponse = {
+  access_token: string
+  token_type: string
+  expires_in: number
+} & FacebookErrorResponse
```
pilcrowOnPaper, line 52: Shouldn't you use an intersection instead of a union?

packages/facebook-auth/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type FacebookMeResponse = {
+  name: string
+  id: string
+} & FacebookErrorResponse
+
+export type FacebookUser = {
```
pilcrowOnPaper, line 59: This type won't be correct if you add custom fields to the `fields` option, or was that intentional?

packages/linkedin-auth/src/authFlow.ts
```diff
@@ -0,0 +1,129 @@
...
+    const params = toQueryParams({
+      grant_type: 'authorization_code',
+      code: this.code,
+      client_id: this.client_id,
+      client_secret: this.client_secret,
+      redirect_uri: this.redirect_uri
+    })
+
+    const response = await fetch(
+      `https://www.linkedin.com/oauth/v2/accessToken?${params}`,
```
pilcrowOnPaper, line 66: FYI you can pass the `URLSearchParams` to the `fetch()` body instead

packages/github-auth/src/types.ts
```diff
@@ -0,0 +1,81 @@
...
+export type GithubErrorResponse = {
+  error: string
+  message: string
+  documentation_url: string
+}
+
+export type GithubTokenResponse = {
+  access_token: string
+  expires_in: number
+  refresh_token: string
```
pilcrowOnPaper, line 30: The refresh token is only included for Github Apps, and not for OAuth Apps

packages/github-auth/src/types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export type GithubErrorResponse = {
+  error: string
+  message: string
+  documentation_url: string
+}
+
+export type GithubTokenResponse = {
+  access_token: string
+  expires_in: number
```
pilcrowOnPaper, line 29: This field is omitted for OAuth Apps (but not GitHub Apps) since access tokens don't expire

packages/facebook-auth/README.md
monoald: Didn't thought about it, I will handle the creation and storage of the `state` internally. Thanks!

packages/google-auth/README.md
monoald: Yeah It makes no sense, the OAuth flow should use the corresponding method to server side. Thanks!

packages/google-auth/src/authFlow.ts
monoald: Could't get the params because of the hash on the redirect uri from google. I'll use only the code flow now so this will be removed.

packages/google-auth/src/googleAuth.ts
monoald: Right, thanks!

packages/github-auth/src/githubAuth.ts
monoald: I forgot it there, my bad.

packages/facebook-auth/src/authFlow.ts
monoald: Ok

packages/facebook-auth/src/types.ts
monoald: Yep I totally forget that, thanks!

packages/facebook-auth/src/types.ts
monoald: Didn't test all the fields due to the app verification process to get access to all the permissions. To avoid type errors an ugly "type union" can be used. Do you have a suggestion for this case?

packages/linkedin-auth/src/authFlow.ts
monoald: Oh right, it is way safer

packages/github-auth/src/types.ts
monoald: My bad, thanks!

packages/facebook-auth/src/types.ts
pilcrowOnPaper: I don't have a good suggestion either. I guess you can decouple the user fetching and export `getFacebookUser()` which would allow the package to provide a "recommended" way while giving an option to implement their own way of getting the user

monoald:
Yes, as the provider use OAuth 2.0 the authentication on each share a lot of logic, may be some extra steps are required for certain providers, But I still don't know if it is better to have a unique middleware or split it and make more. For maintenance it could be better have them separate, but there will be an unnecessary amount of middlewares. 
I will refactor and think about the best solution, whether to have a single middleware or keep them separate

monoald:
By the way thank you @pilcrowOnPaper, I really appreciate all the insights

pilcrowOnPaper:
I mean like a single OAuth package that exports middleware for each provider, and not a single unified OAuth middleware, if that wasn't clear

monoald:
Ooh, I didn't understand. It may work well as a single package.

yusukebe:
@monoald, sorry for any confusion caused by my previous statements.

@pilcrowOnPaper, thank you so much for your comments and reviews.

I'll try using these middlewares and get a sense of how they work. I'll comment later!

yusukebe:
@monoald

I've tried the GitHub OAuth, and it works really well and easily. Cool!

@monoald @pilcrowOnPaper

After rethinking, I believe it should be into one package. We need to decide on a name. How about `@hono/oauth-providers`? It could be structured like the following:

```ts
import { OAuthVariables } from '@hono/oauth-providers'
import { githubAuth } from '@hono/oauth-providers/github'
```

PR #255, opened 2023-11-15
fix(zod-openapi): Fix incorrect specification of the exports field in…

Karibash:
… package.json


yusukebe:
Hi @Karibash !

Thanks for the PR. LGTM.

The CI is failing, but it's only an issue with the tests. I'll fix it after this PR is merged.

PR #258, opened 2023-11-16
fix(zod-openapi): Fix a bug that slashes were duplicated when mountin…

Karibash:
…g a path using the route method

Before:
スクリーンショット 2023-11-16 14 14 45

After:
スクリーンショット 2023-11-16 14 15 02

### DIFF
--- a/.changeset/four-rice-listen.md
+++ b/.changeset/four-rice-listen.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Fix a bug that slashes were duplicated when mounting a path using the route method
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @Karibash!

It's a bug! It should be fixed.

Instead of `normalizePath`, I suggest using `mergePath`, which is exported from `hono/utils/url`:

https://github.com/honojs/hono/blob/1ed742bdb80518437e6d6f9c6ec7a5fbe300f254/src/utils/url.ts#L89

This function is used in the core logic of Hono's `app.route()`, making it a better choice. Ordinarily, I wouldn't recommend using methods from `hono/utils/*` to **end-users**, as these are intended for internal use and might have breaking changes without announcement. But, in this case, it's appropriate to use the same function.

yusukebe:
@Karibash 

Looks good! Thanks!

yusukebe (approved):

PR #260, opened 2023-11-16
feat(zod-openapi): Make it possible to do method chaining even for do…

Karibash:
…c methods

I implemented it so that doc methods can also do method chaining.
I am not so sure if I implemented it properly because the type is too complicated.

### DIFF
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -28,6 +28,7 @@ import type {
 import type { MergePath, MergeSchemaPath } from 'hono/types'
 import type { RemoveBlankRecord } from 'hono/utils/types'
 import { mergePath } from 'hono/utils/url'
+import type { OpenAPIObject } from 'openapi3-ts/oas30'
 import type { AnyZodObject, ZodSchema, ZodError } from 'zod'
 import { z, ZodType } from 'zod'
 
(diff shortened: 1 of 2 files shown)


packages/zod-openapi/src/index.ts
```diff
@@ -277,18 +278,24 @@ export class OpenAPIHono<
     return document
   }
 
-  doc = (path: string, config: OpenAPIObjectConfig) => {
-    this.get(path, (c) => {
+  doc = <P extends string>(
+    path: P,
+    config: OpenAPIObjectConfig
+  ): OpenAPIHono<E, S & ToSchema<'get', P, {}, OpenAPIObject>, BasePath> => {
```
yusukebe, line 284: The below is OK.

```ts
OpenAPIHono<E, S & ToSchema<'get', P, {}, {}>, BasePath>
```

packages/zod-openapi/src/index.ts
```diff
@@ -277,18 +278,24 @@ export class OpenAPIHono<
...
-    })
+    }) as any
   }
 
-  doc31 = (path: string, config: OpenAPIObjectConfig) => {
-    this.get(path, (c) => {
+  doc31 = <P extends string>(
+    path: P,
+    config: OpenAPIObjectConfig
+  ): OpenAPIHono<E, S & ToSchema<'get', P, {}, OpenAPIObject>, BasePath> => {
```
yusukebe, line 294: Same as above.

yusukebe (commented):
Hi @Karibash 

Yeah, it's complicated as you said. I've left the comments.

packages/zod-openapi/src/index.ts
Karibash: @yusukebe 
Since this method does not use the jsonT method, it looks like your suggestion is more correct.
Please check the following commit for the fix.
https://github.com/honojs/middleware/pull/260/commits/3915c72217bae435fdae3335ca1a496bc28370d6

yusukebe: > Since this method does not use the jsonT method

Yes, you are right. It does not use `jsonT()`! Your fix is good.

yusukebe:
@Karibash 

Thanks! Looks good to me. I'll merge this now.

PR #265, opened 2023-11-20
fix(swagger-ui): support `persistAuthorization` and boolean option

### LINKED ISSUE #264: swagger-ui: support `persistAuthorization`
[`persistAuthorization`](https://github.com/swagger-api/swagger-ui/blob/d3bb64f1de150ea1ff677accf705c9ce0ae9fe85/flavors/swagger-ui-react/README.md?plain=1#L161) allows credentials to be stored in localstorage. this is useful for quickly refreshing and testing changes in API with swagger UI.

```ts
new Hono()
	.get("/openapi.json", (c) => c.json(doc))
	.get("/openapi", swaggerUI({ url: "/openapi.json", persistAuthorization: true }))
```
however, this doesn't type checks and will not render properly,

```ts
new Hono()
	.get("/openapi.json", (c) => c.json(doc))
	.get("/openapi", swaggerUI({
          url: "/openapi.json",
          manuallySwaggerUIHtml: (asset) => /*html*/`
          <div>
              <div id="swagger-ui"></div>
              ${asset.css.map((url) => `<link rel="stylesheet" href="${url}" />`)}
              ${asset.js.map((url) => `<script src="${url}" crossorigin="anonymous"></script>`)}
              <script>
                  window.onload = () => {
                      window.ui = SwaggerUIBundle({
                          dom_id: '#swagger-ui',
                          url: '/openapi.json',
                          persistAuthorization: true,
                      })
                  }
              </script>
          </div>
          `
    }))
```
while this works.

I'm interested in opening PR.

scarf005:
- fix #264

### DIFF
--- a/packages/swagger-ui/src/swagger/renderer.ts
+++ b/packages/swagger-ui/src/swagger/renderer.ts
@@ -14,6 +14,7 @@ export type DistSwaggerUIOptions = {
   operationsSorter?: string
   requestInterceptor?: string
   responseInterceptor?: string
+  persistAuthorization?: boolean
 }
 
 const RENDER_TYPE = {
@@ -37,6 +38,7 @@ const RENDER_TYPE_MAP = {
   operationsSorter: RENDER_TYPE.RAW,
   requestInterceptor: RENDER_TYPE.RAW,
   responseInterceptor: RENDER_TYPE.RAW,
+  persistAuthorization: RENDER_TYPE.RAW,
 } as const satisfies Record<
   keyof DistSwaggerUIOptions,
   (typeof RENDER_TYPE)[keyof typeof RENDER_TYPE]
--- a/.changeset/soft-llamas-melt.md
+++ b/.changeset/soft-llamas-melt.md
@@ -1,0 +1,5 @@
+---
+'@hono/swagger-ui': minor
+---
+
+Added support for `persistAuthorization`.
--- a/packages/swagger-ui/test/option-renderer.test.ts
+++ b/packages/swagger-ui/test/option-renderer.test.ts
@@ -124,4 +124,12 @@ describe('SwaggerUIOption Rendering', () => {
       })
     ).toEqual('responseInterceptor: (res) => res')
   })
+
+  it('renders correctly with persistAuthorization', () => {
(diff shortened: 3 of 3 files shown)


sor4chi:
Hi @scarf005, Thanks for PR! Nice work.
I’m creator of @hono/swagger-ui.
I didn’t notice such a option field existed.

BTW, I found that more unsupported fields from [this](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/).
Should we support all of these?

cc: @yusukebe

packages/swagger-ui/src/swagger/renderer.ts
```diff
@@ -37,6 +39,7 @@ const RENDER_TYPE_MAP = {
   operationsSorter: RENDER_TYPE.RAW,
   requestInterceptor: RENDER_TYPE.RAW,
   responseInterceptor: RENDER_TYPE.RAW,
+  persistAuthorization: RENDER_TYPE.BOOLEAN,
```
sor4chi, line 42: I don't think it is necessary to add `RENDER_TYPE.BOOELAN` because you can just use `RENDER_TYPE.RAW` for boolean.

.changeset/tiny-sheep-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/swagger-ui': patch
```
sor4chi, line 2: I think this changes will be minor release. How about this, @yusukebe?

sor4chi (commented):
I've left some comments!

packages/swagger-ui/src/swagger/renderer.ts
scarf005: yeah that'd be better.

yusukebe:
Hi @scarf005 

Looks good to me!

@sor4chi 

> BTW, I found that more unsupported fields from [this](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/).
> Should we support all of these?

As you like. But I'll merge this first.

sor4chi:
@yusukebe OK, I'll create the issue!

PR #269, opened 2023-11-20
feat(zod-openapi): add easy way to configure custom components

### LINKED ISSUE #261: authorization header not passing while sending request in swagger-ui using zod-openapi
Hi, I am trying to send bearer token via header in swagger UI but after putting the token inside required field, it is not using the header bearer token while sending request that user gives.
 I have tried using "custom-header", it seems that it is being passed while sending request
```
swagger.openapi(
	createRoute({
		method: "get",
		path: "/me",
		tags: ["Users"],
		summary: "Get user details",

		request: {
			headers: z.object({
				Authorization: z.string(),
			}),
		},
		responses: {
			200: {
				description: "Respond with a message",
				content: {
					"application/json": {
						schema: z.object({
							message: z.string(),
						}),
					},
				},
			},
			401: {
				description: "Unauthorized",
				content: {
					"application/json": {
						schema: z.object({
							error: z.string(),
						}),
					},
				},
			},
		},
	}),
	(c) => {
		return c.jsonT({
			message: "success",
		});
	},
);
```

sor4chi:
close #261

### DIFF
--- a/.changeset/clever-seahorses-sing.md
+++ b/.changeset/clever-seahorses-sing.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+The `OpenAPIHono` constructor can now register OpenAPI components.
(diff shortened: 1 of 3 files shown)


sor4chi:
@yusukebe 

Sorry for the lack of additional information.

I added a feature to make it easier to register a component so that it can be set from the constructor, but I realized that it can be accessed from `new OpenAPIHono().openAPIRegistory`, so I closed this PR once, and then I added a separate one on how to register a component I would like to write a separate document on how to register components.

yusukebe:
Okay! Thanks!

Issue #243, opened 2023-11-07
How access context in `app.doc` with  `@hono/zod-openapi` ?


Manouchehri:
I'm trying to set the server to be whatever the current request origin is, example:

```typescript
app.doc('/openapi.json', {
  openapi: '3.0.0',
  info: {
    version: '1.0.0',
    title: 'Example API'
  },
  servers: [
    {
      url: new URL(c.req.url).hostname,
      description: 'Current environment'
    }
  ],
})
```

yusukebe:
Hi @Manouchehri 

Currently, there is no APIs to access the context from `app.doc`, but we need it. I or someone can implement it.

PR #272, opened 2023-11-22
feat(zod-openapi): Make context accessible in the doc route

Karibash:
close: #243

### DIFF
--- a/.changeset/tough-eels-exist.md
+++ b/.changeset/tough-eels-exist.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @Karibash 

I'm sorry to be late.

Looks good! One thing. Could you write the short usage of this feature on README? Thanks!

Karibash:
@yusukebe 
I wasn't sure where to add the explanation, but I added it to the Tips section anyway.
Also, I found some typos, so I fixed them as well.

yusukebe:
Hi @Karibash 

That place is good!

Let's ship it now. Thanks!

PR #279, opened 2023-11-28
fix: ensure state is passed through for both github app types

rawkode:
When using a GitHub App, instead of a GitHub OAuth App, no user can successfully authenticate because the state isn't added to the redirect URI.

### DIFF
--- a/.changeset/hip-owls-give.md
+++ b/.changeset/hip-owls-give.md
@@ -1,0 +1,5 @@
+---
+'@hono/oauth-providers': patch
+---
+
+ensure CSRF state returned to GitHub apps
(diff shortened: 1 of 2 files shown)


rawkode:
cc @yusukebe @monoald

monoald:
Hi @rawkode, thanks for the fix!

@yusukebe it works fine.

yusukebe:
@rawkode Thanks!

One thing. Could you run the `yarn changeset` command at the top level of this project? This will be a `patch` change.

rawkode:
Sure thing

yusukebe:
LGTM!

PR #280, opened 2023-11-28
fix: redirect url for github apps

rawkode:
OAuth apps can't have multiple callback URLs, but GitHub Apps can.

As such, we want to make sure we call back to the same location for GitHub apps and not the first configured callbackURL in the app config.

cc @yusukebe @monoald

### DIFF
--- a/.changeset/two-crabs-speak.md
+++ b/.changeset/two-crabs-speak.md
@@ -1,0 +1,5 @@
+---
+'@hono/oauth-providers': patch
+---
+
+ensure redirect to original URL for GitHub apps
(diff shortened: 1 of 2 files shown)


rawkode:
@monoald I did start with that implementation, but it felt a bit wierd to pass the `c.req.url` down into the `GitHubAuthFlow` only to bubble it back up.

Of course, if that's the preferred approach I'm happy to make that change.

yusukebe:
Hmm. I think current @rawkode 's way is better.

yusukebe:
@rawkode @monoald 

Ready to merge?

monoald:
Yep

PR #291, opened 2023-12-06
Use origin for zod-openapi example

Manouchehri:
`.hostname` doesn't include the protocol, which is important. =)


yusukebe:
@Manouchehri 

You are right! Thanks!

PR #309, opened 2023-12-13
feat: introduce React Renderer Middleware

yusukebe:
This PR introduces the 'React Renderer Middleware'.

It is similar to the JSX Renderer Middleware in Hono's core, but it uses React. This might be used in Sonik or other meta-frameworks built on Hono. React supports not only server-side but also client-side rendering. Therefore, you can use it in a different way compared to Hono's JSX. Additionally, you can use a lot libraries available in React's ecosystem.

### DIFF
--- a/packages/react-renderer/package.json
+++ b/packages/react-renderer/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "@types/react": "^18",
     "@types/react-dom": "^18.2.17",
-    "hono": "^3.11.6",
+    "hono": "^3.11.7",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "tsup": "^8.0.1",
--- a/packages/react-renderer/package.json
+++ b/packages/react-renderer/package.json
@@ -1,0 +1,49 @@
+{
+  "name": "@hono/react-renderer",
+  "version": "0.0.0",
+  "description": "React Renderer Middleware for Hono",
+  "type": "module",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "vitest --run",
+    "build": "tsup ./src/index.ts --external hono,react,react-dom --format esm,cjs --dts",
+    "publint": "publint"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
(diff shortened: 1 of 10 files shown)


yusukebe:
Hi @usualoma 

Could you review this one?

packages/react-renderer/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "devDependencies": {
+    "@types/react": "^18",
+    "@types/react-dom": "^18.2.17",
+    "hono": "^3.11.7",
+    "react": "^18.2.0",
```
usualoma, line 41: I'm not familiar with package.json, so I apologize if I'm wrong.

I think dependencies is more appropriate for the meaning since it calls "react" and "react-dom/server" at runtime. (e.g. in "packages/react-renderer/src/react-renderer.ts") Is there any reason why you dare to use "devDependencies"?

usualoma:
@yusukebe Oh, Great job!

The `useRequestContext` is also implemented, which should make it easy to migrate to and from hono's `jsxRenderer` 👍 

I made one comment about devDependencies, please confirm.

packages/react-renderer/package.json
yusukebe: @usualoma

I'm also not super familiar with `package.json` :)

I expect that users will use this React Renderer with installing `react`, `react-dom`:

```txt
npm i @hono/react-renderer react react-dom hono
npm i -D @types/react @types/react-dom
```

So, I think we don't need to list them as `dependencies`, and I've added them to `devDependencies`. If we want to specify versions for `react` and `react-dom`, we should include them in `peerDependencies`.

usualoma: Uh, sorry. I understand, devDependencies is good.
Maybe we should put react in peerDependencies. For example

```
  "peerDependencies": {
    "hono": "*",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  },
```

usualoma: I had added my comment without reading your 9:29 comment.

Anyway, I understand that the current code is what was intended and that there are no problems, so I think we can merge this PR!

usualoma (approved):

packages/react-renderer/package.json
yusukebe: Thanks!

PR #313, opened 2023-12-15
change: Export Hook in @hono/zod-openapi

WildEgo:
This will let people create shareable defaultHooks, that way I can just use one snippet of code in all my sub routers.


yusukebe:
LGTM! Thanks!

PR #321, opened 2023-12-18
docs(qwik-city) support all request methods

tleperou:
🙌 

The Qwik City's actions request against the method POST.

In the event Qwik City may request against the others, I changed to use the `all` instead of the `get`.

(1) Open to comments; (2) shall I create a branch?


yusukebe:
Hi @tleperou !

Thanks! This is okay. I'll merge it now.

PR #330, opened 2023-12-28
added react as peer

divyam234:
@yusukebe  react should be added as peer  dependency some how I missed in 1st pr.

### DIFF
--- a/.changeset/beige-sloths-raise.md
+++ b/.changeset/beige-sloths-raise.md
@@ -1,5 +1,5 @@
 ---
-'@hono/auth-js': major
+'@hono/auth-js': patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/beige-sloths-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/auth-js': major
```
yusukebe, line 2: This seems to be `patch` release. So please write `patch` instead of `major`.

yusukebe:
Hi @divyam234 

I've left a comment. Please check it.

yusukebe:
@divyam234 

Thanks!

PR #340, opened 2024-01-04
Adding hono/prometheus middleware

dios-david:
Fixes https://github.com/honojs/hono/issues/263

As suggested in [this comment](https://github.com/honojs/hono/issues/263#issuecomment-1193129541), I'm adding this to the `honojs/middleware` repository instead of the main one as a third-party middleware.

I've tried to come up with sensible configuration options and default values based on my previous experiences with [express-prom-bundle](https://www.npmjs.com/package/express-prom-bundle) and [koa-prometheus](https://www.npmjs.com/package/@uswitch/koa-prometheus).

### DIFF
--- a/packages/prometheus/README.md
+++ b/packages/prometheus/README.md
@@ -2,6 +2,16 @@
 
 This middleware adds basic [RED metrics](https://www.weave.works/blog/the-red-method-key-metrics-for-microservices-architecture/) to your Hono application, and exposes them on the `/metrics` endpoint for Prometheus to scrape.
 
+## Installation
+
+This package depends on `prom-client`, so you need to install that as well:
+
+```bash
+npm install -S @hono/prometheus prom-client
+# or
+yarn add @hono/prometheus prom-client
+```
+
 ## Usage
 
 ```ts
--- a/packages/prometheus/README.md
+++ b/packages/prometheus/README.md
@@ -19,6 +19,52 @@ app.get('/', (c) => c.text('foo'))
 export default app
 ```
 
+Making a GET request to `/metrics` returns the string representation of the metrics:
+
+```
+# HELP http_request_duration_seconds Duration of HTTP requests in seconds
+# TYPE http_request_duration_seconds histogram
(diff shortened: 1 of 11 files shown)


dios-david:
Hey @yusukebe, can you review this please?

packages/prometheus/README.md
```diff
@@ -0,0 +1,166 @@
+# Prometheus middleware for Hono
+
+This middleware adds basic [RED metrics](https://www.weave.works/blog/the-red-method-key-metrics-for-microservices-architecture/) to your Hono application, and exposes them on the `/metrics` endpoint for Prometheus to scrape.
+
```
yusukebe, line 4: I think it would be better to have an installation guide. Users may not understand that they have to install `prom-client`.

packages/prometheus/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/honojs/middleware.git"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "^3.9.2",
```
yusukebe, line 35: Since `3.11.6` or lower is a security risk, specify the latest version, `3.12.0` or higher.

yusukebe:
Hi @dios-david !

Awesome work! API design and implementation are great to me. I've left some comments. Check them.

packages/prometheus/package.json
dios-david: I used another package as a base which turned out wasn't the best choice :) Thanks, I've updated this to 3.12.0

dios-david:
Hey @yusukebe,
Sorry for the late response, I couldn't find time earlier to apply your suggestions. Now these should be fine :)

yusukebe (approved):
Looks good!

yusukebe:
@dios-david 

Thanks! Let's land it.

PR #349, opened 2024-01-08
Fix: Apply headers when using serveStatic; Improvement: Make options optional;

floriankapaun:
I noticed the previous implementation did not work when using serveStatic middleware. This fixes this.

### DIFF
--- a/packages/bun-transpiler/src/index.ts
+++ b/packages/bun-transpiler/src/index.ts
@@ -2,30 +2,32 @@ import Bun from 'bun'
 import { createMiddleware } from 'hono/factory'
 
 type BunTranspilerOptions = {
-  extensions: string[]
-  headers: Record<string, string | string[]>
-  transpilerOptions: Bun.TranspilerOptions
+  extensions?: string[]
+  headers?: Record<string, string | string[]>
+  transpilerOptions?: Bun.TranspilerOptions
 }
 
-export const bunTranspiler = (
-  options: BunTranspilerOptions = {
-    extensions: ['.ts', '.tsx'],
-    headers: { 'content-type': 'application/javascript' },
-    transpilerOptions: {
-      minifyWhitespace: true,
-      target: 'browser',
-    },
-  }
-) => {
+const DEFAULT_OPTIONS: BunTranspilerOptions = {
+  extensions: ['.ts', '.tsx'],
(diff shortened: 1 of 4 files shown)


packages/bun-transpiler/src/index.ts
```diff
@@ -2,52 +2,48 @@ import Bun from 'bun'
...
-  options: BunTranspilerOptions = {
-    extensions: ['.ts', '.tsx'],
-    headers: { 'content-type': 'application/javascript' },
-    transpilerOptions: {
-      minifyWhitespace: true,
-      target: 'browser',
-    },
-  }
-) => {
+const DEFAULT_OPTIONS: BunTranspilerOptions = {
```
yusukebe, line 10: How about exporting `DEFAULT_OPTIONS` and having users use it?

Like this:

```ts
import { defaultOptions } from '@hono/bun-transpiler'

app.use(
  '*',
  bunTranspiler({
    extensions: ['mts', ...defaultOptions.extensions],
  })
)
```

For this reason, I would like to do these.

* Because users use it, so rename it to `defaultOptions`.
* Set the type as `Required<BunTranspilerOptions>`.

yusukebe:
Hi @floriankapaun !

I've left the comment. Please check it.

packages/bun-transpiler/src/index.ts
floriankapaun: That's a good point! I like that suggestion. Will expose `deufaltOptions`. I'm not convinced about requiring it as an input tough because why would I force the developer to provide me with the defaultOptions? Or did I understand your second suggestion wrong @yusukebe ?

yusukebe: I'm sorry! I didn't explain enough. As it is now, we get the following error.

Screenshot 2024-01-10 at 7 06 58

This is because the property is optional. So you can avoid the error by adding `Required` to the `defaultOptions` type as follows.

```ts
export const defaultOptions: Required<BunTranspilerOptions> = {
  extensions: ['.ts', '.tsx'],
  //...
}
```

floriankapaun: Oh! Now I understand. Didn't check that. Good catch! Adjusted to your suggested solution.

yusukebe:
@floriankapaun 

Thanks! Merge now!

PR #361, opened 2024-01-28
    fix(ctx): fix the context not propagated down to resolvers.

metrue:
This MR is to fix one of the issues raised in https://github.com/honojs/middleware/issues/198 , the context should be included in the execution options.


yusukebe (approved):
LGTM!

PR #363, opened 2024-01-28
Make it possible to define a different HTTP Status Code without using…

Matzu89:
… a hook

Hope you open to the following..

I usually return 422 on invalid data, would be nice to set this without writing out a hook on every request.


yusukebe:
Hi @Matzu89 

Use a hook instead of it:

```ts
app.get(
  '/',
  zValidator('query', schema, (result, c) => {
    if (!result.success) {
      return c.json(result, 422)
    }
  }),
  (c) => {
    return c.json(0)
  }
)
```

PR #370, opened 2024-01-30
feat: add @hono/schema-validator middleware

sebastianwessel:
This middleware leverages TypeSchema, offering an abstraction layer that facilitates interaction with a variety of validation libraries through a unified interface.

Consequently, there is no immediate requirement to develop a dedicated middleware for each validation library.  
This not only reduces maintenance efforts but also extends support to validation libraries that may currently lack compatibility.

List of possible validation libs: [https://typeschema.com/](https://typeschema.com/#coverage)

```typescript
import { z } from 'zod'
import { schemaValidator, type ValidationError } from '@hono/schema-validator'

// schema could also some other validation lib
// see https://typeschema.com/#coverage
const schema = z.object({
  name: z.string(),
  age: z.number(),
})

app.post('/author', schemaValidator('json', schema), (c) => {
  const data = c.req.valid('json')
  return c.json({
    success: true,
    message: `${data.name} is ${data.age}`,
  })
})

app.onError(async (err, c) => {
  if (err instanceof ValidationError) {
    return c.json(err, err.status)
  }
})

```


yusukebe:
Hi @sebastianwessel !

This is super interesting. I'd like to merge this, but please wait a bit. We'll release the v4 of Hono core soon. There will be some breaking changes for the validator. So it is better to wait until v4 is released than to merge now.

yusukebe:
@sebastianwessel 

Until that, could you add the CI in `.github/workflows`?

yusukebe:
@sebastianwessel 

Thanks.

> From my side, there is no rush, and it's fine to focus on 4.0

Yeah. Please wait a moment.

inetol:
What is the status of this PR?

yusukebe:
Hi @sebastianwessel 

v4 is available now. Can you update this PR?

sebastianwessel:
Updated the pr

Also changed the typeschema packages to their new nameings - here I'm not sure if this package should use the [@typeschema/main](https://github.com/decs/typeschema/tree/main/packages/main) or [@typeschema/all](https://github.com/decs/typeschema/tree/main/packages/all) (currently used)

see: https://typeschema.com/#quickstart

yusukebe:
Hi @sebastianwessel 

Why not change the name to `typeschema-validator` or `type-schema-validator` instead of `schema-validator`? I think `schema` is too generic, and that library is called `TypeSchema`.

sebastianwessel:
@yusukebe 
makes sense. what do you think about the included packages.
should it use `all` which is larger, but without the need to add peer-dependencies, or should it use `main` which is smaller, but requires to in install peer deps?

sebastianwessel:
@yusukebe renamed it.

I also switched to the `main` instead of all. Here, the user needs to add peer dependencies, depending on the requirements

packages/typeschema-validator/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "description": "Validator middleware for multiple schema validation libraries based on schema.com",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest",
+    "build:cjs": "tsc -p tsconfig.cjs.json",
```
yusukebe, line 13: Can you use `tsup` to build and `vitest` for testing, referring to the `hello` middleware?

https://github.com/honojs/middleware/tree/main/packages/hello

packages/swagger-editor/package.json
yusukebe, line 1: This file's change is not needed.

PR #372, opened 2024-01-30
Add OpenID Connect authentication middleware

hnw:
Hi.

This PR introduces a new middleware for Hono which provides OpenID Connect authentication.

It also provides storage-less login sessions. It should be useful for serverless/edge applications.


packages/oidc-auth/README.md
```diff
@@ -0,0 +1,111 @@
...
+| OIDC_ISSUER                   | The issuer URL of the OpenID Connect (OIDC) discovery. This URL is used to retrieve the OIDC provider's configuration. | None, must be provided |
+| OIDC_CLIENT_ID                | The OAuth 2.0 client ID assigned to your application. This ID is used to identify your application to the OIDC provider. | None, must be provided |
+| OIDC_CLIENT_SECRET            | The OAuth 2.0 client secret assigned to your application. This secret is used to authenticate your application to the OIDC provider. | None, must be provided |
+| OIDC_REDIRECT_URI             | The URL to which the OIDC provider should redirect the user after authentication. This URL must be registered as a redirect URI in the OIDC provider. | None, must be provided |
+
+## How to Use
+
+```ts
+import { Hono } from 'hono'
+import { requireOidcAuthMiddleware, getAuth, revokeSession, processOAuthCallback } from '@hono/oidc-auth';
```
yusukebe, line 58: `requireOidcAuthMiddleware` should be `oidcAuthMiddleware`.

packages/oidc-auth/README.md
```diff
@@ -0,0 +1,111 @@
...
+})
+
+export default app
+```
+
+## Another example: Cloudflare Pages with OpenID connect login
+
+```ts
+import { Hono } from 'hono'
+import { requireOidcAuthMiddleware, getAuth } from '@hono/oidc-auth';
```
yusukebe, line 82: Same as the above, `requireOidcAuthMiddleware` should be `oidcAuthMiddleware`.

packages/oidc-auth/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@hono/oidc-auth",
+  "version": "1.0.0",
```
yusukebe, line 3: Before releasing, the version should be`0.0.0`.

packages/oidc-auth/src/index.ts
```diff
@@ -0,0 +1,289 @@
...
+  if (oauth2.isOAuth2Error(result)) {
+    throw new Error(`OAuth2Error: [${result.error}] ${result.error_description}`)
+  }
+  return result
+}
+
+/**
+ * Returns a middleware that requires OIDC authentication.
+ */
+export const oidcAuthMiddleware = ()  => {
```
yusukebe, line 255: I think it would be better to pass the variables for this middleware as arguments instead of environment variables. Like the following:

```ts
app.use('*', oidcAuthMiddleware({
  sessionSecret: 'xxx',
  sessionRefreshInterval: 123,
  sessionExpires: 123,
  ...
}))
```

packages/oidc-auth/src/index.ts
```diff
@@ -0,0 +1,289 @@
...
+        setCookie(c, 'state', state, { path: '/' , httpOnly: true, secure: true})
+        setCookie(c, 'nonce', nonce, { path: '/' , httpOnly: true, secure: true})
+        setCookie(c, 'code_verifier', code_verifier, { path: '/' , httpOnly: true, secure: true})
+        setCookie(c, 'continue', c.req.url, { path: '/' , httpOnly: true, secure: true})
+        return c.redirect(url)
+      }
+    } catch (e) {
+      console.log(e)
+      deleteCookie(c, oidcSessionCookieName)
+      throw new Error('Invalid session')
```
yusukebe, line 279: We can use `HTTPException`:

https://hono.dev/api/exception

yusukebe:
Hi @hnw !

Thanks for the PR. This is awesome work!

As for the build and CI, they were too much trouble, so I fixed them myself. And, I've left some comments. Please check them.

packages/oidc-auth/src/index.ts
hnw: I understand that passing configuration via environment variables is not a good idea. It should have portability issues.

However, if settings are passed as constructor arguments for middleware, the settings cannot be used in paths where middleware is not applied.

Authentication is required for paths to which `oidcAuthMiddleware` is applied. On the other hand, the callback and logout path should be processed without authentication, so we cannot apply `oidcAuthMiddleware` to these paths. Nevertheless, `revokeSession` and `processOAuthCallback` functions used in these paths require OIDC endpoint information.

For the above reasons, I created a separate function for configuration.

hnw:
Hi @yusukebe

Thanks for your feedback. I've made the changes you suggested.

packages/oidc-auth/src/index.ts
yusukebe: Thank you for the explanation! Understood well.

And I know you created the `configureOidcAuth,` which seems to be good, but I think there is a problem.

For Cloudflare Workers, we can't get the environment variables globally. We can get only in the handler. So, we have to use the approach such as Bearer Auth middleware does like the following:

```ts
import { bearerAuth } from 'hono/bearer-auth'

type Bindings = {
  TOKEN: string
}

const app = new Hono<{ Bindings: Bindings }>()

app.get('/auth/*', async (c, next) => {
  const auth = bearerAuth({
    token: c.env.TOKEN,
  })
  await auth(c, next)
})
```

Based on that, I see two ways to go about it.

### 1. Options pattern

In the case of Cloudflare, we want to use environments, but as you mentioned, they are used in other functions. The solution is to have a function like `setupOidcAuthMiddleware` and write it at the top to get environment variables first.

```ts
const app = new Hono()

// Put this at the top
app.use('*', setupOidcAuthMiddleware({
  sessionSecret: 'xxx',
  sessionRefreshInterval: 123,
  sessionExpires: 123,
  ...
}))
```

### 2. Environment variables pattern

The other is to get the value of the setting in an environment variable, as you did at first. For this, `env()` in `hono/adapter` may be useful. This will give you multi-runtime support.

https://hono.dev/helpers/adapter#env

---

Either way is fine, but 2 might be better since it is opnionative. Also, it might be good to see how other middleware solves that problem.

hnw: Thank you for your detailed feedback on my code.

I have followed your 2nd suggestion, making use of environment variables for the middleware's configuration, and have successfully implemented it using `env()` from `hono/adapter`. This indeed provides us with the flexibility of multi-runtime support, which I find quite beneficial. For example, we can use it on Netlify to provide OpenID authentication to static pages.

I've committed these changes and pushed them to the repository. Could you please review them when you get a chance?

Thank you again for your constructive advice.

yusukebe: I think that choice is good!

We will release version 4.0.0 tomorrow, so let's hold off on the merge until then. I may have the `peerDependencies` changed.

yusukebe (approved):
LGTM!

yusukebe:
Hi @hnw !

It's finished. Let's land it! Thank you so much for your contribution.

PR #373, opened 2024-02-01
Export RouteConfig to allow wrappers around createRoute

Gablas:
Make it easy to wrap createRoute


yusukebe:
Hi @Gablas 

Not a bad idea, but you can import directly from `@asteasolutions/zod-to-openapi`.

```ts
import type { RouteConfig } from '@asteasolutions/zod-to-openapi'
```

If other users don't have much chance to use `RouteConfig`, we don't have to accept this PR.

yusukebe:
#532, the same purpose was merged. Thanks!

PR #377, opened 2024-02-04
feat: add support for authorization cookies in firebase-auth middleware

Datron:
Hi,

This PR adds support for cookies to be used in firebase-auth middleware instead of headers. The use case for this is if you want access to a dashboard or component available only to users who have logged in and don't want to/can't add headers for HTTP requests made by said users.

### DIFF
--- a/.changeset/quick-bikes-work.md
+++ b/.changeset/quick-bikes-work.md
@@ -1,0 +1,5 @@
+---
+'@hono/firebase-auth': minor
+---
+
+Allow cookies to be used instead of headers for providing firebase auth token
--- a/packages/firebase-auth/README.md
+++ b/packages/firebase-auth/README.md
@@ -74,9 +74,17 @@ app.fire()
 
 This field indicates your firebase project ID.
 
+### `useCookie: boolean` (**required**)
+
+This field, if set to true will pick the token to verify from a cookie instead of Authorization headers.  
+
 ### `authorizationHeaderKey?: string` (optional)
 
 Based on this configuration, the JWT created by firebase auth is looked for in the HTTP headers. The default is "Authorization".
+
+### `cookieName?: string` (optional)
+
+Based on this configuration, the JWT created by firebase auth is looked for in the HTTP cookie using this name. The default is "authorization".
 
 ### `keyStore?: KeyStorer` (optional)
 
@@ -99,6 +107,44 @@ Throws an exception if JWT validation fails. By default, this is output to the e
 You can specify a host for the Firebase Auth emulator. This config is mainly used when **Service Worker Syntax** is used.
 
 If not specified, check the [`FIREBASE_AUTH_EMULATOR_HOST` environment variable obtained from the request](https://github.com/Code-Hex/firebase-auth-cloudflare-workers#emulatorenv).
+
+## Security Considerations when using cookies for authentication
+
+When considering that a web framework uses tokens via cookies, security measures related to traditional browsers and cookies must be considered.
+
+    CSRF (Cross-Site Request Forgery)
+    XSS (Cross-Site Scripting)
+    MitM (Man-in-the-middle attack)
+
+Let's consider each:
+
+**CSRF**
+
+This is provided by hono as a standard middleware feature
+
+https://hono.dev/middleware/builtin/csrf
+
+**XSS**
+
+An attacker can inject a script and steal JWTs stored in cookies. Set the httpOnly flag on the cookie to prevent access from JavaScript. Additionally, configure "Content Security Policy" (CSP) to prevent unauthorized script execution. It is recommeded to force httpOnly and the functionality here: https://hono.dev/middleware/builtin/secure-headers
+
+**MitM**
+
+If your cookie security settings are inappropriate, there is a risk that your cookies will be stolen by a MitM. Use Samesite (or hono csrf middleware) and __Secure- prefix and __Host- prefix attributes.
+
+https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes
+
+An example of good cookie settings:
+
+```
+let secureCookieSettings: CookieOptions = {
+  path: '/',
+  domain: <your_domain>,
+  secure: true,
+  httpOnly: true,
+  sameSite: 'Strict',
+}
+```
 
 ## Author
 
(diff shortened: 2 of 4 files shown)


yusukebe (commented):
First, there are two things:

1. Don't include `package-lock.json`. Use yarn instead of npm.
2. The change to `package.json` is not necessary.

yusukebe:
@Datron 

Thanks for the PR!

Hey @Code-Hex, could you review this?

Datron:
@yusukebe Thanks for reviewing! I have made the changes you requested. I have a query though, why not include package-lock.json? Wouldn't it make sure that all contributors are building and testing the same builds locally? Pinning dependency versions also seem like a safe thing to do.

Datron:
@yusukebe you can ignore my question, I saw that you have a yarn.lock file.

Code-Hex:
@Datron Hi 👋 

I don't think the use case you describe in detail provides a strong reason why cookies should be used. I honestly couldn't understand it.

When considering that a web framework uses tokens via cookies, security measures related to traditional browsers and cookies must be considered.

1. CSRF (Cross-Site Request Forgery)
2. XSS (Cross-Site Scripting)
3. MitM (Man-in-the-middle attack)

Nothing seems to have been considered in this PR.

Let's consider each:

### CSRF

This is provided by hono as a standard middleware feature, so I don't think it's necessary here temporary. 

https://hono.dev/middleware/builtin/csrf

### XSS

An attacker can inject a script and steal JWTs stored in cookies. Set the `httpOnly` flag on the cookie to prevent access from JavaScript. Additionally, configure "Content Security Policy" (CSP) to prevent unauthorized script execution. I think we need to force `httpOnly` and the functionality here: https://hono.dev/middleware/builtin/secure-headers

### MitM

If your cookie security settings are inappropriate, there is a risk that your cookies will be stolen by a MitM. We must use [`Samesite`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) (or hono csrf middleware) and `__Secure-` prefix and `__Host-` prefix attributes.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes 


## Wrap up

@Datron  I don't think you need to consider `csrf` for now, but as a minimum modification you should set the cookie settings described here.

- `httpOnly`
- `__Secure` prefix and `__Host` prefix.

Here's a summary of how to set them up.

|          | Secure Attribute | Path Attribute | Domain Attribute      |
|----------|------------------|----------------|-----------------------|
| __Secure | Required         | (None)         | (None)                |
| __Host-  | Required         | /              | Must not be present   |

@yusukebe Actually, I think it would be nice if csrf could also be enabled here. Do you have any ideas?

Datron:
Hello @Code-Hex,

> I don't think the use case you describe in detail provides a strong reason why cookies should be used. I honestly couldn't understand it.

My bad, let me explain with an example. I use hono in my app and all endpoints return HTML. The state is managed by the server, and the frontend merely links users to different pages. Some pages are authenticated, but when I redirect a user via an `<a>` tag I can't set headers for the GET HTTP request. So I used cookies in my application to store the firebase auth token, and modified firebase-auth middleware to read the cookie instead of the header (since I couldn't send the token in the header).

> When considering that a web framework uses tokens via cookies, security measures related to traditional browsers and cookies must be considered.
>
> CSRF (Cross-Site Request Forgery)
> XSS (Cross-Site Scripting)
> MitM (Man-in-the-middle attack)
>
> Nothing seems to have been considered in this PR.

Wouldn't these considerations be made at the time of setting the cookie? The code I have added only reads it if available, and if the cookie cannot be read a 401 HTTP error is returned by the server.

I handled this in my application using hono's [cookie helpers](https://hono.dev/helpers/cookie)

Code-Hex:
@Datron  It seems that it is not supported yet. This might be your chance to contribute!
https://github.com/honojs/hono/issues/1203

> Wouldn't these considerations be made at the time of setting the cookie? 

Anyway, I understand your motivation. Can I include the notes I wrote in the README?
It would be even better if you could include an example of how to use `setCookie` with these points in mind.

Datron:
> It seems that it is not supported yet. This might be your chance to contribute!
Sure! I'd love to help out, I really like working with hono.

> Anyway, I understand your motivation. Can you include the notes I wrote in the README?
> It would be even better if you could include an example of how to use setCookie with these points in mind.

Understood. Shall I include it under a header "Security Considerations when using cookies for authentication"? I will add your write up, along with an example of `setCookie` in this section

Code-Hex:
@Datron Good!
Sure! Thank you so much!

.changeset/quick-bikes-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/firebase-auth': major
```
yusukebe, line 2: This is a `minor` update, not a `major`.

yusukebe (commented):
Can you also write a test for when `useCookie: true`?

yusukebe:
@Code-Hex Thanks!

@Datron I've commented. Check them.

Datron:
@yusukebe @Code-Hex  I've added tests and resolved comments.

Code-Hex:
@Datron Looks good.
However, this issue hasn't been resolved.

This PR should be merged after resolved this issues.

Datron:
@Code-Hex let me look into it, I will get some time off work soon.

Code-Hex:
@Datron Hi 👋 

I have supported session cookie management following [this guide](https://firebase.google.com/docs/auth/admin/manage-cookies).

https://github.com/Code-Hex/firebase-auth-cloudflare-workers/releases/tag/v1.2.1


I'm considering submitting a PR for a verify middleware that uses session cookies. What do you think?

Datron:
@Code-Hex Hello!

I took a look at the guide and your commit [here](https://github.com/Code-Hex/firebase-auth-cloudflare-workers/commit/caed773d9fd29f567025552aab1830cb6bca02f6#diff-41614821a46376cf92cd25a6f7ce60dc32b9e7e22c179e52ed3440967c020233)

Everything looks good and I like the idea. I have one concern, I thought that f[irebase Admin SDK did not work on cloudflare workerd runtime as it requires nodejs](https://github.com/firebase/firebase-admin-node/issues/2069).

Code-Hex:
@Datron I have released [firebase-auth-cloudflare-workers v2](https://github.com/Code-Hex/firebase-auth-cloudflare-workers/tree/v2.0.0).

In this version, it is now possible to verify revoked users. Considering this change, I will update the middleware on my side, so please make adjustments to the `__Secure` prefix and `__Host` prefix.

Code-Hex:
@Datron Merged my PR!

Datron:
@Code-Hex I've raised a PR here, please take a look and let me know what you think
https://github.com/honojs/hono/pull/2269

PR #380, opened 2024-02-09
Change peer dependency to support v4.0.0

CarlosZiegler:
Update hono peerDependency to support all future versions starting from 3.0.0, ensuring compatibility with the latest releases without specifying an upper limit.

To avoid some warings :

image


yusukebe:
Hi @CarlosZiegler 

These should be fixed immediately. I've updated the `yarn.lock` and added the changeset by myself. I'll merge it now. Thanks!

marcosrjjunior:
Was this update released? I'm still getting the same warning message to install `@hono/sentry`

Screenshot 2024-02-14 at 4 20 23 pm
Screenshot 2024-02-14 at 4 20 14 pm

probably because it is missing the reference [on the changeset file](https://github.com/honojs/middleware/pull/380/files#diff-3ed4f6a9a89e8074a1fe796039eaca8671cd4a99ed974c11821f1153df0d3cadR1)?

yusukebe:
Hi @CarlosZiegler 

Yeah, you are right. We missed the changeset for the sentry middleware. I've released the new version now. Please check it!

marcosrjjunior:
that was quick! thanks @yusukebe

PR #384, opened 2024-02-12
fix(valibot-validator): make validation input optional when schema is optional

MonsterDeveloper:
Fixes types for optional schema by making input optional (`?`) when the schema is optional.

Exactly the same case as with `zod-validator` honojs/hono#1583


yusukebe (approved):
LGTM!

yusukebe:
Hi @MonsterDeveloper !

Thanks! Looks good. Merging now.

PR #386, opened 2024-02-13
fix(oidc-auth): Fix "yarn release" for Yarn v4 / Build ESM only

hnw:
Hi @yusukebe !
Thanks for releasing `@hono/oidc-auth`!

However, I noticed that there are no `dist/*` files in the npm package for `@hono/oidc-auth@1.0.0`.
This is due to the `yarn release` not executing the `build` script in the `package.json`.
I have fixed the `release` script to address this issue.

Additionally, `yarn publish` was a Yarn v1 command, so I have updated it to be compatible with v4.

Lastly, I noticed that the dependent library, `oauth4webapi`, does not support CommonJS.
As a result, I have updated the `package.json` to only build ESM.

### DIFF
--- a/.changeset/curvy-lemons-give.md
+++ b/.changeset/curvy-lemons-give.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': patch
+---
+
+Fix "yarn release" and fix npm package
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "build:auth-js": "yarn workspace @hono/auth-js build",
     "build:bun-transpiler": "yarn workspace @hono/bun-transpiler build",
     "build:prometheus": "yarn workspace @hono/prometheus build",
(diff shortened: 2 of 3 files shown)


yusukebe:
@hnw 

Ah, sorry for missing it. Looks good but could you run the `yarn changeset`?

hnw:
Finally, I've understood what "yarn workspace" is. I apologize for the multiple force pushes.

.changeset/curvy-lemons-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/oidc-auth': minor
```
yusukebe, line 2: This should be `patch` instead of `minor`.

yusukebe: Hi @hnw 

 This may be last!

yusukebe:
@hnw 

Thanks! Merge now.

PR #393, opened 2024-02-15
fix(firebase-auth): retrive header values for v4

Code-Hex:
fixed `TypeError: Cannot read properties of undefined (reading 'get')` error in hono v4


Code-Hex:
@yusukebe Could you review please 🙏

yusukebe (approved):
Ah, Indeed, `jest.config.js` is not necessary now!

LGTM!

PR #399, opened 2024-02-17
ci: update actions/checkout and actions/setup-node to v4

ryuapp:
As the title suggests, update actions/checkout and actions/setup-node to v4.


yusukebe:
@ryuapp 

Thanks! Merge now.

PR #408, opened 2024-03-03
fix(zod-openapi): Fix basePath method to transfer defaultHook of the parent app.

DavidHavl:
There is currently an issue with basePath method where it does not transfer defaultHook to the newly created instance.
This pull request fixes it.

In the following code (taken from the current docs) the defaultHook is never called because it does not exist on the new instance created when basePath method is called.

```
const app = new OpenAPIHono({
  defaultHook: (result, c) => {
    if (!result.success) {
      return c.json(
        {
          ok: false,
          errors: formatZodErrors(result),
          source: 'custom_error_handler',
        },
        422
      )
    }
  },
}).basePath('/api/v1')

### DIFF
--- a/packages/zod-openapi/test/index.test.ts
+++ b/packages/zod-openapi/test/index.test.ts
@@ -1279,9 +1279,7 @@ describe('Named params in nested routes', () => {
   it('Should return a correct path', async () => {
     const res = await root.request('/doc')
     expect(res.status).toBe(200)
-    const data: {
-      paths: { string: unknown }
-    } = await res.json()
+    const data = await res.json()
     expect(Object.keys(data['paths'])[0]).toBe('/root/{rootId}/sub/{subId}')
   })
 })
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/zod-openapi/test/index.test.ts [hunk 1]
```diff
@@ -1270,7 +1279,9 @@ describe('Named params in nested routes', () => {
   it('Should return a correct path', async () => {
     const res = await root.request('/doc')
     expect(res.status).toBe(200)
-    const data = await res.json()
+    const data: {
```
yusukebe, line 1282: I think this change is not necessary. What is it for?

yusukebe:
Hi @DavidHavl 

Thank you for the PR. I've left the comment. And could you run `yarn changeset`? This will be a `patch` change.

packages/zod-openapi/test/index.test.ts
yusukebe: We can ignore the typescript error, so please remove it!

DavidHavl: Ok. Removed.

yusukebe:
@DavidHavl 

Thanks! Last one. thing. Run `yarn changeset` on the top of the project, add it, and push!

DavidHavl:
Done

yusukebe (approved):
LGTM!

yusukebe:
@DavidHavl 

Merge now. Thanks!

PR #412, opened 2024-03-06
feat(zod-openapi): Add `toDoc` function to generate OpenAPI specification files

watany-dev:
Introduces a new function called `toDoc` that allows users to generate OpenAPI specification files.

### features:

Accepts a Hono application instance, a file system module, and optional configurations
Supports exporting in JSON or YAML format
Returns an export result object with success status, output directory, generated files, and error (if any)

### Usage:

```ts
import { OpenAPIHono, toDoc } from '@hono/zod-openapi'
import fs from 'fs/promises'

const app = new OpenAPIHono()

// Define routes and schemas...

await exportOpenAPI(app, fs, {
  outDir: './docs',
  format: 'yaml',
  extension: '.yml',
})
```

### Configuration options:

- outDir: Output directory for the generated files (default: './dist')
- format: File format, either 'json' or 'yaml' (default: 'json')
- extension: File extension (default: based on format)

### DIFF
--- a/packages/zod-openapi/package.json
+++ b/packages/zod-openapi/package.json
@@ -38,7 +38,8 @@
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
     "hono": ">=3.11.3",
-    "zod": "3.*"
+    "zod": "3.*",
+    "yaml": "^2.4.0"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20240117.0",
@@ -52,8 +53,7 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^5.5.0",
-    "@hono/zod-validator": "0.1.11",
-    "yaml": "^2.4.0"
+    "@hono/zod-validator": "0.1.11"
   },
   "engines": {
     "node": ">=16.0.0"
--- a/packages/zod-openapi/package.json
+++ b/packages/zod-openapi/package.json
@@ -52,7 +52,8 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^5.5.0",
-    "@hono/zod-validator": "0.1.11"
+    "@hono/zod-validator": "0.1.11",
+    "yaml": "^2.4.0"
   },
   "engines": {
     "node": ">=16.0.0"
--- a/packages/zod-openapi/src/export.ts
+++ b/packages/zod-openapi/src/export.ts
@@ -2,7 +2,6 @@ import { stringify as yamlStringify } from 'yaml'
 import type { OpenAPIHono } from './index'
 
 export interface ToFilesOptions {
-  paths?: string[]
   outDir?: string
   format?: 'json' | 'yaml'
   extension?: string
@@ -37,10 +36,11 @@ const fetchAndSaveOpenAPIDocument = async (app: OpenAPIHono, fs: FileSystem, doc
 export const toFiles = async (
   app: OpenAPIHono,
   fs: FileSystem,
+  paths: string[],
   options: ToFilesOptions = {}
 ): Promise<ToFilesResult> => {
-  const { paths, outDir = './dist', format = 'json', extension } = options
-  if (!paths || paths.length === 0) {
+  const { outDir = './dist', format = 'json', extension } = options
+  if (paths.length === 0) {
     return { success: false, outDir, files: [], error: new Error('OpenAPI document endpoints are not specified') }
   }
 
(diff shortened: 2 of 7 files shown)


packages/zod-openapi/package.json
```diff
@@ -52,7 +52,8 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^5.5.0",
-    "@hono/zod-validator": "0.1.11"
+    "@hono/zod-validator": "0.1.11",
+    "yaml": "^2.4.0"
```
yusukebe, line 56: I don't want to include `yaml` in the dependencies as much as possible. So, it might be a good idea to specify `yaml` in `peerDependencies`.

packages/zod-openapi/src/export.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export interface ToDocResult {
+  success: boolean
+  outDir: string
+  files: string[]
+  error?: Error
+}
+
+export const toDoc = async (
```
yusukebe, line 22: I like the name `toDoc`, but the word `Doc` can be confusing in the case of OpenAPI. `toFiles` might be fine. Bu, any other good names?

packages/zod-openapi/src/export.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const getContent = (openApiJson: unknown, format: string): string =>
+  format === 'yaml' ? yamlStringify(openApiJson) : JSON.stringify(openApiJson, null, 2)
+
+const docEndpointConditions = ['getOpenAPIDocument', 'getOpenAPI31Document']
+
+const findDocEndpoint = (app: OpenAPIHono) => {
+  const docEndpoints = app.routes.filter((route) => {
+    const configure = route.handler.toString()
```
yusukebe, line 79: I think this code is risky.

I have tried `route.handler.toString()` and what I can get is the string of the content of the `handler` function. It's possible if `getOpenAPIDocument` is used in an unintended handler and the handler will be detected as the endpoint handler. Also, if the code is minified, I don't think it will work as intended.

What do you think?

yusukebe:
Hi @watany-dev !

This is a cool feature. However, we have issues to consider. Please check my comments.

watany-dev:
Sorry for the delay. I've made the changes.

yusukebe:
Hi @watany-dev !

What do you think about using `toSSG()` for this implementation? For example, for a path `/doc` and YAML, it would look like this:

```ts
import { toSSG } from 'hono/ssg'
import { stringify as yamlStringify } from 'yaml'
import fs from 'fs/promises'
import app from './src'

toSSG(app, fs, {
  beforeRequestHook: (req) => {
    const url = new URL(req.url)
    if (url.pathname === '/doc') {
      return req
    }
    return false
  },
  // TODO: afterResponseHook does not support async function
  afterResponseHook: async (res) => {
    if (res.headers.get('content-type')?.startsWith('application/json')) {
      const json = await res.json()
      const yaml = yamlStringify(json)
      res = new Response(yaml, {
        headers: {
          // TODO: toSSG does not support application/yaml
          'content-type': 'application/yaml'
        }
      })
      return res
    }
    return false
  }
})
```

watany-dev:
@yusukebe
This is nice! It looks like the toSSG hook needs to be able to accept asynchronous functions in order to switch to this implementation. I will fix that first.

packages/zod-openapi/src/export.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const generateFilePath = (basePath: string, docEndpointPath: string, format: string, extension?: string): string => {
+  const ext = extension || (format === 'yaml' ? '.yaml' : '.json')
+  const sanitizedPath = docEndpointPath.replace(/\//g, '-').replace(/^-*|-*$/g, '')
+  return `${basePath}/openapi-${sanitizedPath}${ext}`
+}
+
+const fetchAndSaveOpenAPIDocument = async (app: OpenAPIHono, fs: FileSystem, docEndpointPath: string, filePath: string, format: string): Promise<string> => {
+  const response = await app.request(`http://localhost${docEndpointPath}`)
```
fumieval, line 29: IIUC you don't have to add a doc endpoint in the first place, because you can obtain the OpenAPI document via `getOpenAPIDocument`. Am I missing something?

fumieval:
After this patch #445, users can just do `writeFile('/path/to', YAML.stringify(app.getOpenAPIDocument(...))`

yusukebe:
Hi @fumieval 

> After #445, users can just do `writeFile('/path/to', YAML.stringify(app.getOpenAPIDocument(...))`

You are right. This is cool.

@watany-dev Have you tried it? I think this PR is not necessary if you like the way.

watany-dev:
@yusukebe @fumieval 
Thank you. As you have given me a lot of advice, this PR is unnecessary and I will close it.

platform-kit:
@fumieval I'm sorry to revive this thread, but I have been unable to find the correct documentation. Can you give a full example of how to output the generated openapi spec to a file? (Including any necessary imports, etc)

PR #421, opened 2024-03-14
OAuth Providers Github App email problem

monoald:
When trying authentication with Github App, Github only sends the user's profile public email if there is any we receive `null`.
I am now explicitly requesting user emails to get a non-null value.

### DIFF
--- a/.changeset/sharp-fans-explain.md
+++ b/.changeset/sharp-fans-explain.md
@@ -1,0 +1,5 @@
+---
+'@hono/oauth-providers': minor
+---
+
+Github App user email problem
--- a/packages/oauth-providers/src/providers/github/authFlow.ts
+++ b/packages/oauth-providers/src/providers/github/authFlow.ts
@@ -21,6 +21,9 @@ type Token = {
   token: string
   expires_in?: number
 }
+
+const userAgent = 'Hono-Auth-App'
+
 export class AuthFlow {
   client_id: string
   client_secret: string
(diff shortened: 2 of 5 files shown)


packages/oauth-providers/src/providers/github/authFlow.ts
```diff
@@ -103,6 +109,26 @@ export class AuthFlow {
       throw new HTTPException(400, { message: response.message })
     }
 
+    if (!this.oauthApp) {
+      const emails = (await fetch('https://api.github.com/user/emails', {
+        headers: {
+          Authorization: `Bearer ${this.token?.token}`,
+          'User-Agent': 'Hono-Auth-App',
```
yusukebe, line 116: It's better to make `Hono-Auth-App` as a constant.

.changeset/sharp-fans-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/oauth-providers': minor
```
yusukebe, line 2: I think it will be `patch` release because this PR is `fix`.

How do you think?

monoald: Right! My bad. I'll change it now

monoald:
Changes made it, `"Hono-Auth-App"` as a constant  and minor bump changed for patch bump.

yusukebe (commented):
LGTM!

yusukebe:
@monoald 

Thanks! Let's go with it.

PR #456, opened 2024-04-16
fix(firebase-auth): error handling to wrap root errors

Code-Hex:
Summary:
- fix error handling to wrap root errors
- updated deps
  - hono to use v4
  - vitest to use v1.5.0


Code-Hex:
@yusukebe Could you review this one please? 🙏

packages/firebase-auth/src/index.ts
```diff
@@ -142,10 +143,12 @@ export const verifySessionCookieFirebaseAuth = (
       })
       setFirebaseToken(c, idToken)
     } catch (err) {
-      const res = c.redirect(config.redirects.signIn)
-      throw new HTTPException(res.status, {
+      const status = 302
+      const res = c.redirect(config.redirects.signIn, status)
+      throw new HTTPException(status, {
         res,
         message: `failed to verify the requested firebase token: ${String(err)}`,
```
yusukebe, line 150: Is `${String(err)}` necessary? I think it may not be needed if passing an `err` to a `cause`.

yusukebe:
Hi @Code-Hex 

I've made one comment. And is it difficult to write tests?

packages/firebase-auth/src/index.ts
Code-Hex: @yusukebe Thanks!
This is for backward compatibles. Until now, there was no way to determine a root error, only whether it was contained in a string. Removing this process breaks existing application logic.

So I left it as it was!

Code-Hex:
@yusukebe It's OK!
> write tests

I will add test to check unwrap errors.

packages/firebase-auth/src/index.ts
yusukebe: Thank you for the explanation! Understood well! Let's go with it.

Code-Hex:
@yusukebe I added some tests for this!

yusukebe (approved):
LGTM!

yusukebe:
@Code-Hex 

Thanks! I'll merge this now.

Issue #455, opened 2024-04-15
access hono context from trpc route


benjamine:
when running hono in cloudflare, I'd need the hono context to access for example a D1 db at `c.env.DB`.
In general I think the trpc routes should get access to the hono context.

This could be achieved by adding a `createContext` parameter to `trpcServer` but if provided as a parameter it won't have access to the specific hono context,

this could be done modifying `tprcServer` implementation [here](https://github.com/honojs/middleware/blob/ffd67e5412815cd407ecab31ea10d23b7f5a761c/packages/trpc-server/src/index.ts#L9-L18) like this:

``` ts
export const trpcServer = ({
	endpoint = "/trpc",
	...rest
}: tRPCOptions): MiddlewareHandler => {
	return async (c) => {
		const res = fetchRequestHandler({
			...rest,
			endpoint,
			req: c.req.raw,
			createContext: () => ({
				env: c.env,
			}),
		});
		return res;
	};
};
```

although maybe a more generic solution is preferrable (this would interfere with a `createContext` provided as config in `rest`)

yusukebe:
Hi @benjamine Thank you for raising the issue.

> this could be done modifying `tprcServer` implementation [here](https://github.com/honojs/middleware/blob/ffd67e5412815cd407ecab31ea10d23b7f5a761c/packages/trpc-server/src/index.ts#L9-L18) like this:

This seems to be good! Can you create the PR with this change and tests?

benjamine:
@yusukebe I gave it a try at #458 , similar to the code example above, but I also allowed people to specify a `createContext` without breaking the expected behavior, and still have control to customize how data is copied from hono context if they want to.

PR #458, opened 2024-04-16
feat(trpc): access hono context from trpc procedures

benjamine:
this would solve #455

change should be backwards compatible (so not bumping the minor version).

### DIFF
--- a/.changeset/dull-carrots-wash.md
+++ b/.changeset/dull-carrots-wash.md
@@ -1,0 +1,5 @@
+---
+'@hono/trpc-server': minor
+---
+
+access hono context from trpc procedures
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @benjamine 

Look good! One thing: can you run `yarn changeset` on the top level of the project to create a changeset? I think it will be a `minor` change (not patch or major), because it provide a new feature.

benjamine:
@yusukebe sorry for the delay, just added the minor changeset

yusukebe:
@benjamine 

Thank you very much! I'll merge this and release it now.

benjamine:
you're welcome, thank you for hono! 🙏

PR #478, opened 2024-04-25
fix(auth-js): env AUTH_URL not working

DIYgod:
In the previous implementation, `AUTH_URL` was invalid because `c.req.raw` already includes the URL base. The provided URL base, `c.env.AUTH_URL`, will not be utilized. This issue will prevent the hono + auth.js project deployed behind a proxy from obtaining the correct `callback-url`.

### DIFF
--- a/.changeset/itchy-dolls-crash.md
+++ b/.changeset/itchy-dolls-crash.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': patch
+---
+
+fix env AUTH_URL not working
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @DIYgod !

Thank you for the PR.

It looks good, but we'd like to have a test. I think only the unit test is OK. How about adding the following test to `test/index.test.ts`?

```ts
describe('reqWithEnvUrl()', () => {
  const req = new Request('http://request-base/request-path')
  const newReq = reqWithEnvUrl(req, 'http://auth-url-base/auth-url-path')
  it('Should rewrite the base path', () => {
    expect(newReq.url.toString()).toBe('http://request-base/request-path')
  })
})
```

yusukebe:
Plus, can you run `yarn changeset` at the top-level of the directory to create a changeset? It will be a patch release.

DIYgod:
Hi @yusukebe 

Thanks for the reminder, I have added the test case and changeset.

yusukebe (approved):
LGTM!

yusukebe:
@DIYgod 

Thanks! Shipping now!

PR #481, opened 2024-04-27
fix(auth-js): AUTH_URL not working in getAuthUser

DIYgod:
It is the same URL base issue as #478, sorry for not fully fixing it last time.


yusukebe:
Hi @DIYgod !

It looks good to me, though I may add tests later. Thanks!

PR #488, opened 2024-04-30
fix: deprecated json response method

Frog-kt:
Fixed a part of deprecated response json method in hono since v4.

`c.jsonT -> c.json`

source: [Migration Guide v3.12.x to v4.0.0](https://github.com/honojs/hono/blob/f2985492a30f14f9b2fbeffd5cc6b3aaeb34a2d1/docs/MIGRATION.md?plain=1#L11)


yusukebe:
Hi @Frog-kt 

Thanks. These are good fixes. Merging now.

PR #492, opened 2024-05-02
feat: Introduce WebSocket helper for Node.js | `@hono/node-ws`

nakasyou:
I created WebSocket helper implementation for Node.js

Usage:
```ts
import { createNodeWebSocket } from '@hono/node-ws'
import { Hono } from 'hono'
import { serve } from '@hono/node-server'

const app = new Hono()

const { injectServer, upgradeWebSocket } = createNodeWebSocket({ app })

app.get('/ws', upgradeWebSocket((c) => ({
  // https://hono.dev/helpers/websocket
})))

const server = serve(app)
injectServer(server)
```


yusukebe:
Hi @nakasyou

Awesome! But I think it would be better to move this to https://github.com/honojs/node-server and include it into the `@hono/node-server` package. Then, we can rename it like `@hono/node-server/ws`.

nakasyou:
@yusukebe, thank you for your advice!

I have thought about that, but `@hono/node-server` doesn't have any dependencies.
`@hono/node-ws` depends on `ws`, so I have decided to create WebSocket Helper on honojs/middleware.

What do you think about?

yusukebe:
@nakasyou 

Ahhhh, I see!! Okay, I'll check this later. Thanks!

yusukebe:
@nakasyou 

There are diffs in other packages, not only this package. I've updated the main repo after formatting them. So, could you rebase this PR to the main? Maybe works well. Then, I'll review it.

nakasyou:
@yusukebe, I do it. But it problem couldn't resolve.

yusukebe:
@nakasyou 

> Do I need recreate PR?

Yes, please! You don't have to include any changes in other packages.

yusukebe:
Closing

PR #494, opened 2024-05-03
fix(auth-js): react package work with ssr

divyam234:
@yusukebe  This pr deploy a fix for ssr  in react package  to work with remix and nextjs


yusukebe:
Hi @divyam234 

Okay, we can merge this. Thanks!

Issue #501, opened 2024-05-05
[valibot-validator] Type inference failure in RPC routes when `"query"` contains `number` type(s).


y-hiraoka:
### Description:

When using `vValidator` to validate a query, if the valibot schema includes a number type, the type inference for an RPC route fails.

### Code to Reproduce:

```ts
import { vValidator } from "@hono/valibot-validator";
import { zValidator } from "@hono/zod-validator";
import { hc } from "hono/client";
import { Hono } from "hono";
import { z } from "zod";
import * as v from "valibot";

const routeWithZod = new Hono().get(
  "/zod",
  zValidator(
    "query",
    z.object({
      page: z.coerce.number().int(),
      perPage: z.coerce.number().int(),
    })
  ),
  (c) => c.json({ withZod: true })
);

type RouteWithZod = typeof routeWithZod;

const routeWithValibot = new Hono().get(
  "/valibot",
  vValidator(
    "query",
    v.object({
      page: v.coerce(v.number([v.integer()]), Number),
      perPage: v.coerce(v.number([v.integer()]), Number),
    })
  ),
  (c) => c.json({ withValibot: true })
);

type RouteWithValibot = typeof routeWithValibot;

// inferred route type successfully
const clientWithZod = hc<RouteWithZod>("/");

// failed to infer route type
const clientWithValibot = hc<RouteWithValibot>("/");

```

### Actual Type Inference:

```ts
const clientWithValibot: {
    [x: string]: ClientRequest<{
        [Method: `$${Lowercase<string>}`]: {
            input: Partial<ValidationTargets> & {
                param?: Record<string, string> | undefined;
            };
            output: any;
        };
    }>;
}
```

### Expected Type Inference:

```ts
const clientWithValibot: {
    valibot: ClientRequest<{
        $get: {
            input: {
                query: {
                    page: string | string[];
                    perPage: string | string[];
                };
            };
            output: {
                withValibot: boolean;
            };
        };
    }>;
}
```

zod-validator correctly infers the type. We expect similar behavior.

### package versions:

```json
{
  "name": "hono-validator-example",
  "type": "module",
  "dependencies": {
    "@hono/valibot-validator": "^0.2.3",
    "@hono/zod-validator": "^0.2.1",
    "hono": "^4.2.0-rc.1",
    "typescript": "^5.4.5",
    "valibot": "^0.30.0",
    "zod": "^3.23.6"
  }
}
```

yusukebe:
This was fixed by #502. Thanks!

PR #502, opened 2024-05-05
[valibot-validator] Fix Type Inference for number Types in `valibot-validator` Query

y-hiraoka:
This PR will solves #501.

This includes changes below:

- Update `hono` version: Updated `hono` to the same version as zod-validator.
- Fix type inference: Corrected the type inference in `vValidator` based on the reference implementation in `zod-validator`.
- Add test for `number` query: Added a test case to validate type inference for `number` types in queries.


yusukebe (approved):
LGTM!

Issue #509, opened 2024-05-10
[zod-openapi] Hook type error with async function


taku-hatano:
Hi, team.

Following code occur type error.
(when not async function not occur type error.)

```ts
new OpenAPIHono({
	async defaultHook() {},
});
```

So, I want to add `Promise<void>` to `Hook` type.

PR #510, opened 2024-05-10
fix(zod-openapi): add Promise<void> to Hook

taku-hatano:
fixes https://github.com/honojs/middleware/issues/509


yusukebe:
Hi @taku-hatano 

This is a good fix! I'll merge this and ship it. Thanks!

PR #517, opened 2024-05-14
feat(trpc-server): use c.body to create response to avoid header loss

BlankParticle:
in `trpc-server` middleware, we currently return the response directly which causes the respose to loose information about context. So if we set a header/cookie inside a trpc route, we don't get that in the respose.

I have fixed it by creating a new response object with `c.body` and returning that.

A reproduction of the issue can be found at https://github.com/BlankParticle/hono-header-reproduction/tree/main

### DIFF
--- a/.changeset/chatty-plants-dance.md
+++ b/.changeset/chatty-plants-dance.md
@@ -1,0 +1,5 @@
+---
+'@hono/trpc-server': patch
+---
+
+use `c.body` to create response to avoid header loss
--- a/packages/trpc-server/package.json
+++ b/packages/trpc-server/package.json
@@ -28,11 +28,11 @@
   "homepage": "https://honojs.dev",
   "peerDependencies": {
     "@trpc/server": "^10.10.0",
-    "hono": ">=3.*"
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @BlankParticle 

Thank you for the PR. Can you see the error on the CI?

BlankParticle:
> Hi @BlankParticle
> 
> Thank you for the PR. Can you see the error on the CI?

I didn't get this error while developing. Turns out I was using node 20, I was able to reproduce the errors in node 18. The fix for it is to update hono dep in the package to ` 4.x.x`. Should I update the Hono dependency?

yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

Thanks! Looks good. Updating Hono to `v4.*` is good. I'll merge this now.

codeflows:
Thank you so much, was just struggling with this issue!

codeflows:
It seems setting multiple cookies at once does not work, I've created a PR in the main project to try and fix this.

Issue #432, opened 2024-03-24
Zod OpenAPI: Types export


arsaizdihar:
I want to make wrapper functionality on createRoute that inject 400 errors in every route, but I can't access the internal typing like RouteConfig, ResponseConfig, etc. I suggest to export all the core typings.

flipvh:
Hi @arsaizdihar, sounds interesting. Could this be related to our feature request? https://github.com/honojs/middleware/issues/417

yusukebe:
Hi @arsaizdihar !

Sorry for the delayed response.

The types are exported from [Zod to OpenAPI](https://github.com/asteasolutions/zod-to-openapi) which is used in Zod OpenAPI. So you can use them like the following:

```ts
import type { ResponseConfig, RouteConfig } from '@asteasolutions/zod-to-openapi'
```

PR #532, opened 2024-05-19
feat(zod-openapi): export RouteConfig type

akineko:
#373
#432

I want to create a wrapper function for createRoute.
But, the current RouteConfig differs from the RouteConfig in @asteasolutions/zod-to-openapi.
It would be easy to define it myself, but it would be helpful if it is exported as there may be more differences in the future.


yusukebe:
Hi @akineko 

Exactly, it should be exported. Thanks.

Issue #166, opened 2023-09-19
No "query"-procedure on path "trpc/hello"


mohitxskull:
TRPC routes are there but still getting 404

TURBOREPO

`src/trpc/init.ts`
```ts
import { initTRPC } from '@trpc/server';
import superjson from 'superjson';
import { ZodError } from 'zod';

const t = initTRPC.create({
  transformer: superjson,
  errorFormatter({ shape, error }) {
    return {
      ...shape,
      data: {
        ...shape.data,
        zodError:
          error.cause instanceof ZodError ? error.cause.flatten() : null,
      },
    };
  },
});

export const publicProcedure = t.procedure;

export const TRPCRouter = t.router;

```

`src/trpc/index.ts`
```ts
import { z } from 'zod';
import { publicProcedure, TRPCRouter } from './init';

export const appRouter = TRPCRouter({
  hello: publicProcedure.input(z.string().nullish()).query(({ input }) => {
    return `Hello ${input ?? 'World'}!`;
  }),

  got: publicProcedure.query(() => {
    return `Hello 'World'!`;
  }),

  nest: TRPCRouter({
    hello: publicProcedure.input(z.string()).query(({ input }) => {
      return `Hello ${input}!`;
    }),
  }),
});

export type AppRouter = typeof appRouter;

```

`src/index.ts`
```ts
import { Hono } from 'hono';
import { trpcServer } from '@hono/trpc-server';
import { serve } from '@hono/node-server';
import { renderTrpcPanel } from 'trpc-panel';
import { compress } from 'hono/compress';
import { cors } from 'hono/cors';
import { logger } from 'hono/logger';
import { appRouter } from './trpc';
import { env } from './env';

const app = new Hono();

app.use('*', cors());
app.use('*', compress());
app.use('*', logger());

console.log('Starting server...');

console.log({
  env,
});

app.get('/panel', (c) => {
  return c.html(
    renderTrpcPanel(appRouter, {
      url: 'http://localhost:9879/back/trpc',
      transformer: 'superjson',
    })
  );
});

app.use(
  '/back/trpc/*',
  trpcServer({
    router: appRouter,
  })
);

serve({
  fetch: app.fetch,
  port: 9879,
});

```

```json
{
  "name": "goat",
  "version": "0.0.0",
  "license": "UNLICENSED",
  "private": true,
  "scripts": {
    "start": "tsx src/index.ts",
    "dev": "tsx watch src/index.ts"
  },
  "dependencies": {
[...]

PR #538, opened 2024-05-23
docs: add custom endpoint configuration example

luiisca:
This PR adds documentation for configuring custom endpoints with the @hono/trpc-server middleware. Ensuring the endpoint parameter matches the middleware's path helps avoid issues like the one described in [honojs/middleware#166](https://github.com/honojs/middleware/issues/166).


yusukebe:
@luiisca 

Thanks! Looks good to me. Merging now.

Issue #537, opened 2024-05-22
behind a reverse proxy, X-Forwarded-Proto isn't used and plain http is used for the signin action


bkerin:
I'm running deno behind an apache2 (reverse proxy).  SSL is terminated at the apache2 and it uses plain http to communicate with the deno server.

The system I'm migrating from worked with: `RequestHeader set X-Forwarded-Proto "https"` but this middleware doesn't seem to notice this header, the /api/auth/signin page ends up containing:

```
<form action="http://betatest.mydomain.com/api/auth/signin/google" method="POST">
  <input type="hidden" name="csrfToken" value="8dc4c49dfe160f6a9737d67af01f22e42ba9e71adc7d55ebeecbd57c6076fb15">
  <input type="hidden" name="callbackUrl" value="http://betatest.mydomain.com">
```

I found that I could open the dev console and edit the element to just change that first http to https and then it submits without a warning.

I've got environment vars like so:

```
AUTH_TRUST_HOST=true
AUTH_URL=https://betatest.mydomain.com/api/auth/signin
```

But these apparently aren't enough to do it (the Auth.js docs seem to be saying the second one isn't needed).

Is there some way I'm missing to get https to be used here?

beorn:
I found that although @auth/core supports trustHost (and rewrite the URL based on the headers), it seems @hono/auth-js also needs to do that a few places - otherwise the req.url that's used is wrong - e.g., getAuthUser(), reqWithEnvUrl().   You shouldn't have to set AUTH_URL - the default should work in most cases.  Even if you set AUTH_URL it doesn't work for some reason.

I found by wrapping the request with getXForwardedRequest (or fetch with behindProxy) as per https://jsr.io/@hongminhee/x-forwarded-fetch/0.2.0/mod.ts - it seems to work.

(A related problem is that Bun's websockets system isn't happy with wrapping app.fetch.)

beorn:
Patching https://github.com/honojs/middleware/blob/main/packages/auth-js/src/index.ts to rewrite the request to take into account X-Forwarded-* headers as follows seems to do the trick:

## Patch getAuthUser
https://github.com/honojs/middleware/blob/39edec1249ec3f9dae474b1abc4ebc1cf97a51c2/packages/auth-js/src/index.ts#L69

```ts
  const origin = ctxEnv.AUTH_URL
    ? new URL(ctxEnv.AUTH_URL).origin
[...]

PR #549, opened 2024-05-28
handle x-forwarded headers to detect auth url

divyam234:
@yusukebe  this pr fixes  https://github.com/honojs/middleware/issues/537

### DIFF
--- a/.changeset/funny-dingos-obey.md
+++ b/.changeset/funny-dingos-obey.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': patch
+---
+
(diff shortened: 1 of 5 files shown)


beorn:
Unfortunately, `new Request()` does not copy over things like credentials so I get a CSRF error.  That's why there's all the additional complexity in the `getXForwardedRequest()` function I guess.

beorn:
So this isn't fixed - we unfortunately need the workaround I described here: https://github.com/honojs/middleware/issues/537#issuecomment-2136187408

PR #553, opened 2024-05-30
chore: Update valibot to version 0.31.0

chimame:
## Overview

Update valibot to version 0.31.0.
**0.31.0 will lose backward compatibility due to many breaking changes.**

0.31.0 has not yet been officially released. Therefore, it is a draft PR.

## Details

Changes have been made to correspond to the changes in 0.31.0.

From 0.31.0, the pipe function, which was previously built-in, is now defined as an independent function. This means that each API was able to perform consecutive validations on the second and third arguments, but from 0.31.0 validations will be performed using an independent pipe function.

https://github.com/fabian-hiller/valibot/pull/502

```ts
// 0.30.0 or lower
number("required", [minValue(1, "min"), maxValue(10, "max")])
// 0.31.0 or higher
pipe(number("required"), minValue(1, "min"), maxValue(10, "max"))
```

Also, from 0.31.0 onwards there will be a change in the type definition name.

https://valibot.dev/guides/migrate-from-v0.30.0/


yusukebe:
> 0.31.0 has not yet been officially released. Therefore, it is a draft PR.

yusukebe:
Hi @chimame 

> 0.31.0 has not yet been officially released. Therefore, it is a draft PR.

Okay! Could you ping me if the PR is ready? Thanks!

chimame:
valibot will officially release 0.31.0 this week or next. I'll let you know at that time!

fabian-hiller:
Thank you @chimame! Your changes look good to me.

packages/valibot-validator/src/index.ts
```diff
@@ -56,6 +56,6 @@ export const vValidator = <
       return c.json(result, 400)
```
fabian-hiller, line 56: @yusukebe is there an advantage to returning `result` instead of `result.issues`?

yusukebe: Hi @fabian-hiller 

I think `result` is not bad. And if they do not prefer it, they can change it themselves with the hook.

@chimame Do you have any thoughts?

chimame: I understand that the intention of fabian-hiller's comment is as follows.

If validation fails, wouldn't it be fine to set the HTTP status to 400 and return only the error details in JSON? I think that is what he is trying to suggest.

If validation is successful, you will get `result.output`, in other words, the JSON that passed validation. However, if validation fails, the entire `result` JSON will be returned to the client. The type of `result` is as follows. The type has been simplified for ease of understanding.

```ts
type SafeParseResult = 
  | {
      readonly typed: true;
      readonly success: true;
      readonly output: InferOutput<TSchema>; // <-- Contains objects that have been successfully validated.
      readonly issues: undefined; // <-- Since the validation was successful, there are no issues.
    }
  | {
      readonly typed: false;
      readonly success: false;
      readonly output: unknown; // <-- Contains the original data that failed validation.
      readonly issues: [InferIssue<TSchema>, ...InferIssue<TSchema>[]]; // <-- This contains the details of the errors for each item that failed validation.
    };

const result: SafeParseResult = await safeParseAsync(schema, value);

/*
example error result
{
  typed: false,
  success: false,
  output: { name: 'Superman', age: '20' },
  issues: [
    {
      kind: 'schema',
      type: 'number',
      input: '20',
      expected: 'number',
      received: '"20"',
      message: 'Invalid type: Expected number but received "20"',
      requirement: undefined,
      path: [Array],
      issues: undefined,
      lang: undefined,
      abortEarly: undefined,
      abortPipeEarly: undefined,
      skipPipe: undefined
    }
  ]
}
*/
```

If that is the intention, it is certainly not a bad idea to only return issues.

```suggestion
      return c.json({ issues: result.issues }, 400)
```

fabian-hiller: In general `result` is not bad and can be better because it contains more info, but if this info is never used it is easier for the user to just get the issues.

```ts
return c.json(result.issues, 400)
```

I haven't used Hono with Valibot, so I can't say. It was just a question that came to mind while reviewing the code. I saw that the Zod adapter also returns `result`. We should keep it consistent.

yusukebe: > In general `result` is not bad and can be better because it contains more info, but if this info is never used it is easier for the user to just get the issues.

Yeah, I can understand both reasons. When I created the Zod Validator, I used `result` because I thought more information was better. Shall we go with `result` as it is? Then, we will not have a breaking change.

Thanks for the comments!

fabian-hiller: Yes, let's leave it for now. 👍🏼

yusukebe:
Congrats!

@chimame Can you do the two things?

* `yarn install` to update the lock file.
* `yarn changeset` at the top to create a changeset file. If this has a breaking change, it's better to be `major`. If not, it should be `minor`.

chimame:
@yusukebe 

Sorry, I forgot.
I updated `yarn.lock` and added `CHANGELOG`.

yusukebe:
@chimame Thanks!

Merging now!

PR #561, opened 2024-06-07

nakasyou:
I replaced yarn with Bun, just like [Hono did](https://github.com/honojs/hono/pull/2105).

### Pros
- Yarn creates `node_modules` in all packages, so we need a lot of disc space.
  - Bun creates `node_modules` each packages, but common dependencies are ignored, so we can save disc spaces.
- To install is faster than yarn.


ArjixWasTaken:
same argument can be made for pnpm, which is more mature

yusukebe:
Already migrated to pnpm! Thanks!

PR #563, opened 2024-06-08
Add @hono/react-compat package

usualoma:
The dependency on Hono version is weak and can be used for "4.x", but currently ">= 4.5.*" is specified for `peerDependencies`, as a merged release of https://github.com/honojs/hono/pull/2929 and https://github.com/honojs/hono/pull/2930 would work more fully.

https://github.com/honojs/middleware/compare/main...usualoma:middleware:feat/react-compat?expand=1#diff-290f95d23c372b783139084b0b97bef9100b22c677ab0497121b834c8ca38f86R28

### DIFF
--- a/packages/react-compat/CHANGELOG.md
+++ b/packages/react-compat/CHANGELOG.md
@@ -1,0 +1,1 @@
+# @hono/react-compat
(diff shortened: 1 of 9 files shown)


yusukebe:
Hey @usualoma !

I fixed some issues to work this package correctly myself. I left what I did in the commit messages. Can you check these changes?

If it's okay, we can release it as `v0.0.1` and use it with the hono `v4.5.0-rc.1`, though it is not adaptable to the peer dependencies.

usualoma:
@yusukebe 
Confirmed.
The default export is indeed right, I made a mistake.
Thanks for the correction.

Let's go with this!

yusukebe:
@usualoma Thanks! I'll release this now.

PR #588, opened 2024-06-20
fix(oauth-providers): oauth-middleware: load `env.SECRETS` with hono/adapter

jokester:
Hi, when using the middleware in node.js, I found it does not automatically read from process.env . Use of `hono/adapter` should fix this.

(correction welcome, I'm rather new to honojs)

### DIFF
--- a/.changeset/dull-frogs-glow.md
+++ b/.changeset/dull-frogs-glow.md
@@ -1,0 +1,5 @@
+---
+'@hono/oauth-providers': minor
+---
+
+load env.SECRET with hono/adapter, to support non-worker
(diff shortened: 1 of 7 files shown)


yusukebe:
Hi @jokester 

Thanks for the PR. Could you run `yarn chageset` at the top of the project and push the changeset created? This will be a path change.

jokester:
@yusukebe I think I did it . Please check 🙏🏽

.changeset/dull-frogs-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/oauth-providers': minor
```
yusukebe, line 2: Sorry! I could not tell you correctly. It's a patch change. So rewrite `minor` to `patch`.

jokester: @yusukebe Fixed again 👍🏽

yusukebe (approved):
LGTM!

yusukebe:
@jokester 

Thank you so much! I'll merge this now.

PR #598, opened 2024-06-29
fix(auth-js): bun req cloning

divyam234:
@yusukebe This pr fixes issues on bun. As cloning  request like   `new Request(url,old)` doesn't work on bun yet which causes CRSF issues this pr uses getRuntimeKey from hono adapter to handle cloning for bun separetely.

### DIFF
--- a/.changeset/good-lizards-smile.md
+++ b/.changeset/good-lizards-smile.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': patch
+---
+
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@divyam234 

> As cloning  request like   `new Request(url,old)` doesn't work on bun yet

Nice you found it. Thanks!

PR #599, opened 2024-06-29
 chore: update `@hono/trpc-server` to allow using Trpc 11

BlankParticle:
Updates `@trpc/server` dependency to allow `11.0.0-rc.*` versions too.
~There has not been any changes to the fetch adapter, so no code change is required~

~There has been changes in the serialisation format of Trpc data, the tests were written with trpc 10.~

~The tests needs to be updated for Trpc 11, but that would break testing for trpc 10 (the package works fine with both 10/11, only the tests are the issue)~

According to Trpc's author batch is a boolean which is `1` for true and anything else is false, so the test is faulty. I have updated the tests.

### DIFF
--- a/packages/trpc-server/test/context.test.ts
+++ b/packages/trpc-server/test/context.test.ts
@@ -115,7 +115,7 @@ describe('tRPC Adapter Middleware passing asynchronous Context', () => {
   it.only('Should return 200 response', async () => {
     const searchParams = new URLSearchParams({
       input: JSON.stringify({ '0': 'Hono' }),
-      batch: '2',
+      batch: '1',
     })
     const req = new Request(`http://localhost/trpc/hello?${searchParams.toString()}`)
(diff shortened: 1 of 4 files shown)


packages/trpc-server/test/context.test.ts
```diff
@@ -115,15 +115,15 @@ describe('tRPC Adapter Middleware passing asynchronous Context', () => {
   it.only('Should return 200 response', async () => {
     const searchParams = new URLSearchParams({
       input: JSON.stringify({ '0': 'Hono' }),
-      batch: '2',
```
KATT, line 118: This was always odd, I'm surprised it ever worked ☺

yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

Thanks! Merging now.

PR #601, opened 2024-06-29
feat(oauth-providers): allow custom redirect_uri

jokester:
Previously oauth middlewares only set redirect_uri from `c.req.url` which I felt restrictive:

1. hono (and the server process) may not be able to infer the redirect uri correctly. e.g. when it doesn't know its internet hostname.
2. developer may want to start oauth flow from routes other than the callback path.
3. developer may want to encode something into `redirect_uri`, like another redirect dest after authorization succeeds.

Allowing a custom `redirect_uri` would support these usages. Please see updated middleware readme for an example.

### DIFF
--- a/packages/oauth-providers/src/providers/discord/discordAuth.ts
+++ b/packages/oauth-providers/src/providers/discord/discordAuth.ts
@@ -1,5 +1,5 @@
 import type { MiddlewareHandler } from 'hono'
-import { setCookie, getCookie } from 'hono/cookie'
+import { getCookie, setCookie } from 'hono/cookie'
 import { env } from 'hono/adapter'
 import { HTTPException } from 'hono/http-exception'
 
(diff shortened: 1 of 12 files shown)


jokester:
This may break the following state check and CSRF protection. I need to consider more 🤔 .

jokester:
this change is backward-compatible so I wrote a `minor` changeset. Hope it's the right thing to do.

packages/oauth-providers/src/providers/discord/discordAuth.ts
```diff
@@ -30,14 +31,6 @@ export function discordAuth(options: {
       },
     })
 
-    // Avoid CSRF attack by checking state
```
jokester, line 33: To my understanding the state-based CSRF prevention should check the request with Authorization Code, not the request get redirected to `302 https://the-auth-server.com/oauth`. So moving this check later should be safe. Same for other providers.

Previously it could even be a bug? If a guest visits path `/callback?unrelated_query` I image the oauth flow would fail to start.

jokester:
ready for review 🙏🏽

yusukebe:
Hi @jokester 

Sorry for my late reply.

@monoald Can you review this?

monoald:
Hi @jokester @yusukebe 

So Sorry for all this time, I was in a long vacation

I reviewed this, I think is a necessary change and all seems to work fine, thanks jokester!

LGTM!

yusukebe:
@jokester @monoald Thanks! Let's go with this.

Issue #603, opened 2024-07-01
[@hono/node-ws] Duplicate Event Listeners in Middleware


totoraj930:
In upgradeWebSocket, a 'connection' event listener is being set on `wss`. As the number of new connections increases, the event is called the same number of times.

https://github.com/honojs/middleware/blob/ec6ec4ec02f8db46e3151e7334535e562dfc47e3/packages/node-ws/src/index.ts#L45-L98

In `upgradeWebSocket`, only event listeners ('message', 'close', 'error') should be set on `ws`.

Initially, a new `wss` was created within `upgradeWebSocket`, so this issue did not occur. However, since `wss` is now reused, generated in `injectWebSocket`, this problem has arisen. (https://github.com/honojs/middleware/pull/539)

I attempted to fix the issue, but due to my shallow understanding of the middleware lifecycle, I couldn't succeed. Do you have any ideas?

@nakasyou @mikestopcontinues

inaridiy:
I'll try to fix it a bit.

inaridiy:
I fixed some things in the next PR. Maybe it will work.

https://github.com/honojs/middleware/pull/605

yusukebe:
This was fixed by #605 thanks to @inaridiy !

PR #605, opened 2024-07-01
Bugfix/603 hono node ws duplicate event listeners

inaridiy:
Fixed a bug when multiple connections are created in node-ws, as pointed out in #603 

Summary of changes:
- Fixed bug with multiple connections in node-ws
- Add relevant tests
- Fixed an issue that prevented access to node.js-specific APIs within upgradeWebocket handlers

### DIFF
--- a/packages/node-ws/src/index.ts
+++ b/packages/node-ws/src/index.ts
@@ -52,7 +52,10 @@ export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
           return
         }
         const events = await createEvents(c)
-        wss.on('connection', (ws) => {
+        wss.on('connection', (ws, request) => {
+          if (request !== c.env.incoming) {
+            return
+          }
           const ctx: WSContext = {
             binaryType: 'arraybuffer',
             close(code, reason) {
--- a/packages/node-ws/src/index.ts
+++ b/packages/node-ws/src/index.ts
@@ -3,7 +3,9 @@ import type { Server } from 'node:http'
 import type { Http2SecureServer, Http2Server } from 'node:http2'
 import type { Hono } from 'hono'
 import type { UpgradeWebSocket, WSContext } from 'hono/ws'
+import type { WebSocket } from 'ws'
 import { WebSocketServer } from 'ws'
+import type { IncomingMessage } from 'http'
 
 export interface NodeWebSocket {
   upgradeWebSocket: UpgradeWebSocket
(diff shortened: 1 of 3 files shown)


yusukebe:
@inaridiy Thanks!

@nakasyou Can you review this?

packages/node-ws/src/index.ts
```diff
@@ -49,8 +68,11 @@ export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
           await next()
           return
         }
-        const events = await createEvents(c)
-        wss.on('connection', (ws) => {
+
+        ;(async () => {
+          const events = await createEvents(c)
+          const ws = await nodeUpgradeWebSocket(c.env.incoming)
```
nakasyou, line 74: What do you think about accepting `{ server: { incoming: request, outgoing: undefined } }` as Env?

When user use env as custom value, WebSocket helper may not work.

References: https://github.com/honojs/hono/issues/2645, https://github.com/honojs/hono/issues/2696, https://github.com/honojs/hono/pull/2812, https://github.com/honojs/hono/pull/2595#discussion_r1589946383.

nakasyou (commented):
Thank you for fixing bug, I wrote a review, please check it.

packages/node-ws/src/index.ts
inaridiy: First, there are many theories about the implementation of `{ server: { incoming: request, outgoing: undefined } }`, but I consider this a minimal change to fix a bug.

Secondly, this is just my subjective view.
I see env as an area to store values given by the runtime.
Rather than storing values in the env hypothetically, I think a Variables area and middleware is a more natural way of doing this.

Again, I am not familiar with HonoJS core ideas, so this is subjective.

yusukebe: @nakasyou @inaridiy 

Accepting `{ server: { incoming: request, outgoing: undefined }` for Env is a good idea! But, as @inaridiy said, we don't have to add the change about it to this PR. And you can create another PR after merging this. Shall we go with it?

inaridiy: What do you do about this problem? Once we have discussed how to use env, should I reflect that discussion?

@nakasyou @yusukebe

yusukebe (approved):
LGTM!

packages/node-ws/src/index.ts
yusukebe: Let's merge this first without including Env things!

yusukebe:
@inaridiy 

Looks good! Let's go with this. Thanks!

Issue #398, opened 2024-02-16
how to access env outside fetch, when working with oauth-handlers


KunalSin9h:
```js
  githubAuth({
    client_id: Bun.env.GITHUB_ID,
    client_secret: Bun.env.GITHUB_SECRET,
```

I see, @yusukebe you have used `Bun.env`, but without using `Bun` how can i get the env from .env files.

yusukebe:
Hi @KunalSin9h 

You may use the `env()` function in `hono/adapter`.

https://hono.dev/helpers/adapter#env

nick-cjyx9:
If you're using node.js.
```js
import process from 'node:process'
...
githubAuth({
    client_id: process.env.GITHUB_ID,
    client_secret: process.env.GITHUB_SECRET,
```

yusukebe:
Or you can write the following:

```ts
app.use('*', async (c, next) => {
  await githubAuth({
    client_id: env(c),
    client_secret: env(c),
    //..
  })(c, next)
})
```

PR #612, opened 2024-07-03
docs(oauth-providers): use process.env instead of Bun.env as default

euxn23:
Now Bun is not standard, so I think README should use `process.env` instead of `Bun.env`, so I replaced `Bun.env` to `process.env` in this PR.
But there is a difference between `Bun.env` and `process.env`, I also added note about Bun.

### DIFF
--- a/packages/oauth-providers/README.md
+++ b/packages/oauth-providers/README.md
@@ -28,8 +28,8 @@ Also, there is two ways to use this middleware:
 app.use(
   '/google',
   googleAuth({
-    client_id: Bun.env.GOOGLE_ID,
-    client_secret: Bun.env.GOOGLE_SECRET,
+    client_id: process.env.GOOGLE_ID,
+    client_secret: process.env.GOOGLE_SECRET,
     scope: ['openid', 'email', 'profile'],
   })
 )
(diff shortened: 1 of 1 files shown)


nick-cjyx9:
> Hi @euxn23
> 
> Regarding env variables, we can use the adapter helper:
> 
> https://hono.dev/docs/helpers/adapter#env
> 
> ```ts
> import { env } from 'hono/adapter'
> 
> app.get('/env', (c) => {
>   // NAME is process.env.NAME on Node.js or Bun
>   // NAME is the value written in `wrangler.toml` on Cloudflare
>   const { NAME } = env<{ NAME: string }>(c)
>   return c.text(NAME)
> })
> ```
> 
> I think using this would be better than using other runtime-specific variables.

But `env()` requires a `c: Context` param, which we cannot get out of a handler.

nick-cjyx9:
the same question in #398

yusukebe:
Hi @nick-cjyx9 

> But `env()` requires a `c: Context` param, which we cannot get out of a handler.

You are right. If you want to use environment variables out of a handler, you have to use `Bun.env`, `process.env`, or another runtime-dependent variable. In this case, it will be used in the middleware handler, so we have to use `env()`.

doroved:
> Hi @euxn23
> 
> Regarding env variables, we can use the adapter helper:
> 
> https://hono.dev/docs/helpers/adapter#env
> 
> ```ts
> import { env } from 'hono/adapter'
> 
> app.get('/env', (c) => {
>   // NAME is process.env.NAME on Node.js or Bun
>   // NAME is the value written in `wrangler.toml` on Cloudflare
>   const { NAME } = env<{ NAME: string }>(c)
>   return c.text(NAME)
> })
> ```
> 
> I think using this would be better than using other runtime-specific variables.

Is there any way to have env() automatically parse the .env file and generate types?

yusukebe (approved):

PR #622, opened 2024-07-07
feat: integrate 18next middleware

nrzky:
This middleware adds a translation function (t) for internationalization (i18next) to the context of each request. This function allows you to translate text based on the language preference.


yusukebe:
Hi @nrzky 

Thank you for the PR. I'll take a look later!

PR #632, opened 2024-07-10
feat(zod-openapi): expose `extendZodWithOpenApi` from `zod-to-openapi`

### LINKED ISSUE #631: Expose `extendZodWithOpenApi`
Reading through https://github.com/asteasolutions/zod-to-openapi they mention

> To keep openapi definitions natural, we add an `openapi` method to all Zod objects..
> ...
> For this to work, you need to call `extendZodWithOpenApi` once in your project.
> ...
> Note: This should be done only once in a common-entrypoint file of your project (for example an `index.ts`/`app.ts`). If you're using tree-shaking with Webpack, mark that file as having side-effects.

Rather than having `@hono/zod-openapi` call `extendZodWithOpenApi` and export `z`, I think it would be useful if `@hono/zod-openapi` export the `extendZodWithOpenApi` method and allowed consumers to call it with their own instance of `z`

https://github.com/honojs/middleware/blob/72d591aa359dff50218162c3c479b713c098fe1e/packages/zod-openapi/src/index.ts#L550-L551

Then the usage would look more like..

```ts
import { extendZodWithOpenApi } from '@hono/zod-openapi';
import { z } from 'zod';

extendZodWithOpenApi(z);
```

I hope that makes sense!

paolostyle:
Fixes #631.

Just to add some context why I think it's valuable - I have plenty of zod schemas already and some of them aren't used by Hono at all, but I don't want to be limited and decide whether to use import from the middleware or directly from `zod`, I just want to use the `zod` export everywhere.

It's your call, but I'd even encourage using this method (i.e. using `extendZodWithOpenApi` instead of importing `z`) in the documentation as it gives much more flexibility.

### DIFF
--- a/.changeset/tricky-bugs-switch.md
+++ b/.changeset/tricky-bugs-switch.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+expose `extendZodWithOpenApi` from `zod-to-openapi`
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -548,7 +548,7 @@ export const createRoute = <P extends string, R extends Omit<RouteConfig, 'path'
 }
 
 extendZodWithOpenApi(z)
-export { z }
+export { extendZodWithOpenApi, z }
 
 function addBasePathToDocument(document: Record<string, any>, basePath: string) {
   const updatedPaths: Record<string, any> = {}


yusukebe (approved):
LGTM!

yusukebe:
Hi @paolostyle 

Thank you for your contribution. I'll merge this now and release a new version.

Issue #619, opened 2024-07-05
[@hono/oauth-providers/google] Add ability to pass access_type when requesting a token


aaronware:
Users do not have the ability to set access_type=offline when implementing oauth

aaronware:
Added a pull request for this but I don't entirely understand how I could actually test it when my fork wasn't actually part of the project? I tried to do a `npm link` and that helps me locally but not sure how to test this in the context of a worker and a common workflow. Any shove in the right direction for proper testing would be greatly appreciated.

PR #634, opened 2024-07-10
feat(oauth-providers): Allow for refresh tokens to be used via access_type "offline" in Issue/#619

aaronware:
I might need some support here as I wasn't able to figure out the most efficient way to test my changes based on how this all works within a worker? Any shove in the right direction would be greatly appreciated for proper testing.


aaronware:
I also don't really understand why it had my other pull requests in the commit logs... No changes for those files are actually in this pull request

aaronware:
Closes #619

yusukebe:
Hi @aaronware 

Thanks for the PR. I think it would be fine if you tested it with the `npm link` though we should add more test codes.

@monoald Do you have any thoughts on this?

aaronware:
Thanks @yusukebe , I got pulled away from this onto some items but I should have some time over the next week to confirm it's all working properly within the worker.  @yusukebe I'll keep an eye out for any recommendations/changes you or @monoald have and adjust the PR as needed

Titou325:
Hi! We are also interested in this behaviour to configure long-running sessions (eg. polling GMail on behalf of a user). Right now the flow does not expose the token, but could very easily when correctly configured with prompt: "consent" and access_type: "offline".

It is hard to monkey patch at this time as this is internal to AuthFlow. I don't know if this PR is still under consideration or if I should open a new one to cover this use case.

yusukebe:
Hi @Titou325 

This PR is not progressing, so creating a new PR is okay. If so, I will review it. It would be better if you added proper tests there.

Sama-004:
yeah please allow refresh tokens to be used, right now i've written my oauth using this package, but because of no response token support i'll have to write again everything on my own

yusukebe:
Hi @aaronware 

Can you still work on this PR? I think this feature is okay, but I want to do two things:

* Writing tests
* Update README.md

If these are done, I'll review them!

aaronware:
@yusukebe I can pick this up again in the next week or so

PR #639, opened 2024-07-13
fix(node-ws): binary messages have wrong byteLength

Yovach:
Using the `Buffer.from` method, it creates an 8192-byte buffer (ref. https://nodejs.org/api/buffer.html#class-property-bufferpoolsize) so by using small 8-byte messages in Websocket communication, it creates extra stuff we don't need.

In addition, it eliminates the need to create an unnecessary buffer instance, as `ws` already creates one.

I also added a test for binary messages

### DIFF
--- a/packages/node-ws/src/index.test.ts
+++ b/packages/node-ws/src/index.test.ts
@@ -3,6 +3,7 @@ import type { ServerType } from '@hono/node-server/dist/types'
 import { Hono } from 'hono'
 import { WebSocket } from 'ws'
 import { createNodeWebSocket } from '.'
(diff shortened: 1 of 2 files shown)


yusukebe:
@Yovach 

Thank you for the PR. Can you run `yarn changeset` to create a changeset and push it?

@nakasyou Could you review this?

Yovach:
I set the version bump to `patch`

yusukebe (approved):
LGTM!

yusukebe:
@Yovach @nakasyou Thanks!! Merging now.

PR #640, opened 2024-07-14
fix(graphql-server): update example code to Typescript and parameter name

3w36zj6:
This Pull Request makes the following changes for consistency with other middleware:

- Changed the README example code to TypeScript.
- Renamed the variable representing `Context` from `ctx` to `c`.


yusukebe:
Hi @3w36zj6 

Thanks! Good improvement. Merging now.

PR #645, opened 2024-07-16
Support for other json content-types such as application/vnd.api+json, application/problem+json, etc.

DavidHavl:
While writing a JSON:API compliant REST API using zod-openapi, I ran into an issue where the `c.req.valid('json')` does not contain any data when sending request with Content-Type 'application/vnd.api+json', even though I did specify it in route definition.
So I fixed it.

It now also supports other similar content-types such as:
'application/json'
'application/vnd.api+json'
'application/problem+json'
'application/csp-report+json'
'application/vnd.custom+json'
...

### DIFF
--- a/.changeset/serious-moose-push.md
+++ b/.changeset/serious-moose-push.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

PR #648, opened 2024-07-17
fix(node-ws): `CloseEvent` is not defined

### LINKED ISSUE #595: [@hono/node-ws] ReferenceError: CloseEvent is not defined
I'm using Node `v20.14.0` with @hono/node-ws `v1.0.1` and when listening to the onClose event it throws the following error:
```js
events.onClose?.(new CloseEvent("close"), ctx);
                     ^

ReferenceError: CloseEvent is not defined
```

Yovach:
Currently, `CloseEvent` doesn't exist in the latest versions of Node.js (v18.20.4, v20.15.1 and v22.4.) and you'll get the following error `Uncaught ReferenceError: CloseEvent is not defined` if you have this :

```javascript
app.get(
  '/',
  upgradeWebSocket(() => ({
    onClose() {
      // doing some stuff
    },
  }))
)
```

Fixes #595

### DIFF
--- a/packages/node-ws/src/events.ts
+++ b/packages/node-ws/src/events.ts
@@ -1,0 +1,32 @@
+interface CloseEventInit extends EventInit {
+    code?: number;
+    reason?: string;
+    wasClean?: boolean;
+}
+
+/**
+ * @link https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+ */
+export class CloseEvent extends Event {
+    #eventInitDict
+
+    constructor(
+        type: string,
+        eventInitDict: CloseEventInit = {}
+    ) {
+        super(type, eventInitDict)
+        this.#eventInitDict = eventInitDict
+    }
+
+    get wasClean(): boolean {
+        return this.#eventInitDict.wasClean ?? false
+    }
+
+    get code(): number {
+        return this.#eventInitDict.code ?? 0
+    }
+
+    get reason(): string {
+        return this.#eventInitDict.reason ?? ''
+    }
+}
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @Yovach 

Looks good to me! One thing. Can you run `yarn changeset` on the top of the project and push the changeset?

@nakasyou Can you also review this?

packages/node-ws/src/events.ts
```diff
@@ -0,0 +1,32 @@
+interface CloseEventInit extends EventInit {
+    code?: number;
+    reason?: string;
+    wasClean?: boolean;
+}
+
+/**
+ * @link https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+ */
+export class CloseEvent extends Event {
```
nakasyou, line 10:
```suggestion
export const CloseEvent = globalThis.CloseEvent ?? class extends Event {
```
What do you think about that?
If `CloseEvent` defined, the middleware should use it. This is effective for runtimes where `CloseEvent` is compatible with Node.js is defined.

nakasyou (changes requested):

packages/node-ws/src/events.ts
yusukebe: Seems to be good.

If we do so, we may have to have the CI on both Node.js, which has `CloseEvent` and does not.

Yovach: > What do you think about that? If `CloseEvent` defined, the middleware should use it. This is effective for runtimes where `CloseEvent` is compatible with Node.js is defined.

Nice idea, I added your code to the `events.ts`. Thanks for your suggestion!

yusukebe: > Currently, `CloseEvent` is not exposed to `globalThis` in Node.js 22.5.0 so I think we don't need to have multiple Node.js versions in the CI for now.

Okay! Let's go with it.

yusukebe:
@nakasyou Thank you for your review!

@Yovach I'll merge it into the main and release the new version. Thank you!

Issue #628, opened 2024-07-09
Use Bun as package manager


gunta:
Currently Hono is using Bun as package manager, but the Middleware is not.
https://github.com/honojs/hono/commit/ead367760180437610639637c0eeefe1e5789c9f

For consistency and easiness of contribution, it might be a good idea to use Bun here too.

yusukebe:
Hi @gunta, good suggestion.

Actually, @nakasyou created the PR for the same purpose: https://github.com/honojs/middleware/pull/561

@nakasyou Can you make it ready for review?

nakasyou:
@yusukebe 
With Yarn, each package generates a node_modules that follows the lockfile, including the same version of Hono as when it was first installed. (This consumes a lot of storage.)

In Bun, dependencies are consolidated as much as possible if they are compatible according to each package's package.json.

Therefore, packages tested with a certain version of Hono in Yarn may fail tests in Bun, as the version becomes the latest one within the range specified in package.json.

There are two solutions:

Fix the version dependency on Hono for the problematic packages.
Update the problematic packages to be compatible with the latest Hono.
I have been working on solution 2, but it is difficult for me as I am not familiar with each package.
On the other hand, solution 1 would clearly mean no longer supporting the latest Hono.

yusukebe:
@nakasyou Thank you for your comment and investigation.

Hmmm. I also think the ideal is 2 - following the last `hono`. However, I want to increase the maintenance cost for this middleware project.

nakasyou:
Update Hono version:

- [ ] arktype-validator
- [ ] auth-js
- [ ] bun-transpiler (#661)
- [ ] clerk-auth (#661)
- [ ] effect-validator (#661)
- [ ] esbuild-transpiler (#661)
- [ ] eslint-config
- [ ] event-emitter (#661)
- [ ] firebase-auth
- [ ] graphql-server (#661)
- [ ] ~~hello~~
- [ ] medley-router (#661)
- [ ] node-ws (#661)
- [ ] oauth-providers
- [ ] oidc-auth
- [ ] prometheus (#661)
- [ ] qwik-city (#661)
- [ ] react-compat (#661)
- [ ] react-renderer (#661)
- [ ] sentry (#661)
- [ ] swagger-ui (#661)
- [ ] trpc-server (#661)
- [ ] typebox-validator
- [ ] typia-validator
- [ ] valibot-validator
- [ ] zod-openapi (#661)
- [ ] zod-validator

[...]

PR #661, opened 2024-07-22
chore: update Hono version for some packages

nakasyou:
Reference: #628

I updated Hono version for some packages.

The package below is not included because the test failed.
* auth-js
* arktype-validator
* firebase-auth
* oauth-providers
* oidc-auth
* typebox-validator
* typia-validator
* valibot-validator

Some of these have been resolved with #561, so I will fix it with another PR.


.changeset/ninety-foxes-care.md
yusukebe, line 1: If the change is only upgrading up t`hono` for the `devDependencies`, it does not need to bump the version, and a changeset will not be needed.

nakasyou: I changed it.

nakasyou:
@yusukebe sorry for late reply. Could you check it?

yusukebe:
@nakasyou 

Yes, I'll do it. Wait a bit.

PR #662, opened 2024-07-22
fix(oauth-providers): support latest hono

nakasyou:
In Hono before the update, the argument of c.set was also any, but after the update, I couldn't do it anymore, so I got a TypeScript error. So I fixed it.


.changeset/fluffy-lions-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/oauth-providers': major
```
yusukebe, line 2: This should be a `patch`.

neolectron: @nakasyou any updates on this PR ? If not available I would like to get back to it 👍 .

nakasyou: Sorry for very late response, I fixed it.

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou Thanks!

@neolectron Thanks for pinging.

PR #663, opened 2024-07-22
fix(valibot-validator): update Hono version and fix test code

nakasyou:
I changed the test code to match the new Hono version.


.changeset/plenty-pandas-press.md
yusukebe, line 1: We don't have to add the changeset if the change is only for testing. The changeset should only affect things for users. If you add a changeset, the version will be incremented.

yusukebe:
@nakasyou Thank you for the PR! I've left a comment.

.changeset/plenty-pandas-press.md
nakasyou: I did it.

yusukebe:
@nakasyou 

Thanks. Merging.

PR #666, opened 2024-07-23
feat: Conform Validator Middleware

uttk:
Added a validator that allows validation using [conform](https://conform.guide/). 
Although the processing content is highly abstract and does not do much, it can increase the affinity with  [Hono RPC](https://hono.dev/docs/guides/rpc).

### DIFF
--- a/.changeset/pretty-eels-prove.md
+++ b/.changeset/pretty-eels-prove.md
@@ -1,0 +1,5 @@
+---
+'@hono/conform-validator': patch
+---
+
+Create Conform validator middleware
--- a/.changeset/pretty-eels-prove.md
+++ b/.changeset/pretty-eels-prove.md
@@ -1,5 +1,5 @@
 ---
-'@hono/conform-validator': patch
+'@hono/conform-validator': major
 ---
 
 Create Conform validator middleware
--- a/.github/workflows/ci-conform-validator.yml
+++ b/.github/workflows/ci-conform-validator.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.x
+          node-version: 20.x
       - run: yarn install --frozen-lockfile
       - run: yarn build
       - run: yarn test
--- a/.github/workflows/ci-conform-validator.yml
+++ b/.github/workflows/ci-conform-validator.yml
@@ -1,0 +1,25 @@
+name: ci-conform-validator
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/conform-validator/**'
+  pull_request:
+    branches: ['*']
+    paths:
+      - 'packages/conform-validator/**'
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./packages/conform-validator
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18.x
+      - run: yarn install --frozen-lockfile
+      - run: yarn build
+      - run: yarn test
--- a/packages/conform-validator/package.json
+++ b/packages/conform-validator/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hono/conform-validator",
-  "version": "0.0.1",
+  "version": "0.0.0",
   "description": "Validator middleware using Conform",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/conform-validator/package.json
+++ b/packages/conform-validator/package.json
@@ -47,7 +47,6 @@
     "@conform-to/zod": "^1.1.5",
     "conform-to-valibot": "^1.10.0",
     "hono": "^4.5.1",
-    "rimraf": "^5.0.9",
     "tsup": "^8.2.3",
     "valibot": "^0.36.0",
     "vitest": "^2.0.4",
--- a/packages/conform-validator/package.json
+++ b/packages/conform-validator/package.json
@@ -27,18 +27,18 @@
   },
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
-    "hono": ">=3.9.0",
-    "@conform-to/dom":">=1.1.5"
+    "@conform-to/dom": ">=1.1.5",
+    "hono": ">=4.5.1"
   },
   "devDependencies": {
     "@conform-to/dom": "^1.1.5",
     "@conform-to/yup": "^1.1.5",
     "@conform-to/zod": "^1.1.5",
     "conform-to-valibot": "^1.10.0",
-    "hono": "^4.0.10",
-    "rimraf": "^5.0.5",
+    "hono": "^4.5.1",
+    "rimraf": "^5.0.9",
     "valibot": "^0.36.0",
-    "vitest": "^2.0.3",
+    "vitest": "^2.0.4",
     "yup": "^1.4.0",
     "zod": "^3.23.8"
   }
--- a/packages/conform-validator/package.json
+++ b/packages/conform-validator/package.json
@@ -2,19 +2,30 @@
   "name": "@hono/conform-validator",
   "version": "0.0.0",
   "description": "Validator middleware using Conform",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "types": "dist/esm/index.d.ts",
+  "type": "module",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
   "files": [
     "dist"
   ],
   "scripts": {
     "test": "vitest --run",
-    "build:cjs": "tsc -p tsconfig.cjs.json",
-    "build:esm": "tsc -p tsconfig.esm.json",
-    "build": "rimraf dist && yarn build:cjs && yarn build:esm",
+    "build": "tsup ./src/index.ts --format esm,cjs --dts",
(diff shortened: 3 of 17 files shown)


yusukebe:
Hi @uttk !

Cool! I've tried it the feeling of usage is good. Also, It's great the implementation is simple. Before I review the details, I would like to know @chimame 's thoughts.

chimame:
Before I write my thoughts, let me first say thank you.
This middleware will be very important for Hono to use Form with JSX. I am grateful for creating such a great middleware. 👍 

If I were to create this middleware, I would create it with the same interface. I'm glad we're on the same page.

```ts
import { conformValidator } from '@hono/conform-validator'

conformValidator((formData) => parseWithZod(formData, { schema })),
// parseWithZod or parseWithYup or parseWithValibot
```

This allows middleware users to choose which Validator to use to validate form data, and makes it possible to validate form data with minimal code.

The process of returning errors is defined by the user each time, but how about a method to customize it by allowing an optional hook function to be passed as the second argument of the conformValidator function, just like other validators? If there is no hook function, a process to return the default error content defined by the middleware is necessary.

https://github.com/honojs/middleware/blob/e488b9f08acf6cdb756114831d7aa5e7da3fc8f7/packages/valibot-validator/src/index.ts#L48-L60

Also, although it may be limited to React, I think it would be a good idea to return the Conform `SubmissionResult` type, that is, `submission.reply()`, to the client to pass server-side errors to the client's form data.

```ts
const submission = c.req.valid('form')

if (submission.status !== 'success') {
  const res = c.json({ success: false, result: submission.reply() }, 400)
  throw HTTPException(400, { res })
}
```

uttk:
@yusukebe @chimame 

Thank you for reviews ! 

>The process of returning errors is defined by the user each time, but how about a method to customize it by allowing an optional hook function to be passed as the second argument of the conformValidator function, just like other validators? 

The optional Hook function is a great idea ! 
I would definitely like to implement it !


>If there is no hook function, a process to return the default error content defined by the middleware is necessary.

As for the behavior when the hook function is omitted, I think it would be best to perform only validation, as with the current process, to prevent unintentional transmission of confidential information to the outside.

Also, I think error handling should be done carefully and should be implemented by the implementer himself, rather than leaving it to another library.

uttk:
Hi @yusukebe . 

I added hook option to `conformValidator()`. There is room for discussion regarding error handling and what happens when the hook is omitted, but I think the basic processing has been implemented.

cc: @chimame

chimame:
Thank you for the great implementation.

I think that when there is a validation error in `Conform`, the default behavior should be `return submission.reply()`, as stated in the `Conform` documentation. If users of the middleware want to change this behavior, I think it would be a good idea to use the Hook function you implemented.

https://conform.guide/#the-gist

> I think it would be best to perform only validation, as with the current process, to prevent unintentional transmission of confidential information to the outside.

I was thinking about what kind of assumptions you made in your opinion. I thought that this might be the case depending on the validation schema implementation of middleware users, but if that's the case, I think the same thing could happen with existing validation middleware.

I would like to respect your opinion, so I would like you to tell me in what kind of situations this could occur. I think it will also be useful for yusukebe's decision-making.

uttk:
@chimame 

>I was thinking about what kind of assumptions you made in your opinion. I thought that this might be the case depending on the validation schema implementation of middleware users, but if that's the case, I think the same thing could happen with existing validation middleware.

That's true.

On second thought, I realized that since the other validator middleware already returns error responses by default, it would be more natural for this middleware to match the behavior.

I would like to modify the default behavior so that if a validation error occurs, the default behavior is to return an error response.

Thanks.

uttk:
Fixed it.

chimame (approved):
Thank you for thinking about the API design together with me.

I think there's no problem with this implementation of the API for this middleware. Since it handles FormData, I wonder if we can make good use of `hono/validator`'s FormData processing.

@yusukebe 
I think there's no problem with this as a middleware design, so would you mind reviewing it in detail?

yusukebe:
@uttk Great! I'll do it.

packages/conform-validator/test/valibot.test.ts
yusukebe, line 1: It's better to add tests for the below cases:

```ts
// The request body is blank
const res = await app.request('/author', {
  method: 'POST',
})
expect(res.status).toBe(400)
```

```ts
// The request body is not a FormData
const res = await app.request('/author', {
  method: 'POST',
  body: JSON.stringify({
    name: 'foo',
    age: 'bar',
  }),
})
expect(res.status).toBe(400)
```

Currently, the latter test has not been passed. It return `500`. I think if the content type is not FormData's `getFormDataFromContext` should return a blank `FormData` and throw `400` response.

These are the common tests that do not depend on the validator library. So, it may be good to add `common.test.ts` for these.

packages/conform-validator/test/valibot.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      return c.json({
+        success: true,
+        message: `${value.name} is ${value.age}, nickname is ${
+          value?.nickname || 'nothing yet :3'
+        }`,
+      })
+    }
+  )
+
+  it.skip('check the route object types', () => {
```
yusukebe, line 39: Is this `skip` intended?

packages/conform-validator/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "description": "Validator middleware using Conform",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "vitest --run",
+    "build:cjs": "tsc -p tsconfig.cjs.json",
```
yusukebe, line 13: Please use `tsup` like the `hello` middleware:

https://github.com/honojs/middleware/blob/main/packages/hello/package.json#L13

If it's a hassle for you, I can fix it.

.github/workflows/ci-conform-validator.yml
```diff
@@ -0,0 +1,25 @@
...
+  ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./packages/conform-validator
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18.x
```
yusukebe, line 22: Can you fix it as `20.x` even though some middleware uses `18.x`?

yusukebe:
Hey @uttk 

I've left some comments. Please check them!

@chimame Thank you for your reviews!

packages/conform-validator/test/valibot.test.ts
uttk: OK. I added `common.test.ts`, and fixed to `getFormDataFromContext()` return blank FormData when the content type is not FormData's. 

commit: https://github.com/honojs/middleware/pull/666/commits/5ed20846e8324995394d992ddb206acd24cc8461

packages/conform-validator/test/valibot.test.ts
uttk: No, so I removed it. Thank you for pointing that out.

packages/conform-validator/package.json
uttk: OK. I fixed it with reference to hello middleware.

commit: https://github.com/honojs/middleware/pull/666/commits/e1b58f86145e72fded5c1a4237a8bd01a230cb29

.github/workflows/ci-conform-validator.yml
uttk: OK. I fixed it.

commit: https://github.com/honojs/middleware/pull/666/commits/b345f847bf44375108a37def31c7d3bce415c12f

uttk:
@yusukebe 

Thank you for reviews!
I guess I was able to fix the issues you pointed out.

packages/conform-validator/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "@conform-to/dom": ">=1.1.5",
+    "hono": ">=4.5.1"
+  },
+  "devDependencies": {
+    "@conform-to/dom": "^1.1.5",
+    "@conform-to/yup": "^1.1.5",
+    "@conform-to/zod": "^1.1.5",
+    "conform-to-valibot": "^1.10.0",
+    "hono": "^4.5.1",
+    "rimraf": "^5.0.9",
```
yusukebe, line 50: `rimraf` is no longer needed.

.changeset/pretty-eels-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/conform-validator': patch
```
yusukebe, line 2: There are differences in which version to start with regarding version numbers. Some start `0.0.1`, and some start `1.0.0`.

I think this could go either way and it's okay to release it as a major version. If you are okay with it, can you write `major` instead of `patch`?

packages/conform-validator/test/valibot.test.ts
yusukebe: Good!

yusukebe:
Hi @uttk 

Thank you for some fixes. I've left two comments. There will be last. Check them!

packages/conform-validator/package.json
uttk: Thank you ! I removed it.

.changeset/pretty-eels-prove.md
uttk: OK. I fixed it to update as `major`.

uttk:
@yusukebe 

Thank you for your kind review.
I have corrected it.

yusukebe (approved):
LGTM!

yusukebe:
@uttk @chimame Thank you!

Let's gooo! I'll merge and release it.

PR #676, opened 2024-07-31
Proposal: Implementation of Casbin Middleware

sugar-cat7:
Casbin is a library that simplifies authorization control.(ACL, RBAC, etc...)
While major frameworks like Express and Nest have user-defined middleware available, Hono did not have such an implementation. Therefore, I have created one for Hono.

https://casbin.org/docs/middlewares/


yusukebe:
Hi @sugar-cat7 !

Sorry for the delayed response. I didin't know the Casbin, but looks good.

I have questions, though I may not understand your intention perfectly. Does the `defaultCheckPermission` check only the user name of the Basic Authentication header? Does this mean it does not authorize a user with a password from the header?

I wonder if we can use some logic for Basic Auth in `hono` core. Or is `defaultCheckPermission` just the default, so we don't have to be so picky?

sugar-cat7:
@yusukebe 
Hello! Thank you for checking this!

First of all, as a premise:
Casbin is more of a framework for authorization rather than authentication. Essentially, it controls "who" can do "what" on "which" resources.
Therefore, it is meant to be used in conjunction with authentication (e.g., something like Basic Auth).
(To give a more concrete example, when using Bearer Auth, you can control whether to allow or deny a user's request by comparing a custom claim, such as "role," contained in the JWT after authentication, with a predefined policy.)

Based on the above:

> Does the defaultCheckPermission check only the user name of the Basic Authentication header?

Since this is a framework for authorization, it primarily looks at information about users and their permissions. Regarding defaultCheckPermission, it checks the username and the predefined policy to determine if a user can access a particular endpoint.
(By the way, the reason the default is implemented to look at the Basic Authentication header is to align with the middleware of other frameworks.)

> Does this mean it does not authorize a user with a password from the header?

No. Authentication is assumed to be implemented separately, and defaultCheckPermission looks at whether access should be allowed after the user has been authenticated.

> I wonder if we can use some logic for Basic Auth in hono core. Or is defaultCheckPermission just the default, so we don't have to be so picky?

It seems that the logic in the following link could be used here:

https://github.com/honojs/hono/blob/8ba02273e829318d7f8797267f52229e531b8bd5/src/middleware/basic-auth/index.ts#L16-L33

However, it would be difficult to use the middleware as-is, so it would require re-implementing the same function (called `auth`) that retrieves the user.
Currently, exporting the private function implemented on the original Hono side seems a bit awkward for users, right? (It’s not middleware in the first place...)

yusukebe:
Hi @sugar-cat7 Thank you for the response!

I understood the purpose of Casbin well thanks to your explanation.

> Therefore, it is meant to be used in conjunction with authentication (e.g., something like Basic Auth).
> (To give a more concrete example, when using Bearer Auth, you can control whether to allow or deny a user's request by comparing a custom claim, such as "role," contained in the JWT after authentication, with a predefined policy.)

I think this is ideal and wonderful if we could use this middleware with the current `hono`'s Basic Auth Middleware, Bearer Auth Middleware, or others. Can we do that? If so, I want to see the code base example (it seems you have to change the current code).

> Currently, exporting the private function implemented on the original Hono side seems a bit awkward for users, right? (It’s not middleware in the first place...)

Exactly. We don't welcome exporting the private funcitons.

sugar-cat7:
@yusukebe 
Sorry for the delay.

I’ve made some modifications to the codebase.

The changes are as follows:
- I separated the main `casbin` middleware and the authorizer we provide into different directories (`src/helper`).
- I explicitly passed the `authorizer` to the `casbin` middleware.
- I renamed `defaultPermissionCheck` to `basicAuthorizer`.
- The `basicAuthorizer` now uses the [same logic as the original Hono implementation](https://github.com/honojs/hono/blob/8ba02273e829318d7f8797267f52229e531b8bd5/src/middleware/basic-auth/index.ts#L13-L33). (This is something I’d like to discuss: I’m thinking it might be better to move the `auth` method to `utils`. It seems that in other middleware, non-middleware processes are called from `utils`.)
- I also defined `jwtAuthorizer` in the helper.
- Updated the Readme.

>I think this is ideal and wonderful if we could use this middleware with the current hono's Basic Auth Middleware, Bearer Auth Middleware, or others. Can we do that? If so, I want to see the code base example (it seems you have to change the current code).

I've included the specific usage for combining them in the `README`! Additionally, you might find it easier to understand how to use them together by looking at the test suites for `BasicAuthorizer with hono/basic-auth` and `JWTAuthorizer with hono/jwt`.

yusukebe:
Hi @sugar-cat7 

Thank you for your explanation!

Creating `basicAuthorizer` and `jwtAuthorizer` is good!

> The `basicAuthorizer` now uses the [same logic as the original Hono implementation](https://github.com/honojs/hono/blob/8ba02273e829318d7f8797267f52229e531b8bd5/src/middleware/basic-auth/index.ts#L13-L33). (This is something I’d like to discuss: I’m thinking it might be better to move the `auth` method to `utils`. It seems that in other middleware, non-middleware processes are called from `utils`.)

As you said, it's better to create `auth` function in `utils`. I'm not 100% sure, but I also faced the same issue when I wanted to implement Basic Auth.

The type of `auth` will be good with the following:

```ts
type Auth = (request: Request) => { username: string; password: string }
```

I can do it later, but can you create a PR to refactor like that on `honojs/hono`?

sugar-cat7:
@yusukebe 

> I can do it later, but can you create a PR to refactor like that on honojs/hono?

I have created a PR to move the functions to the utils directory!
https://github.com/honojs/hono/pull/3359

sugar-cat7:
@yusukebe 
I have incorporated the content from hono v4.5.11 and removed any duplicate sections.

package.json
yusukebe, line 1: Change as following to export modules correctly:

```diff
diff --git a/packages/casbin/package.json b/packages/casbin/package.json
index 6d0be5f..da02259 100644
--- a/packages/casbin/package.json
+++ b/packages/casbin/package.json
@@ -3,9 +3,9 @@
   "version": "1.0.1",
   "description": "Casbin middleware for Hono",
   "type": "module",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "types": "dist/esm/index.d.ts",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
   "exports": {
     ".": {
       "import": {
@@ -16,6 +16,16 @@
         "types": "./dist/index.d.cts",
         "default": "./dist/index.cjs"
       }
+    },
+    "./helper": {
+      "import": {
+        "types": "./dist/helper/index.d.ts",
+        "default": "./dist/helper/index.js"
+      },
+      "require": {
+        "types": "./dist/helper/index.d.cts",
+        "default": "./dist/helper/index.cjs"
+      }
     }
   },
   "files": [
@@ -23,7 +33,7 @@
   ],
   "scripts": {
     "test": "vitest --run",
-    "build": "tsup ./src/index.ts --format esm,cjs --dts",
+    "build": "tsup ./src/index.ts ./src/helper/index.ts --format esm,cjs --dts",
     "publint": "publint",
     "release": "yarn build && yarn test && yarn publint && yarn publish"
   },
@@ -48,4 +58,4 @@
     "typescript": "^5.5.3",
     "vitest": "^2.0.1"
   }
-}
+}
```

packages/casbin/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+Let's look at an example.
+Use the `model` and `policy` files from the [Configuration](#configuration) section.
+You can implement a scenario where `alice` and `bob` have different permissions. Alice has access to all methods on `/dataset1/test`, while Bob has access only to the `GET` method.
+
+```ts
+import { Hono } from 'hono'
+import { basicAuth } from 'hono/basic-auth'
+import { newEnforcer } from 'casbin'
+import { casbin } from '@hono/cabin'
```
yusukebe, line 58: It's a typo. Should be `@hono/casbin'`.

packages/casbin/README.md
```diff
@@ -0,0 +1,143 @@
...
+import { basicAuthorizer } from '@hono/cabin/helper'
```
yusukebe, line 59: Also a typo.

packages/casbin/README.md
```diff
@@ -0,0 +1,143 @@
...
+e = some(where (p.eft == allow))
+
+[matchers]
+m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
+```
+
+### Example policy.csv
+
+```csv
+p, alice, /dataset1/*, *
```
yusukebe, line 41: Is the wildcard `*` working correctly?

It does not work but changed like the following, it works well:

```csv
p, alice, /dataset1/test, GET
```

yusukebe:
@sugar-cat7 

Thanks! I've added some comments.

yusukebe:
Sorry, I accidentally closed it.

packages/casbin/README.md
sugar-cat7: Fixed!
https://github.com/honojs/middleware/pull/676/commits/d0325f1b31bb097eb73ae02120dc5cdc74108bd2

packages/casbin/README.md
sugar-cat7: Fixed!
https://github.com/honojs/middleware/pull/676/commits/d0325f1b31bb097eb73ae02120dc5cdc74108bd2

package.json
sugar-cat7: Fixed!
https://github.com/honojs/middleware/pull/676/commits/18b9d2a9214450689a5c8697acabad697b3ed366

packages/casbin/README.md
sugar-cat7: Sorry, the way the Casbin model was written in the README was incorrect.
You need to write the model in a way that allows wildcards, as shown below. (I've fixed it.)

https://github.com/honojs/middleware/pull/676/commits/4749e9d499375406a4d83b1f169e87656df1278b

sugar-cat7:
@yusukebe 
Thank you for the review!
I've made the corrections!

yusukebe (approved):
LGTM!

packages/casbin/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@hono/casbin",
+  "version": "1.0.1",
```
yusukebe, line 3: Please set the version number as `0.0.0`. The changeset will set it automatically.

sugar-cat7: @yusukebe 
Thank you, I fixed it.

yusukebe:
@sugar-cat7 

Thank! Let's go!

sugar-cat7:
@yusukebe 
I'm sorry, it seems like the method I used to resolve the yarn.lock conflict wasn't good, and it's causing the CI to fail. Should I revert it?

PR #695, opened 2024-08-17
feat(zod-validator): pass target from zod-validator to a hook

bartekbp:
Today, when we use `zod-openapi` and pass schema for `body` and `params`, we can pass a hook which is called with the result of the validation. 

The function doesn't get information about the target being validated. This is problematic when we want to return custom error responses depending on the target being validated. For example, one may want to:
-  return status `415 Unsupported Media Type` when body parsing failed due to missing content-type. Such behaviour doesn't make sense for `params`
- include in the error message the place where the validation failed. For `json` we may write in the response body 'body validation failed', for `params` 'path validation failed' and for `query` 'search params validation failed' .

This pull request adds this functionality.

### DIFF
--- a/packages/zod-validator/src/index.ts
+++ b/packages/zod-validator/src/index.ts
@@ -3,8 +3,8 @@ import { validator } from 'hono/validator'
 import type { z, ZodSchema, ZodError } from 'zod'
 
 export type Hook<T, E extends Env, P extends string, O = {}> = (
-  result: { success: true; data: T } | { success: false; error: ZodError; data: T },
-  c: Context<E, P>
+  result: ({ success: true; data: T} | { success: false; error: ZodError; data: T }) & {target: keyof ValidationTargets },
+  c: Context<E, P>,
 ) => Response | void | TypedResponse<O> | Promise<Response | void | TypedResponse<O>>
 
 type HasUndefined<T> = undefined extends T ? true : false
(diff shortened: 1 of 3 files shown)


packages/zod-validator/src/index.ts
```diff
@@ -3,8 +3,8 @@ import { validator } from 'hono/validator'
 import type { z, ZodSchema, ZodError } from 'zod'
 
 export type Hook<T, E extends Env, P extends string, O = {}> = (
```
lucaschultz, line 5: We could the target as another generic here: 

```ts
export type Hook<T, E extends Env, P extends string, Target extends keyof ValidationTargets = keyof ValidationTargets, O = {}> = (
```

we can then pass the `Target` generic of `zValidator` to the hook

```ts
  hook?: Hook<z.infer<T>, E, P, Target>
```

and have the exact target reflected in the hook:

Bildschirmfoto 2024-09-24 um 11 37 29

lucaschultz (commented):
I just made the same change to the middleware. I'm so glad I double-checked and found your PR! I went a slightly different route by adding another generic to the hook (see comments), which creates a narrower type for the target passed to the hook. Would you consider including that change in your PR, @bartekbp?

bartekbp:
Thanks for the suggestion @lucaschultz! I applied it!

yusukebe (approved):
LGTM!

yusukebe:
Hi @bartekbp 

Sorry for the super late reply. Looks good to me. I'll merge this and release a new version now. Thank you!

PR #701, opened 2024-08-20
perf(swagger-ui): use jsdelivr instead of unpkg to serve swagger-ui assets

rwv:
https://6sense.com/tech/content-delivery-network-cdn/jsdelivr-vs-unpkg

In the Content Delivery Network (CDN) market, jsDelivr has a 21.23% market share in comparison to UNPKG's 1.24%.

Also, unpkg is blocked in China while jsdelivr is not

### DIFF
--- a/.changeset/beige-snails-cover.md
+++ b/.changeset/beige-snails-cover.md
@@ -1,0 +1,5 @@
+---
+'@hono/swagger-ui': patch
+---
+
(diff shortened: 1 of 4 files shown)


yusukebe:
@rwv 

I've updated the format and added the changeset. Looks good. Thank you!

Issue #655, opened 2024-07-20
Improve type definitions for arguments in `RootResolver` of graphql-server


3w36zj6:
The argument `c` in `RootResolver`, which represents `Context`, is defined as an optional property. However, I believe there is no need to consider it as `undefined`.
In the current implementation, it is necessary to check whether `c` is `undefined` before using it within `RootResolver`, which I find inconvenient.

https://github.com/honojs/middleware/blob/ab1b831e1a2dd311ce1752e0fa845761515e57fb/packages/graphql-server/src/index.ts#L23-L27

yusukebe:
@3w36zj6 Thank you for the issue.

@metrue Could you take a look?

metrue:
@yusukebe @3w36zj6  
PR is ready for review, please take a look https://github.com/honojs/middleware/pull/705

PR #705, opened 2024-08-22
fix(graphql-server): the resolver always has context as the argument

metrue:
This PR is to improve the type of root resolver function.

Closes  https://github.com/honojs/middleware/issues/655

### DIFF
--- a/packages/graphql-server/src/index.ts
+++ b/packages/graphql-server/src/index.ts
@@ -23,7 +23,7 @@ import { parseBody } from './parse-body'
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type RootResolver<E extends Env = any, P extends string = any, I extends Input = {}> = (
   // eslint-enable-next-line @typescript-eslint/no-explicit-any
-  c?: Context<E, P, I>
+  c: Context<E, P, I>
 ) => Promise<unknown> | unknown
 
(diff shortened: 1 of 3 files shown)


packages/graphql-server/src/index.ts
```diff
@@ -23,7 +23,7 @@ import { parseBody } from './parse-body'
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type RootResolver<E extends Env = any, P extends string = any, I extends Input = {}> = (
   // eslint-enable-next-line @typescript-eslint/no-explicit-any
-  c?: Context<E, P, I>
+  ctx: Context<E, P, I>
```
3w36zj6, line 26: IMO: In PR #640, I changed the name to align with the variable naming conventions used in other projects within the Hono organization. Therefore, I believe it is more appropriate to stick with `c` rather than changing it to `ctx`.

3w36zj6 (commented):
Thank you for addressing the issue I opened.

yusukebe:
Hey @metrue 

Can you run `yarn changeset` to create a changeset and push it?

metrue:
@yusukebe done, please help review it.

3w36zj6 (approved):
LGTM

yusukebe:
@metrue 

I think this change is a patch fix. So, I've updated the changeset and pushed it myself. Let's go with it. Thanks!

PR #709, opened 2024-08-24
feat(oidc-auth): restrict cookie paths

ameinhardt:
I'd like to restrict the cookie path, e.g. to '/api' only. For that, this pull request introduced an envvar OIDC_COOKIE_PATH that defaults back to '/'.
Also, it restricts the path of the callback cookies to the pathname of OIDC_REDIRECT_URI


yusukebe:
Hi @ameinhardt 

Can you add tests for this?

Issue #703, opened 2024-08-22
zod-openapi: mixing mimetypes in response


ameinhardt:
In zod-openapi, when mixing json and other mimetype in one response, typescript refuses to compile:
```
... - error TS2345: Argument of type '(c: Context) => Promise<JSONRespondReturn<{ msg: string; }, 200> | (Response & TypedResponse<string, 200, "text">)>' is not assignable to parameter of type 'Handler<Env, any, {}, MaybePromise<TypedResponse<string | { msg: string; }, 200, "json">>>'.
  Type 'Promise<JSONRespondReturn<{ msg: string; }, 200> | (Response & TypedResponse<string, 200, "text">)>' is not assignable to type 'MaybePromise<TypedResponse<string | { msg: string; }, 200, "json">>'.
    Type 'Promise<JSONRespondReturn<{ msg: string; }, 200> | (Response & TypedResponse<string, 200, "text">)>' is not assignable to type 'Promise<TypedResponse<string | { msg: string; }, 200, "json">>'.
      Type 'JSONRespondReturn<{ msg: string; }, 200> | (Response & TypedResponse<string, 200, "text">)' is not assignable to type 'TypedResponse<string | { msg: string; }, 200, "json">'.
        Type 'Response & TypedResponse<string, 200, "text">' is not assignable to type 'TypedResponse<string | { msg: string; }, 200, "json">'.
          Types of property '_format' are incompatible.
```
Example:
```js
import { OpenAPIHono, z } from '@hono/zod-openapi';
import { revokeSession } from '@hono/oidc-auth';
import { accepts } from 'hono/accepts';
import type { Context } from 'hono';

const logoutMsg = 'You have been successfully logged out!';

export default new OpenAPIHono()
  .openapi({
    method: 'get',
    path: '/auth/logout',
    responses: {
      200: {
        content: {
          'application/json': {
            schema: z
              .object({
                msg: z.string()
              })
              .openapi({
                example: {
                  msg: logoutMsg
                }
              })
          },
          'text/plain': {
            schema: z.string()
              .openapi({
                example: logoutMsg
              })
          }
        },
        description: 'Log out the user'
      }
    }
  }, async (c: Context) => {
    await revokeSession(c);
    const mimeTypes = ['application/json', 'text/plain']
    if (accepts(c, {
      default: mimeTypes[0],
      header: 'Accept',
      supports: mimeTypes
    }) === mimeTypes[0]) {
[...]

PR #710, opened 2024-08-24
feat(zod-openapi): allow multiple mimetype

ameinhardt:
quick fix for #703. Does not differentiate between json-only response and mixed text/json response type, yet.


yusukebe (approved):
LGTM!

PR #711, opened 2024-08-24
feat(oidc-auth): access and set claims

ameinhardt:
I'd like to use '`profile`' scope to get additional claims from google. In order to not having to forward all id claims to the session token, this pull request proposes an "`oidcClaimsHook`" as environment variable. If used, it has access to the claims, actual response (i.e. for retrieving full profile from access server) and returning claims for the session token.

Define environment variable to define oauth scope instead of full `scopes_supported`, e.g.
```
OIDC_SCOPES=openid email profile
```

Define an `oidcClaimsHook` to access original claims and return the ones for session token, e.g.:
```ts
import type { IDToken, OidcAuth, TokenEndpointResponses } from '@hono/oidc-auth';

declare module 'hono' {
  interface OidcAuthClaims {
    name: string
    sub: string
  }
}

const oidcClaimsHook = async (orig: OidcAuth | undefined, claims: IDToken | undefined, _response: TokenEndpointResponses): Promise<OidcAuthClaims> => ({
  name: claims?.name as string ?? orig?.name ?? '',
  sub: claims?.sub ?? orig?.sub ?? ''
}),
...
app.get('/callback', async (c) => {
  c.set('oidcClaimsHook', oidcClaimsHook);
  return processOAuthCallback(c);
})
...
...
```

### DIFF
--- a/.changeset/gorgeous-berries-float.md
+++ b/.changeset/gorgeous-berries-float.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': minor
+---
+
+define custom scope, access oauth response and set custom session claims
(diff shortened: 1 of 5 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@ameinhardt 

I've merged the current main and refactored about type things. Looks good. Let's go with it. Thanks!

PR #742, opened 2024-09-20
fix(tbValidator): Fix tbValidator not calling hook when there's a fail case

Micnubinub:


packages/typebox-validator/src/index.ts
```diff
@@ -72,6 +72,15 @@ export function tbValidator<
       }
       return data
     }
-    return c.json({ success: false, errors: [...Value.Errors(schema, data)] }, 400)
+    
+    const errors = Array.from(Value.Errors(schema, data));
```
yusukebe, line 76: I think it's not necessary to make it as an array. How about using the return object from `Value.Errors()` directly?

```ts
const errors = Value.Errors(schema, data)
```

yusukebe (commented):
Can you add proper tests?

yusukebe:
Hi @Micnubinub 

Thank you for the PR. There are something I want you todo. Can you check the comments?

And can you run `yarn changeset` at the top of this repo to create Changeset? This will be a `minor` change since introducing a new feature.

Micnubinub:
I'm good with it being the generator object, but I only did that to follow your convention here:
image

yusukebe:
@Micnubinub 

> I'm good with it being the generator object, but I only did that to follow your convention here... 

Ah, I see! Let's go with the array.

Micnubinub:
@yusukebe let me know if you want me to change anything

yusukebe (approved):
LGTM!

yusukebe:
@Micnubinub 

Great!! I'll merge it and release a new version. Thank you!

PR #762, opened 2024-10-04
chore: update valibot dependency to v1

fabian-hiller:
I've updated the Valibot dependency of the adapter to v1. We've changed the type signature, so this change is necessary. All other changes in our v1 beta version are only internal.

---

An union of schema library authors (Zod, Valibot, ArkType, ...) collaborating on a standard interface for schema libraries called [Standard Schema](https://github.com/standard-schema/standard-schema). This simplifies implementation and prevents vendor lock-in. So far Valibot v1 supports this spec and Zod will follow soon. In the long run, the explicit implementation of a Valibot and Zod adapter may no longer be necessary.

### DIFF
--- a/.changeset/sharp-ravens-shave.md
+++ b/.changeset/sharp-ravens-shave.md
@@ -1,0 +1,5 @@
+---
+'@hono/valibot-validator': minor
+---
+
+Upgrade Valibot peer dependency to v1.0.0
--- a/packages/valibot-validator/package.json
+++ b/packages/valibot-validator/package.json
@@ -32,12 +32,12 @@
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
     "hono": ">=3.9.0",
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @fabian-hiller 

Ah, Standard Schema seems to be great!

One thing I want you to do is to run the `yarn changeset` and push the changeset. This is a `patch` change. Thank you!

fabian-hiller:
Are you sure that this is a `patch` change? There will be TS errors if devs use Valibot v1 beta with the old adapter. I would release it as a `1.0.0-beta.0` or `v0.4.0`.

yusukebe:
@fabian-hiller 

Ah, I see! Shall we go with `v0.4.0` (minor version up)?

fabian-hiller:
Sounds good to me. We can bumb to v1 once v1 is out.

yusukebe:
@fabian-hiller 

Thanks! Let's go!

PR #783, opened 2024-10-18
[@hono/auth-js] don't use request body

bordeo:
auth-js don't need request body.
accessing request body prevent future usage by other hono handlers

### DIFF
--- a/.changeset/hono-auth-js-stop-use-body.md
+++ b/.changeset/hono-auth-js-stop-use-body.md
@@ -1,0 +1,5 @@
+---
+"@hono/auth-js": patch
+---
+
+[@hono/auth-js] stop using request body
--- a/packages/auth-js/src/index.ts
+++ b/packages/auth-js/src/index.ts
@@ -42,8 +42,6 @@ async function cloneRequest(input: URL | string, request: Request, headers?: Hea
     return new Request(input, {
       method: request.method,
       headers: headers ?? new Headers(request.headers),
-      body:
-        request.method === 'GET' || request.method === 'HEAD' ? undefined : await request.blob(),
       // @ts-ignore: TS2353
       referrer: 'referrer' in request ? (request.referrer as string) : undefined,
       // deno-lint-ignore no-explicit-any


yusukebe:
Hi @bordeo Thank you for the PR.

@divyam234 Can you review this?

divyam234:
@bordeo  request body is always used in login action https://github.com/honojs/middleware/blob/2d1a89df26f2ec42e1dc0e5845cf691bdcbe23e2/packages/auth-js/src/react.tsx#L313 see here.

yusukebe:
@divyam234 Thanks! I'll close this PR now.

bordeo:
I understand your point @divyam234, I'm gonna create a new PR with different approach because if we use `verifyAuth` and we try to read the request body we have the error.

divyam234:
Are you using bun?

bordeo:
yes. the problem isn't the runtime but the fact that request body can be used only once

divyam234:
@bordeo This was added as a workaround for bun as   req cloning was not working  prior to 1.26  but now this can be replaced by simple request cloning like other runtime.You can raise pr for that.

divyam234:
```js
import { test, expect } from "bun:test";

test("should clone request with a readable stream body correctly", async () => {
  const originalUrl = "https://example.com";
  const clonedUrl = "https://example1.com";
  const bodyContent = "test";

  const readableStream = new ReadableStream({
    start(controller) {
      controller.enqueue(new TextEncoder().encode(bodyContent));
      controller.close();
    }
  });

  const req = new Request(originalUrl, {
    method: "POST",
    body: readableStream,
    headers: {
      test: "test"
    }
  });

  const cloneRequest = new Request(clonedUrl, req);

  expect(cloneRequest.url).toBe(clonedUrl);

  expect(cloneRequest.method).toBe(req.method);

  expect(cloneRequest.headers.get("test")).toBe(req.headers.get("test"));

  const clonedBody = await cloneRequest.text();
  expect(clonedBody).toBe(bodyContent);
});
```
I had this test in one of my repo which was faling in earlier versions for bun but its passing on latest version

bordeo:
yes your code is going to work because you are reading the body once with `.text`.
`verifyAuth` calls `getAuthUser`, which calls `reqWithEnvUrl`, which calls `cloneRequest`. However, `cloneRequest` is not using the proper method `.clone` but is instead creating a new Request object with the body read from the original request using `request.blob`.
Once you read from the original request, you can't read it again.
The original request can be used in the handler after `verifyAuth` to read body JSON params, for example, but attempting to read it throws an error because `cloneRequest` has already read it

divyam234:
https://github.com/honojs/middleware/pull/790 this pr will fix this issue

PR #790, opened 2024-10-23
fix(auth.js): clone request directly for bun

divyam234:
As bun now supports request cloning from  version 1.26 onwards  so workaround for cloning can be removed now @yusukebe. Ran test suite using bun its working fine. Also added popup login hook from example repo so that it can be imported directly from here.

### DIFF
--- a/.changeset/slimy-geese-promise.md
+++ b/.changeset/slimy-geese-promise.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': patch
+---
+
+clone request directly for bun
--- a/packages/auth-js/README.md
+++ b/packages/auth-js/README.md
@@ -112,8 +112,7 @@ const useSession = () => {
   return { session: data, status }
 }
 ```
-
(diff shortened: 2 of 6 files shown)


yusukebe (approved):
LGTM!

yusukebe:
> As bun now supports request cloning from version 1.26

Ah, I see! Thank you.

985563349:
@divyam234 @yusukebe 

I found that using the following cloning behavior will result in the inability to obtain c.req.json () in the future:

```typescript
const newReq = new Request(req)
const url = new URL(newReq.url)

// ...

return new Request(url.href, newReq)
```
I tried using the following code and was able to obtain it correctly:

```typescript
const url = new URL(req.url);
const newReq = new Request(url.href, req);

// ...

return new Request(url.href, newReq)
```

985563349:
If it's a bug, I want to submit a PR to fix it.😄

PR #793, opened 2024-10-24
feat(sentry): configure `HTTPException` status codes to report

### LINKED ISSUE #490: Sentry middleware captures HTTPException
Hello! We are throwing an HTTPException in our code. hono/sentry captures this and creates an event in Sentry:

image

Is this the expected behavior? To me it seems that HTTPException shouldn't be captured, since it's actually a "handled" exception, in the sense that we are returning it manually (as an "escape-hatch") because we _expect_ it to occur.

ubinatus:
Closes #490

### Summary
In `@hono/sentry`, adds a configuration so that the user can filter which `HTTPException`s are reported to Sentry.

### Description
Currently, the middleware always reports all `HTTPException`s thrown to Sentry, and there was no way for the user to report only 500 status codes, for example. This results in many issues being unnecessarily reported to Sentry.

This was also discussed in another Sentry package, where they have already added support to filter specific HTTP status codes ([see this](https://github.com/getsentry/sentry-python/discussions/2494#discussioncomment-7507924)).

Since in Hono it's recommended to handle these errors by throwing [HTTPException](https://hono.dev/docs/api/exception), having the ability to filter which statuses are sent to Sentry is very helpful.

### New config
```typescript
  sentry({
    dsn: 'https://xxxxxx@xxx.ingest.sentry.io/xxxxxx',
    // Includes only 500-599 status codes by default.
    includeStatusCodes: [{ min: 500, max: 599 }],
    // Or, you can combine a specific status code and a range.
    // includeStatusCodes: [500, { min: 503, max: 599 }],
    // If you want to exclude all HTTPExceptions, you can set an empty array.
  })
```

By default, if the error is an `HTTPException` instance, only those with status codes between 500 and 599 are reported to Sentry.

### Tests
- Added tests to handle HTTPExceptions through the includeStatusCodes config.
- Added a test to ensure that if no Error is present, it should not report to Sentry (obvious but was missing).
- Added jest.clearAllMocks() before each test run to ensure that all mocks are cleared before each test. This prevents any state from previous tests from affecting the current test (for example, verifying expect(captureException).not.toHaveBeenCalled() after a successful expect(captureException).toHaveBeenCalled() leads to incorrect state).

### DIFF
--- a/.changeset/wet-foxes-eat.md
+++ b/.changeset/wet-foxes-eat.md
@@ -1,0 +1,5 @@
+---
+'@hono/sentry': major
+---
+
+Adds a configuration so that the user can filter which `HTTPException`s are reported to Sentry. By default, if the error is an `HTTPException` instance, only those with status codes between 500 and 599 are reported to Sentry.
--- a/packages/sentry/README.md
+++ b/packages/sentry/README.md
@@ -76,6 +76,24 @@ app.onError((e, c) => {
 })
 ```
 
+### Handling `HTTPException`
+
+You can specify which HTTP exceptions (`HTTPException`) should be captured by using the `includeStatusCodes` option.
+
+```ts
+app.use(
+  '*',
+  sentry({
+    dsn: 'https://xxxxxx@xxx.ingest.sentry.io/xxxxxx',
+    // Includes only 500-599 status codes by default.
+    includeStatusCodes: [{ min: 500, max: 599 }],
+    // Or, you can combine a specific status code and a range.
+    // includeStatusCodes: [500, { min: 503, max: 599 }],
+    // If you want to exclude all HTTPExceptions, you can set an empty array.
+  })
+)
+```
+
 ## Authors
 
 - Samuel Lippert - <https://github.com/sam-lippert>
--- a/packages/sentry/src/index.ts
+++ b/packages/sentry/src/index.ts
@@ -86,7 +86,7 @@ export const sentry = (
         : true
       : false
 
-    if (shouldCapture && c.error) {
+    if (shouldCapture) {
       sentry.captureException(c.error)
     }
   }
--- a/packages/sentry/src/index.ts
+++ b/packages/sentry/src/index.ts
@@ -80,12 +80,13 @@ export const sentry = (
 
     await next()
 
-    const shouldCapture =
-      c.error instanceof HTTPException
+    const shouldCapture = c.error
+      ? c.error instanceof HTTPException
         ? shouldCaptureHTTPException(c.error, includeStatusCodes)
-        : !!c.error
+        : true
+      : false
 
-    if (shouldCapture) {
+    if (shouldCapture && c.error) {
       sentry.captureException(c.error)
     }
   }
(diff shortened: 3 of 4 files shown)


.changeset/wet-foxes-eat.md
yusukebe, line 1: Do you think this change introduces breaking changes?

ubinatus: Hi @yusukebe 👋!

Thought of a breaking change just because this update will make the middleware report `HttpException` to Sentry only for non-client errors (status codes >= 500). Other than that, it does not introduce a breaking change into the code. 

Making the default to send all `HttpException`s to Sentry would have been incongruent with other Sentry plugins, as most of them does not send http status codes between 400-499 by default.

Let me know your thoughts if this should be considered just a minor change. Or if I should change the default config.

PD: Thanks for this amazing framework!

yusukebe:
Hi @sam-lippert!

Can you review this?

.changeset/wet-foxes-eat.md
yusukebe: Thank you for the explanation! Understood well. I also think this should be a major change.

@sam-lippert What do you think of it?

sam-lippert: @yusukebe Regarding the approach to the way this PR solves the issue, is it possible to filter the errors further up the Hono stack, maybe by clearing the error from the context in `app.onError()`? Since this middleware is designed to be as lightweight as possible of an interface to `toucan-js`, it doesn't seem correct to me to add more API surface to it directly.

ubinatus: Hi @sam-lippert . So basically doing the following:

```ts
app.onError((err, c) => {
   ...

  if (err instanceof HTTPException) {
   // your custom logic to clear any particular `HttpException`s
    c.error = undefined
  }

   ...
})
```

image
-
As seen on the logs, no longer gets reported to Sentry.

________________________

Btw, is that the correct way of clearing the error from context?

The only concern with this approach is if you have other middleware (such as metrics, loggers, etc.) that rely on `HttpException` (i.e. the cleared error instance) for their logic.

sam-lippert: @ubinatus I understand the concern about removing an error object. Maybe a better way is to call `c.get('sentry').setEnabled(false)` rather than deleting the error, and re-enabling it at some other point in the app lifecycle.

ubinatus: @sam-lippert Yeah thats a great workaround. Would work as long as it's re-enabled afterwards. 

Thoughts on letting the middleware allow the user opt-out of the automatic `sentry.captureException()`? Anyways, I appreciate your response 🙏 

Will close this PR as it can be solved without the need to introduce `includeStatusCodes`. Also, the user can simply implement it's own toucan-js middleware in case they need more tailored configs.

sam-lippert: >Any thoughts on allowing the middleware to let users opt out of the automatic sentry.captureException()?
Technically, you don't have to use the middleware as a middleware. You can pass a context, blank or custom `next()`, and callback to the middleware with something like `await sentry({ ...config }, (sentry) => sentryInstance = sentry)(c, () => {})` but at that point, you may as well just use `toucan-js`, `@sentry/cloudflare`, or another JS-based Sentry SDK directly.

PR #797, opened 2024-10-28
fix(valibot-validator): add support for v1 dist-tags

fabian-hiller:
Looks like the npm install broke with the upgrade to Valibot v1. This PR should fix that.


yusukebe (approved):
LGTM!

yusukebe:
@fabian-hiller 

Thanks! I'll merge this now.

PR #806, opened 2024-11-05
fix(auth-js): fix cloned request causing request body to be unavailab…

985563349:
I fixed some issues and found that auth-js, when used with zod-validator, would throw: malformed JSON in the request body.

The reason for this is that after cloning the request, the request body is not available in hono/validator via c.req.json.

### DIFF
--- a/.changeset/rare-toys-decide.md
+++ b/.changeset/rare-toys-decide.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': major
+---
+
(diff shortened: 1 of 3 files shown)


yusukebe:
@985563349 Thanks!

@divyam234 Can you review this, if you have a time.

divyam234:
LGTM

yusukebe (approved):
LGTM!

yusukebe:
@985563349 @divyam234 Thanks! Let's go with this.

PR #819, opened 2024-11-10
feat(valibot-validator): upgrade valibot peer dependency

fabian-hiller:
We have improved the [Standard Schema spec](https://github.com/standard-schema/standard-schema). Unfortunately, this leads to a breaking change in the Valibot's type signature. The implementation remains the same.


yusukebe (approved):
LGTM!

yusukebe:
Hi @fabian-hiller 

"Standard Schema Spec" looks good! Thank you. Merging now.

PR #821, opened 2024-11-11
fix: exported Hook type in valibot and typebox validators

MathurAditya724:
- [x] Exported the `Hook` type from these validators similar to other validators implementation like [`@hono/zod-validator`](https://github.com/honojs/middleware/blob/main/packages/zod-validator/src/index.ts#L5) and [`@hono/arktype-validator`](https://github.com/honojs/middleware/blob/main/packages/arktype-validator/src/index.ts#L5)


yusukebe:
Hi @MathurAditya724 !

Thank you for the PR. Looks good but we have to add changesets. Can you run `yarn changeset` on the top of this project? These will be `patch` changes.

MathurAditya724:
Done, please verify that I have correctly added the changeset.

yusukebe:
@MathurAditya724 

Thans! Merging now.

PR #850, opened 2024-11-28
:feat Use honoReq's bodyCache to provide a body function for trpc's req

Gaubee:
Some middleware (in createContext) may need to read the body. If the raw request instance is provided directly, the body will be consumed prematurely, blocking the development of middleware.


yusukebe:
Hi @Gaubee 

This is interesting. It looks good. But one thing. Can you run `yarn changeset` on the top of this repo and push the changeset file? I think this is not a new feature for users, so this will be a `patch` update.

yusukebe (approved):
LGTM!

yusukebe:
@Gaubee 

Thank you! I'll merge this and release the new version.

buwon:
When I apply this version, I get an error.

```
The Request.url getter can only be used on instances of Request
```

The error occurs at the following location.

https://github.com/trpc/trpc/blob/4b7757d9f426ab4eaa2c13153c2b3c5c5557eef5/packages/server/src/adapters/fetch/fetchRequestHandler.ts#L35

It occurs when I call mutate with hono + trpc.
Is the usage changed from before?

yusukebe:
@Gaubee Can you handle it?

Gaubee:
> When I apply this version, I get an error.
> 
> ```
> The Request.url getter can only be used on instances of Request
> ```
> 
> The error occurs at the following location.
> 
> https://github.com/trpc/trpc/blob/4b7757d9f426ab4eaa2c13153c2b3c5c5557eef5/packages/server/src/adapters/fetch/fetchRequestHandler.ts#L35
> 
> It occurs when I call mutate with hono + trpc. Is the usage changed from before?

so you use bunjs?

buwon:
Yes. I use Bun v1.1.38. 
#864 works fine. 
Thank you.

PR #852, opened 2024-11-28
fix(zod-openapi): accept `c.html` when html-like content type specified

### LINKED ISSUE #842: using c.html as response gives error in openapi()
using openai in hono and as we know we can send reponse as html page using `c.html()` and below how writing the `openai()`
config

```ts
import { createRoute, z } from "@hono/zod-openapi";
const router = createRouter().openapi(
  createRoute({
    tags: ["Index"],
    method: "get",
    path: "/",
    responses: {
      200: {
        content: {
          "text/html": { schema: z.object({}) }
        },
        description: " API doc Index"
      }
    }
  }),
  (c) => {
    return c.html('<h1>HTML</h1>');
  }
);

export default router;
```
 this gives types error on the second argument at `(c)`

as below

```txt
Argument of type '(c: Context<AppBindings, "/", {}>) => Response' is not assignable to parameter of type 'Handler<AppBindings, "/", {}, MaybePromise<TypedResponse<{}, 200, string>>>'.
  Type 'Response' is not assignable to type 'MaybePromise<TypedResponse<{}, 200, string>>'.`
```
Screenshot from 2024-11-24 03-07-40

 what would be typing of C here to remove this typing error.

sushichan044:
## Outline
closes #842 

This PR prevents type error when using `c.html` in a handler, if the content type is expected to return HTML.

### DIFF
--- a/.changeset/eighty-doors-destroy.md
+++ b/.changeset/eighty-doors-destroy.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Fix type error when using `c.html` in route expected to return html-like content
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -61,6 +61,14 @@ type IsJson<T> = T extends string
     : never
   : never
 
+type IsHtml<T> = T extends string
+  ? T extends `text/html${infer _Rest}` | `application/xhtml+xml${infer _Rest}`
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @sushichan044 

This is a super great PR! But I have another idea. I'll create a PR soon.

yusukebe:
Hey @sushichan044 

I'll close this since #853 was merged. Thank you!

PR #853, opened 2024-11-28
fix(zod-openapi): return `Response` if response is not text or JSON

### LINKED ISSUE #842: using c.html as response gives error in openapi()

yusukebe:
Fixes #842

#852 was a great PR to fix #842. However, type errors should not arise for responses other than HTML. Example: 
`application/xml`. So, the problem will be solved if we set `Response` to contents other than JSON or Text like the following.

* JSON - `TypedRespose` with `json`
* Text - `TypedResponse` with `text`
* Others - `Response`

This PR will correct whether the returned value should be a `TypedResponse` or a `Response` based on the specified content type.

### DIFF
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -69,6 +69,20 @@ type IsForm<T> = T extends string
     : never
   : never
 
+type ReturnJsonOrTextOrResponse<
+  ContentType,
+  Content,
+  Status extends keyof StatusCodeRangeDefinitions | StatusCode
+> = ContentType extends string
+  ? ContentType extends `application/${infer Start}json${infer _End}`
+    ? Start extends '' | `${string}+` | `vnd.${string}+`
+      ? TypedResponse<Content, ExtractStatusCode<Status>, 'json'>
+      : never
+    : ContentType extends `text/plain${infer _Rest}`
+    ? TypedResponse<Content, ExtractStatusCode<Status>, 'text'>
+    : Response
+  : never
+
 type RequestPart<R extends RouteConfig, Part extends string> = Part extends keyof R['request']
   ? R['request'][Part]
   : {}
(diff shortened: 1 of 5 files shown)


packages/zod-openapi/test/index.test.ts
```diff
@@ -1779,21 +1779,21 @@ describe('RouteConfigToTypedResponse', () => {
             age: number
           },
           200,
-          'json' | 'text'
+          'json'
```
yusukebe, line 1782: Since now the `TypedResponse` only allows `json` with `application/json`, I fixed the test.

packages/zod-openapi/src/index.ts
```diff
@@ -69,6 +69,28 @@ type IsForm<T> = T extends string
...
 
+type ReturnJsonOrTextOrResponse<
+  ContentType,
+  Content,
+  Status extends keyof StatusCodeRangeDefinitions | StatusCode
+> = ContentType extends string
+  ? ContentType extends `application/${infer Start}json${infer _End}`
+    ? Start extends '' | `${string}+` | `vnd.${string}+`
+      ? TypedResponse<
+          SimplifyDeepArray<Content> extends JSONValue
```
yusukebe, line 80: This is following `JSONRespondReturn`, which is introduced in `hono`.

sushichan044:
@yusukebe Looks good!
Conditions for branching are simple and easy to understand.

yusukebe:
@sushichan044 

Thanks! I'll merge this. But I like [your implementation](https://github.com/honojs/middleware/pull/852)!

razzeee:
After this I end up with a never instead of a return type. But I should always return a json https://github.com/flathub-infra/website/pull/4743

Issue #857, opened 2024-11-29
[zod-validator] header always failing validation


lorentzlasson:
given this example server

<details>

```ts
import { Hono } from "hono";
import { zValidator } from "@hono/zod-validator";
import { z } from "zod";

const app = new Hono();

const route = app
  .post(
    "/posts",
    zValidator(
      "json",
      z.object({
        title: z.string(),
      }),
    ),
    zValidator(
      "header",
      z.object({
        apiKey: z.string(),
      }),
    ),
    (c) => {
      console.log(c.req.header('apiKey'))
      const data = c.req.valid("json");
      const header = c.req.valid("header");
      return c.json({
        data,
        header,
      });
    },
  );

export type AppType = typeof route;

Deno.serve({ port: 8787 }, app.fetch);
```
</details>

and this client

<details>

```ts
import { AppType } from "./main.ts";
import { hc } from "hono/client";

const client = hc<AppType>("http://localhost:8787");

const res = await client.posts.$post({
  json: {
    title: "Hello"
  },
  header: {
    apiKey: "supersecret",
  },
});

const data = await res.json();
console.log(data);
```
</details>

the type checker is happy but i get a validation error even though `apiKey` is provided in the header (`supersecret` gets logged as expected on line 23 in server if the validation is disabled)

```ts
$ deno task test
Task test deno --allow-net client.ts
{
  success: false,
  error: {
    issues: [
      {
        code: "invalid_type",
        expected: "string",
        received: "undefined",
        path: [ "body" ],
        message: "Required"
      }
    ],
    name: "ZodError"
  }
}
```

using deno. this is the the [full example](https://github.com/lorentzlasson/hono-repro).

`ValidationTargets`  json, query and param seems to work as expected. is there something special about header validation i need to think about?

askorupskyy:
Seems like the headers are not being passed on Deno. Tried both `header` along with `json` like in the example, and `headers` separately.

askorupskyy:
Feels like a Deno-specific error. Using `zod-validator` for `headers` in node works fine. Could you try using something like axios as your `fetch` to check if that works?

askorupskyy:
I initially thought this was an issue with Deno. Turns out request headers are always being casted to lower case, so your zod schema should check for `apikey` and not `apiKey`. Hope this helps!

askorupskyy:
> When c.req.header() is called with no arguments, all keys in the returned record are lowercase.

[...]

PR #860, opened 2024-11-30
fix(zod-validator): support for case-insensitive `headers` target validation

askorupskyy:
Issue explained [here](https://github.com/honojs/middleware/issues/857#issuecomment-2509167288).

While it does make sense to keep `headers` case-insensitive (aka always casting to lowercase), it's easy to miss when writing your own validation schema. My little implementation makes sure the Zod schema doesn't care for lower/upper casing of the headers and treats them the same.

Fixes the issue above, with included test-cases.

### DIFF
--- a/packages/zod-validator/src/index.ts
+++ b/packages/zod-validator/src/index.ts
@@ -1,6 +1,6 @@
-import type { Context, MiddlewareHandler, Env, ValidationTargets, TypedResponse, Input } from 'hono'
+import type { Context, Env, Input, MiddlewareHandler, TypedResponse, ValidationTargets } from 'hono'
 import { validator } from 'hono/validator'
-import type { z, ZodSchema, ZodError } from 'zod'
+import { ZodObject, type ZodError, type ZodSchema, type z } from 'zod'
 
 export type Hook<
   T,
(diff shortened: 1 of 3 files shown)


packages/zod-validator/src/index.ts
```diff
@@ -46,10 +46,26 @@ export const zValidator = <
...
-    const result = await schema.safeParseAsync(value)
+    let validatorValue = value
+
+    // in case where our `target` === `header`, Hono parses all of the headers into lowercase.
+    // this might not match the Zod schema, so we want to make sure that we account for that when parsing the schema.
+    if (target === 'header' && schema instanceof ZodObject) {
+      // create an object that maps lowercase schema keys to lowercase
+      const schemaKeys = Object.keys(schema.shape)
+      const caseInsensitiveKeymap = Object.fromEntries(
+        schemaKeys.map((key) => [key.toLowerCase(), key])
```
yusukebe, line 57: It is a very very slight difference, but `toLowerCase()` takes longer to execute. Therefore, this PR will cause performance degradation. We would like to avoid that as much as possible. Also, we may need to add the same procedure to other validators if we will add this feature for the Zod Validator.

What do you think about these?

We can merge this PR if we can allow for these pain points.

But if not so, for example, could we encourage documents to use only lowercase? We wrote about it in the `Validation` section of the website. https://hono.dev/docs/guides/validation

askorupskyy: I think the issue with documentation is it's very easy to miss. For example, I've been using Hono for over 2 years at this point and never knew headers are casted to lower case if you retrieve them through `c.req.headers`...

As a runtime-performant solution I might have an idea. What if restrict the schema keys for `headers` target to only lowercase characters? Is that something that could work?

yusukebe: > What if restrict the schema keys for `headers` target to only lowercase characters using TypeScript? Is that something that could work? This way it would just let you know in an IDE that you have a problem with your headers validation

This is a super interesting idea. We have to find good ways.

askorupskyy:
@yusukebe like i just noticed in #884, people get confused about this. I was not able to find a pure TS approach to this, and tbh it has its performance drawbacks as well.

Here's my 5 cents on the performance issue: people using Zod will not notice the difference in performance with this PR as Zod is extremely slow anyways. Additionally, the TS solution I proposed before adds type complexity, which will make the already significant IDE performance issue even worse.

Ultimately it's up to you, but I feel like this solution will make things a bit more clear for the general public using this framework. I guess this is the question of tradeoffs between the DX and speed.

yusukebe (approved):
LGTM!

yusukebe:
Hi @askorupskyy !

Thank you for spending time. As you said, the overhead to convert header values lowercase will be very slight. So, this PR is okay. Let's go with this.

PR #864, opened 2024-12-02
fix(trpc-server) The Request.url getter can only be used on instances of Request

Gaubee:
wrap Request with Proxy getter receiver should Request-instance


yusukebe (approved):
LGTM!

yusukebe:
@Gaubee 

LGTM! Honestly, I think we should write more tests for this tRPC Server Middleware, including a test covering this fix. But now, I'll merge this change first. Thanks!

PR #869, opened 2024-12-04
fix(zod-validator): resolve module support issue in Zod validators

mofareedd:
Description:
When importing @hono/zod-validator in a ESM environment, the following error occurs:
```
The current file is a CommonJS module whose imports will produce 'require' calls;
however, the referenced file is an ECMAScript module and cannot be imported with 'require'.
```
This issue arises because the CommonJS and ESM entry points in the package are not correctly configured in package.json. This misconfiguration causes compatibility issues when the library is used in different module systems.

This PR resolves the compatibility issue by updating the entry points in package.json:
- Set the CommonJS entry point to `dist/index.js`
- Set the ESM entry point to `dist/index.mjs`

`package.json`
```json
{
  "main": "dist/index.js",
  "module": "dist/index.mjs",
}

```

### DIFF
--- a/packages/zod-validator/package.json
+++ b/packages/zod-validator/package.json
@@ -1,48 +1,46 @@
 {
-  "name": "@hono/zod-validator",
-  "version": "0.4.1",
-  "description": "Validator middleware using Zod",
-  "type": "module",
-  "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "dist/index.d.ts",
-  "exports": {
-    ".": {
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
-    }
-  },
-  "files": [
-    "dist"
-  ],
-  "scripts": {
-    "test": "vitest --run",
-    "build": "tsup ./src/index.ts --format esm,cjs --dts",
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @mofareedd 

The current setting may be correct. It does not create `dist/index.mjs` for ESM but creates `dist/index.js` as ESM and `dist/index.cjs` as CommonJS:

```
yusuke $ ls -l dist
-rw-r--r--@ 1 yusuke  staff  1698 12  5 17:01 index.cjs
-rw-r--r--@ 1 yusuke  staff  1252 12  5 17:01 index.d.cts
-rw-r--r--@ 1 yusuke  staff  1252 12  5 17:01 index.d.ts
-rw-r--r--@ 1 yusuke  staff   666 12  5 17:01 index.js
```

Can you confirm again that your project is correct?

OpenSource03:
Also getting this issue.

Adrianmjim:
It happens to me too. It seems like an issue with the package exports.

Adrianmjim:
@mofareedd @OpenSource03 This issue was fixed at version `0.4.3`

yusukebe:
Yes. This issue was fixed. I'll close the PR. Thank you!

PR #876, opened 2024-12-09
fix(node-ws): websocket upgrade header value case sensitivity

### LINKED ISSUE #875: WebSocket connection failing due to "Upgrade" header value case sensitivity
I'm encountering an issue with an iOS client that fails to establish a WebSocket connection to my Hono server.  The problem lies in the `upgradeWebSocket` function, which is [unnecessarily case-sensitive to the value of the `Upgrade` header](https://github.com/honojs/middleware/blob/6524a92a9e2216f8bf196849720b9c3da12413f0/packages/node-ws/src/index.ts#L66-L70). Specifically, the client sends the value `WebSocket` as part of the handshake, and the server rejects it.

```
 async function upgradeWebSocket(c, next) {
  if (c.req.header('upgrade') !== 'websocket') {
    // Not websocket
    await next()
    return
  }
```

However per [RFC 6455 Section 4.2.1](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1) this header's value should be treated as case-insensitive:

> 3.   An |Upgrade| header field containing the value "websocket",
treated as an ASCII case-insensitive value.

HusamElbashir:
Fixes #875

### DIFF
--- a/.changeset/cool-coats-try.md
+++ b/.changeset/cool-coats-try.md
@@ -1,0 +1,5 @@
+---
+'@hono/node-ws': patch
+---
+
+Fixed case-sensitivity of the WebSocket "Upgrade" header value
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@HusamElbashir 

Thanks. It may be better to add a test for this change, but it's a small change, and the current test is passed, so it's okay not to have a test for this PR. I'll merge this now.

PR #883, opened 2024-12-12
feat: GitHub Webhooks middleware

oscarvz:
Introduces middleware that allows you to receive and process GitHub Webhooks events in your Hono application.
It conveniently validates the incoming requests and provides you with a simple API to handle the events.

## Usage
```ts
import { Hono } from 'hono'
import { gitHubWebhooksMiddleware } from '@hono/github-webhooks'

type Env = {
  GITHUB_WEBHOOK_SECRET: string
}

const app = new Hono<{ Bindings: Env }>()

app.use('/webhook', gitHubWebhooksMiddleware())

app.post('/webhook', async (c) => {
  const webhooks = c.get('webhooks') /* const webhooks = c.var.webhooks */

  webhooks.on('star.created', async ({ id, name, payload }) => {
    // handle star creation here
  })

  webhooks.on('issues.opened', async ({ id, name, payload }) => {
    // handle opened issue here
  })
})
```

---

TODO
- [ ] update https://github.com/honojs/website/blob/main/docs/middleware/third-party.md

### DIFF
--- a/.changeset/nice-rocks-lay.md
+++ b/.changeset/nice-rocks-lay.md
@@ -1,0 +1,5 @@
+---
+'@hono/github-webhooks': major
+---
+
+GitHub Webhooks middleware that allows for listening to Webhooks in your Hono app.
--- a/.github/workflows/ci-github-webhooks.yml
+++ b/.github/workflows/ci-github-webhooks.yml
@@ -1,0 +1,24 @@
+name: ci-github-webhooks
+on:
+  push:
+    branches: [main]
+    paths:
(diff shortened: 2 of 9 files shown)


yusukebe:
Hi @oscarvz 

Sorry for the late reply. Looks good. I think it will be helpful if you write this as an example on "Examples" in our doc like the "Stripe Webhook": https://hono.dev/examples/stripe-webhook

Maintaining middleware is harder than updating documentation, so I think it's better to make it a document than adding the middleware to this repo.

PR #887, opened 2024-12-13
feat(standard-validator): Add standard schema validation

muningis:
Group of different Schema Validation libraries authors has agreed on a standard schema, so other libraries, don't have even to care what validation library you're using, as typings and output _must_ match spec.


EdamAme-x:
great!!

yusukebe:
Hi @muningis 

This is awesome! The code below actually works!

```ts
import { Hono } from 'hono'
import { sValidator } from '@hono/standard-validator'
import * as v from 'valibot'
import { z } from 'zod'

const aSchema = type({
  agent: 'string',
})

const vSchema = v.object({
  slag: v.string(),
})

const zSchema = z.object({
  name: z.string(),
})

const app = new Hono()

app.get(
  '/:slag',
  sValidator('header', aSchema),
  sValidator('param', vSchema),
  sValidator('query', zSchema),
  (c) => {
    const headerValue = c.req.valid('header')
    const paramValue = c.req.valid('param')
    const queryValue = c.req.valid('query')
    return c.json({ headerValue, paramValue, queryValue })
  }
)

const res = await app.request('/foo?name=foo', {
  headers: {
    agent: 'foo',
  },
})

console.log(await res.json())

/**
{
  headerValue: {
    agent: "foo",
  },
  paramValue: {
    slag: "foo",
  },
  queryValue: {
    name: "foo",
  },
}
 */
```

I'll review this. First, can you add the CI like this: https://github.com/honojs/middleware/blob/main/.github/workflows/ci-hello.yml

packages/standard-validator/test/__schemas__/valibot.ts
```diff
@@ -0,0 +1,46 @@
+import { object, string, number, optional, enum_, pipe, unknown, transform, union, const, picklist } from 'valibot'
```
yusukebe, line 1: `union` and `const` are unnecessary.

```diff
diff --git a/packages/standard-validator/test/__schemas__/valibot.ts b/packages/standard-validator/test/__schemas__/valibot.ts
index 58d83fa..6bd2812 100644
--- a/packages/standard-validator/test/__schemas__/valibot.ts
+++ b/packages/standard-validator/test/__schemas__/valibot.ts
@@ -1,4 +1,4 @@
-import { object, string, number, optional, enum_, pipe, unknown, transform, union, const, picklist } from 'valibot'
+import { object, string, number, optional, pipe, unknown, transform, picklist } from 'valibot'

 const personJSONSchema = object({
   name: string(),
```

packages/standard-validator/package.cjs.json
yusukebe, line 1: Are the files below necessary?

* `package.cjs.json`
* `tsconfig.cjs.json`
* `tsconfig.esm.json`

`tsup` can build both cjs and esm properly.

muningis:
Hy! Sure, will do it later today once I'm back from work.

yusukebe:
Hey @fabian-hiller ! Can you review this?

fabian-hiller:
Yes, I will review it later. Make sure to add Hono to the Standard Schema ecosystem list once this is merged. 😎

packages/standard-validator/test/__schemas__/valibot.ts
muningis: Adjusted.

packages/standard-validator/package.cjs.json
muningis: Indeed. Leftover from when I initially used vite.

.changeset/modern-bugs-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/standard-validator': minor
+---
+
+Initial implementation for standar schema support
```
fabian-hiller, line 5: Typo at "standar" and I recommend to capitalize Standard Schema (but both is not too important)

packages/standard-validator/README.md
```diff
@@ -0,0 +1,46 @@
+# Standard Schema validator middleware for Hono
+
+The validator middleware using [Standard Schema Spec](https://github.com/standard-schema/standard-schema) for [Hono](https://honojs.dev) applications.
+You can write a schema with any validation library supporting Standard Schema and validate the incoming values.
+
+## Usage
+
+```ts
+import { z } from 'zod'
+import { sValidator } from '@hono/standard-schema-validator'
```
fabian-hiller, line 10: I think there is a mismatch in package names. Here `@hono/standard-schema-validator` is used but in the package.json it is named `@hono/standard-validator`.

packages/standard-validator/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@hono/standard-validator",
+  "version": "0.0.0",
```
fabian-hiller, line 3: You may want to do this later, but this should probably be bumped to `0.1.0`.

packages/standard-validator/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/honojs/middleware.git"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "@standard-schema/spec": "1.0.0-beta.4",
```
fabian-hiller, line 37: Please use the latest RC version of the spec.

packages/standard-validator/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@standard-schema/utils": "0.3.0",
```
fabian-hiller, line 38: We are not sure yet if we will keep the `utils` package. Feel free to give us feedback in the long run to help us make a better decision.

packages/standard-validator/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  hook?: Hook<StandardSchemaV1.InferOutput<Schema>, E, P, Target>
+): MiddlewareHandler<E, P, V> =>
+  // @ts-expect-error not typed well
+  validator(target, async (value, c) => {
+    let validatorValue = value
+
+    /**
+     * For headers, all keys are lower-cased, so we have to transform them.
+     */
+    if (target === 'header' && isStandardSchemaValidator(schema) && schema['~standard'].types) {
```
fabian-hiller, line 61: `schema['~standard'].types` is always `undefined` at runtime because we just store type information here. This implementation should probably be changed or removed.

packages/standard-validator/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+          return hookResult.response
+        }
+      }
+    }
+
+    if (result.issues) {
+      return c.json({ data: validatorValue, error: result.issues, success: false }, 400)
+    }
+
+    return result.value as StandardSchemaV1.InferOutput<Schema>
```
fabian-hiller, line 96: Is type casting necessary here? `result.value` should already be typed after we know `result.issues` is `undefined`.

fabian-hiller (changes requested):
Once Standard Schema is accepted by the JS community, Hono could try to implement it natively to remove the adapter altogether.

packages/standard-validator/package.json
fabian-hiller: And I would probably add it as a direct dependency and not as under `peerDependencies` and `devDependencies`.

.github/workflows/ci-standard-validator.yml
```diff
@@ -0,0 +1,25 @@
...
+  ci:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./packages/standard-validator
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18.x
```
yusukebe, line 22: Please replace `18.x` with `20.x`? Some other CI uses `18.x`, but currently `20.x` is better.

packages/standard-validator/package.json
yusukebe: This `0.0.0` is okay since the Changesets will bump the version automatically!

packages/standard-validator/package.json
muningis: @yusukebe WDYT? For example, we have same with other validators, such as zod or valibot.

packages/standard-validator/src/index.ts
muningis: Hm, I guess tests has issues then if they passed. Will have to adjust. 
And it will have to be changed, as we need to re-write keys of a schema/shape, so that all keys are lower-cased.

packages/standard-validator/src/index.ts
muningis: Unfortunately, yeah. (Same is required by zod and valibot validators).

It comes down to some type issues coming from `hono/validator` (haven't had chance to dive in yet).

packages/standard-validator/package.json
fabian-hiller: To clarify, when using the Hono Valibot adapter, you already have Valibot installed, otherwise the adapter makes no sense. With Standard Schema this is different, as any library that supports the spec is compatible. Users will never install Standard Schema themselves. Note that Standard Schema contains only types and no runtime code.

fabian-hiller: I have heard that dev dependencies may not be installed when a package is installed. So adding it as a dependency might be safer. But feel free to investigate and report back.

packages/standard-validator/src/index.ts
fabian-hiller: This is not possible. Standard Schema does not contain an AST-like structure. Why is it necessary to rewrite the keys? Shouldn't users take care of this when writing the schema?

packages/standard-validator/package.json
yusukebe: @muningis 

I think you should add `@standard-schema/spec` as a direct dependency, as @fabian-hiller pointed out. It is used in `packages/standard-validator/src/index.ts`, so it's natural to do it.

packages/standard-validator/src/index.ts
muningis: Hm, looks like my tests ain't correct. Other validators at the moment, rewrites headers to make them lowercase. @yusukebe WDYT, should we perhaps just expect users to write headers in lowercase in validation?

yusukebe: @muningis 

I think you already know that the header keys are always lowercase, so it should have a map of actual keys and lower keys. This is introduced in #860

If creating the map with Standard Schema is challenging, we should expect users to write them in lowercase. This is a problem with the specification that `Headers` always keep the keys in lowercase, and we Hono-side don't need to force implement it. I think it would be enough to inform users to write the keys for validation in lowercase in the documentation.

muningis: Upon further inspection:
1. Nor valibot nor arktype validators handles lower-casing headers.
2. From short glance, with arktype that wouldn't even be possible, as they create a function for validation.

I will drop this, and update readme to contain info.

yusukebe: @muningis 

Thank you! Let's go with that.

yusukebe (approved):
LGTM!

yusukebe:
Hey @fabian-hiller. This looks good to me.  Can you approve this?

packages/standard-validator/package.json
```diff
@@ -0,0 +1,51 @@
...
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "tsc --noEmit && vitest --run",
```
yusukebe, line 20: I added the `tsc` command to test TypeScript types.

packages/standard-validator/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./",
+  },
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
```
yusukebe, line 8: To test TypeScript types, you need to add this line.

packages/node-ws/package.json
yusukebe, line 1: It was accidentally updated by `yarn` command, but I left it changed.

packages/standard-validator/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+            ? In
+            : { [K2 in keyof In]: ValidationTargets[K][K2] }
+        }
+    out: { [K in Target]: Out }
+  },
+  V extends I = I
+>(
+  target: Target,
+  schema: Schema,
+  hook?: Hook<StandardSchemaV1.InferOutput<Schema>, E, P, Target>
```
fabian-hiller, line 52: Should this be `Hook<Schema, E, P, Target>` instead? At least the Valibot validator middleware implementation is using the schema object and not its output.

packages/standard-validator/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        if ('response' in hookResult) {
+          return hookResult.response
+        }
+      }
+    }
+
+    if (result.issues) {
+      return c.json({ data: value, error: result.issues, success: false }, 400)
```
fabian-hiller, line 77: This is probably not a problem in this implementation, but the Valibot validator middleware uses `output` instead of `data` for the output value. Could it be that the Valibot implementation should be updated?

fabian-hiller (commented):
Rest LGTM

packages/standard-validator/src/index.ts
muningis: Valibot and Zod has different `Hook` implementation, this one is based on Zod.

packages/standard-validator/src/index.ts
muningis: Hm. Does valibot transform values if it fails to validate? Personally, I think, ideally, we should return whatever was passed, before transformations and defaults, in case of failed validation.

packages/standard-validator/src/index.ts
fabian-hiller: @yusukebe should this be uniformed?

packages/standard-validator/src/index.ts
fabian-hiller: Yes, Valibot always provides an output that can contain transformations when using `safeParse` or `safeParseAsync`. I have no opinion on the implementation. I just noticed it.

muningis: Working with original data IMO has slightly more better handling if there’s something you need to do what’s not available/possible with schema and/or want to use it for observability. Would be nice to hear what @yusukebe has to say on this.

packages/standard-validator/src/index.ts
yusukebe: > @yusukebe should this be uniformed?

It's better to unify the implementation. Following Zod Validator is good. The current implementation in this PR is okay, but we should change the implementation of the Valibot Validator.

muningis:
Anything else we should do, or this is good to go?

yusukebe:
This is good! Let's land it.

@muningis Thank you for the great work. Standardizing the schema libs for the validator will be a big thing for Hono and the TS/JS ecosystem!

@fabian-hiller Thank you for the review!

Eghizio:
Does it work with OpenAPI like [@hono/zod-openapi](https://hono.dev/examples/zod-openapi) or is it planned to?

muningis:
@Eghizio - `@hono/zod-openapi` depeneds on `@asteasolutions/zod-to-openapi`. 

From my understanding, `@asteasolutions/zod-to-openapi` relies on internals of zod (which differs from Valibot, Arktype or Effect Schema). What they have in common, is output and typings of that. Since this supports zod, it should be fine with `@hono/zod-openapi` when using just zod.

Eghizio:
Thanks @muningis 🙇 
It would be amazing to have all of the standard schema compliant validators allow for generating Open API spec & Swagger out of the box in Hono ✨

Issue #890, opened 2024-12-15
[Zod Validator] When the verification fails, can Zod Validator throw an exception instead of directly returning an error message?


quark-ftc:
When the verification fails, can Zod Validator throw an exception instead of directly returning an error message?

askorupskyy:
You can use the middleware for this. https://github.com/honojs/hono/issues/960

ktKongTong:
This can be handled by a wrapper function:

```ts
import {z, ZodSchema} from "zod";
import type {ValidationTargets} from "hono";
import {zValidator as zv} from "@hono/zod-validator";
export const zValidator = <
  T extends ZodSchema,
  Target extends keyof ValidationTargets,
>(
  target: Target,
  schema: T,
) => zv(target, schema, (result, c) => {
  if(!result.success) {
    throw result.error
  }
})
```

or another version that supports a custom hook
```ts
import {z, ZodSchema} from "zod";
import type { ValidationTargets, Env } from "hono";
import { zValidator as zv , type Hook} from "@hono/zod-validator";

export const zValidator = <
  T extends ZodSchema,
  Target extends keyof ValidationTargets,
  E extends Env,
  P extends string
>(
  target: Target,
  schema: T,
  hook: Hook<z.infer<T>, E, P, Target> = (result, _c) => { if(!result.success) throw result.error }
) => zv(target, schema, hook)
```

PR #897, opened 2024-12-20
feat(zod-validator): add `zErrValidator` with default error throwing hook

ktKongTong:
close: https://github.com/honojs/middleware/issues/890

### DIFF
--- a/packages/zod-validator/src/index.ts
+++ b/packages/zod-validator/src/index.ts
@@ -83,3 +83,15 @@ export const zValidator = <
 
     return result.data as z.infer<T>
   })
+
+
+export const zErrValidator = <
+  T extends ZodSchema,
+  Target extends keyof ValidationTargets,
+  E extends Env,
+  P extends string
+>(
+  target: Target,
+  schema: T,
+  hook: Hook<z.infer<T>, E, P, Target> = (result, _c) => { if(!result.success) throw result.error }
+) => zValidator(target, schema, hook)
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @ktKongTong 

`zErrValidator` is interesting, but we should not add it without thinking about it carefully. This is just a wrapper, so I think just adding the description to the README will be okay.

ktKongTong:
Hi @yusukebe 

> we should not add it without thinking about it carefully.

Totally agree.

I believe this is a common use case. More specifically, it addresses the default behavior when a validation error occurs. 

The current default behavior is excellent and is definitely the best choice for most lightweight use cases. 

However, as the API grows in size (e.g., with 50 or more endpoints), the number of biz-errors also increases significantly. In such scenarios, throwing a validation error and handling it in a centralized location, rather than returning a default response, helps maintain a consistent error response format. This allows the client to parse errors in a uniform manner.

Therefore, I created this pull request.

Alternatively, we also can change  the third parameters of `zValidator` , add a string: `throw-err`
to control the default behavior in a more convenience way.

```ts
export const zValidator = <
// ...
>(
  target: Target,
  schema: T,
  resultHandleStrategy?: Hook<z.infer<T>, E, P, Target> | 'throw-err'
): MiddlewareHandler<E, P, V> =>
  // @ts-expect-error not typed well
  validator(target, async (value, c) => {
    // ...
    const result = await schema.safeParseAsync(validatorValue)
    if(typeof resultHandleStrategy === 'string' && resultHandleStrategy === 'throw-err') {
      if(!result.success) {
        throw result.error
      }
    } else if (typeof resultHandleStrategy === 'function' && resultHandleStrategy) {
      const hook = resultHandleStrategy
      const hookResult = await hook({ data: validatorValue, ...result, target }, c)
      // hook result handle ...
    }
      // ...
  })
```


> This is just a wrapper, so I think just adding the description to the README will be okay.

As your suggestion, another  pr(#900) was created.

PR #898, opened 2024-12-20
feat(cloudflare-access): Handle Access organization does not exist and Access not available cases

G4brym:
In the current version of `@hono/cloudflare-access`, both of the actions bellow will result into an "internal error" with a stack trace from the middleware:
- defining an access team name that does not exist
- fetching the access certificates when access is "generally" unavailable

This pr adds a good error message that correctly describes what the middleware received from Cloudflare Access.

This throws normal JS Error's because this is a fatal exception that the user should handle, but if the Hono maintainers think otherwise, i'm happy to change this into `throw new HTTPException(500, { message: '...' })`

### DIFF
--- a/.changeset/heavy-bugs-perform.md
+++ b/.changeset/heavy-bugs-perform.md
@@ -1,0 +1,5 @@
+---
+'@hono/cloudflare-access': minor
+---
+
+Handle Access organization does not exist and Access not available cases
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @G4brym 

Thank you for the PR. In this case, it should throw a `new HTTPException` with the proper status number(is this `500`?) to indicate the status and the message to the application.

G4brym:
Hey @yusukebe, i've just updated the pr to throw `HTTPException`, not sure what the proper HTTP status is, but i think it should be 500, because the app cannot recover from this until either the team name is updated or the cloudflare access recovers from a possible outage

yusukebe (approved):
LGTM!

yusukebe:
@G4brym 

Thank you for changing it. I also considered which is the best status code. It's not 100% sure, but I think `500` is okay. Looks good. Merging and releasing a new version.

PR #901, opened 2024-12-22
feat(otel): Add OpenTelemetry middleware

dahlia:
This middleware addresses <https://github.com/honojs/hono/issues/1864>.  For usage, see also the README file.

### DIFF
--- a/packages/otel/package.json
+++ b/packages/otel/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@hono/otel",
+  "version": "0.0.0",
+  "description": "OpenTelemetry middleware for Hono",
+  "type": "module",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "vitest --run",
+    "build": "tsup ./src/index.ts --format esm,cjs --dts",
+    "publint": "publint",
+    "release": "yarn build && yarn test && yarn publint && yarn publish"
+  },
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/honojs/middleware.git"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/semantic-conventions": "^1.28.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/sdk-trace-base": "^1.30.0",
+    "@opentelemetry/sdk-trace-node": "^1.30.0",
+    "hono": "^4.4.12",
+    "tsup": "^8.1.0",
+    "vitest": "^1.6.0"
+  }
+}
--- a/packages/otel/src/index.ts
+++ b/packages/otel/src/index.ts
@@ -1,0 +1,75 @@
+import { SpanKind, SpanStatusCode, type TracerProvider, trace } from '@opentelemetry/api'
+import {
+  ATTR_HTTP_REQUEST_HEADER,
+  ATTR_HTTP_REQUEST_METHOD,
+  ATTR_HTTP_RESPONSE_HEADER,
+  ATTR_HTTP_RESPONSE_STATUS_CODE,
+  ATTR_URL_FULL,
+  ATTR_HTTP_ROUTE,
+} from '@opentelemetry/semantic-conventions'
+import type { Env, Input } from 'hono'
+import { createMiddleware } from 'hono/factory'
+import metadata from '../package.json' with { type: 'json'}
+
+const PACKAGE_NAME = metadata.name
+const PACKAGE_VERSION = metadata.version
+
+export type OtelOptions = {
+  augmentSpan?: false;
+  tracerProvider?: TracerProvider
+} | {
+  augmentSpan: true;
+}
+
+export const otel = <E extends Env = any, P extends string = any, I extends Input = {}>(
+  options: OtelOptions = {}
+) => {
+  if (options.augmentSpan) {
+    return createMiddleware<E, P, I>(async (c, next) => {
+      const result = await next()
+      const span = trace.getActiveSpan()
+      if (span != null) {
+        const route = c.req.matchedRoutes[c.req.matchedRoutes.length - 1]
+        span.setAttribute(ATTR_HTTP_ROUTE, route.path)
+        span.updateName(`${c.req.method} ${route.path}`)
+      }
+      return result
+    })
+  }
+  const tracerProvider = options.tracerProvider ?? trace.getTracerProvider()
+  const tracer = tracerProvider.getTracer(PACKAGE_NAME, PACKAGE_VERSION)
+  return createMiddleware<E, P, I>(async (c, next) => {
+    const route = c.req.matchedRoutes[c.req.matchedRoutes.length - 1]
+    await tracer.startActiveSpan(
+      `${c.req.method} ${route.path}`,
+      {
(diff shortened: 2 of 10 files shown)


packages/otel/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  ATTR_HTTP_RESPONSE_HEADER,
+  ATTR_HTTP_RESPONSE_STATUS_CODE,
+  ATTR_URL_FULL,
+  ATTR_HTTP_ROUTE,
+} from '@opentelemetry/semantic-conventions'
+import { createMiddleware } from 'hono/factory'
+import type { Env, Input } from 'hono'
+
+const PACKAGE_NAME = '@hono/otel'
+const PACKAGE_VERSION = '0.1.0'
```
dahlia, line 14: I wanted to eliminate the duplication with the metadata in package.json, but couldn't think of a good way to do it.

GewoonJaap:
Does this also work with cloudflare workers? After install the node-sdk of opentelemetry I am getting startup errors. Looks like that package isn't supported with cloudflare workers

dahlia:
> Does this also work with cloudflare workers? After install the node-sdk of opentelemetry I am getting startup errors. Looks like that package isn't supported with cloudflare workers

I have no idea either, but [@microlabs/otel-cf-workers](https://github.com/evanderkoogh/otel-cf-workers) instead of @opentelemetry/sdk-node might work with Cloudflare Workers?

GewoonJaap:
Ah okay. Yeah I made an issue there about Hono compatibility. Looks like Hono exposes things a bit different than the default Cloudflare worker endpoint stuff.

packages/otel/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const otel = <E extends Env = any, P extends string = any, I extends Input = {}>(
+  options: OtelOptions = {}
+) => {
+  const tracerProvider = options.tracerProvider ?? trace.getTracerProvider()
+  return createMiddleware<E, P, I>(async (c, next) => {
+    const tracer = tracerProvider.getTracer(PACKAGE_NAME, PACKAGE_VERSION)
+    const route = c.req.matchedRoutes[c.req.matchedRoutes.length - 1]
+    await tracer.startActiveSpan(
+      `${c.req.method} ${route.path}`,
```
yusukebe, line 28: Is `c.req.path` not good instead of `route.path`?

packages/otel/src/index.ts
yusukebe: It must get the version number from `package.json` since [Changesets](https://github.com/changesets/changesets) will change the number when the PR is merged automatically.

yusukebe:
Hi @dahlia !

Sorry for the late reply. This is good. I've left the comments and one question. What platforms can this middleware run on? It does not work on Cloudflare Workers, but are there runtimes it does not work on? I think you have to write notes about it on README.

packages/otel/src/index.ts
dahlia: Oh, yeah, it's intended! According to OpenTelemetry's [Semantic Conventions for HTTP Spans](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name):

> HTTP span names SHOULD be `{method} {target}` if there is a (low-cardinality) `target` available. If there is no (low-cardinality) `{target}` available, HTTP span names SHOULD be `{method}`.

The [following guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.39.0/specification/trace/api.md#span) will help you understand what we mean by “low-cardinality”:

> The *span name* concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name SHOULD be the most general string that identifies a
(statistically) interesting *class of Spans*, rather than individual Span instances while still being human-readable. That is, "get_user" is a reasonable name, while "get_user/314159", where "314159" is a user ID, is not a good name due to its high cardinality. Generality SHOULD be prioritized over human-readability.
>
> For example, here are potential span names for an endpoint that gets a
hypothetical account information:
>
> | Span Name         | Guidance     |
> | ----------------- | ------------ |
> | `get`             | Too general  |
> | `get_account/42`  | Too specific |
> | `get_account`     | Good, and account_id=42 would make a nice Span attribute |
> | `get_account/{accountId}` | Also good (using the "HTTP route") |

packages/otel/src/index.ts
dahlia: Sorry, as I'm not familiar with packaging JavaScript packages, I have no idea how can I extract the version number from *package.json* file. Actually, I tried to import it like:

```ts
import metadata from '../package.json' with { type: 'json' }

const PACKAGE_NAME = metadata.name
const PACKAGE_VERSION = metadata.version
```

… but it didn't work due to the following errors from `tsc`:

```
src/index.ts(12,22): error TS2732: Cannot find module '../package.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.
src/index.ts(12,40): error TS2823: Import attributes are only supported when the '--module' option is set to 'esnext' or 'nodenext'.
```

Could you give me some idea for this?

yusukebe: How about trying to use the following `tsconfig.json`?

```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "rootDir": "./src",
    "outDir": "./dist",
    "module": "ESNext",
    "resolveJsonModule": true
  },
  "include": [
    "src/**/*.ts"
  ],
}
```

e2digital-eurostar:
Any update on this? Keen to use it. 🙂

packages/otel/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface OtelOptions {
+  tracerProvider?: TracerProvider
+}
+
+export const otel = <E extends Env = any, P extends string = any, I extends Input = {}>(
+  options: OtelOptions = {}
+) => {
+  const tracerProvider = options.tracerProvider ?? trace.getTracerProvider()
+  return createMiddleware<E, P, I>(async (c, next) => {
+    const tracer = tracerProvider.getTracer(PACKAGE_NAME, PACKAGE_VERSION)
```
lucacasonato, line 25: You should do this outside of `createMiddleware()` (right after `tracerProvider`).

lucacasonato:
I think there should be an option to either create a new span (what is implemented in this PR now), or augment an existing span by adding `http.route` and updating the span name with the HTTP route. This is useful if you already have an HTTP span coming from auto-instrumentation, like the one in Deno: https://docs.deno.com/runtime/fundamentals/open_telemetry/#deno.serve

If a user already has an HTTP span and just needs the route augmentation, it is enough to do:

```ts
app.use(async (c, next) => {
  const result = await next();
  const span = trace.getActiveSpan();
  if (span) {
    span.setAttribute("http.route", c.req.routePath);
    span.updateName(`${c.req.method} ${c.req.routePath}`);
  }
  return result;
});
```

packages/otel/src/index.ts
dahlia: Fixed!

packages/otel/src/index.ts
dahlia: Addressed!

dahlia:
@yusukebe @lucacasonato Addressed things you pointed out. Could you take a look again?

packages/otel/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (options.augmentSpan) {
+    return createMiddleware<E, P, I>(async (c, next) => {
+      const result = await next()
+      const span = trace.getActiveSpan()
+      if (span != null) {
+        const route = c.req.matchedRoutes[c.req.matchedRoutes.length - 1]
+        span.setAttribute(ATTR_HTTP_ROUTE, route.path)
+        span.updateName(`${c.req.method} ${route.path}`)
+      }
+      return result
```
lucacasonato, line 36: I think you also need to handle the `c.error` case here.

packages/otel/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@hono/otel",
+  "version": "0.1.0",
```
yusukebe, line 3: The version should be `0.0.0`. The changeset will update the version automatically.

packages/otel/CHANGELOG.md
yusukebe, line 1: Can you remove this file? The changeset will add it automatically.

dahlia:
@yusukebe Done!

yusukebe (approved):
LGTM!

yusukebe:
@dahlia 

Looks good! Let's land it. Thank you for your hard work!

PR #928, opened 2025-01-09
feat: an middleware for compatibility with express/connect for Hono

EdamAme-x:
resolve: https://github.com/honojs/hono/issues/3293

Still some concerns and bugs.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/pink-trainers-destroy.md
+++ b/.changeset/pink-trainers-destroy.md
@@ -1,0 +1,5 @@
+---
+'@hono/connect': major
+---
+
+Connect middleware for Hono
--- a/packages/connect/README.md
+++ b/packages/connect/README.md
@@ -1,0 +1,26 @@
+# Connect middleware for Hono
+
+An middleware for compatibility with express/connect for Hono
+
+## Usage
+
+```ts
+import { connect } from '@hono/connect'
+import { Hono } from 'hono'
+import helmet from 'helmet'
+
+const app = new Hono()
+
+app.use('*', connect(helmet()))
+app.get('/', (c) => c.text('foo'))
+
+export default app
+```
+
+## Author
+
+EdamAme-x <https://github.com/EdamAme-x>
+
+## License
+
+MIT
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@hono/connect",
+  "version": "1.0.0",
+  "description": "An middleware for compatibility with express/connect for Hono",
+  "type": "module",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "vitest --run",
+    "build": "tsup ./src/index.ts --format esm,cjs --dts",
+    "publint": "publint",
+    "release": "yarn build && yarn test && yarn publint && yarn publish"
+  },
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/honojs/middleware.git"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "devDependencies": {
+    "@types/connect": "^3",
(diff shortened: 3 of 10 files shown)


EdamAme-x:
hi @yusukebe 

There are two problems with this implementation.

The first is regarding the header update process.

This test fails.
https://github.com/honojs/middleware/pull/928/files#diff-05c9dabcbf8c76abe4ba387843468024649359e039e0f521ed755f44de39ca44R38-R43

Here is the cause.
https://github.com/honojs/middleware/pull/928/files#diff-e82bf8d64b2d70f2d39b9aa64ada59b6d93549da4aa4f72f7c9f15e3d7426fb0R43

The header is not removed successfully.

EdamAme-x:
Second, even if you use “next” internally, it does not count as Hono's overall middleware.
It is instantiated once internally and the request is passed on.

However, this problem can be sufficiently avoided by being careful about the order of execution.

(It is obvious that the performance is quite bad.)

EdamAme-x:
ToDo: mark as dev deps
    "helmet": "^8.0.0",

EdamAme-x:
hi @yusukebe 
Do you know any reason why the header cannot be removed successfully here?

https://github.com/honojs/middleware/pull/928/files#diff-e82bf8d64b2d70f2d39b9aa64ada59b6d93549da4aa4f72f7c9f15e3d7426fb0R43

yusukebe:
Hi @EdamAme-x 

I think this approach is okay (actually, there is only this way), though the overheads are heavy as you said. I'll check the problems.

dajulia3:
Just here to say that this would be incredibly useful. If you need any additional help or eyes on test edge cases let me know.

EdamAme-x:
There are still a few bugs that need to be fixed.

mbrevda:
> There are still a few bugs that need to be fixed.

Can you elaborate? Perhaps someone can jump in here and help

mridang:
Hi @EdamAme-x any help needed with the issue? One thing I've found is that there is a chance of of Prematurely Finalized Response being Ignored.

The potentiald bug is in how the middleware handles responses sent directly from a Connect middleware (e.g., `res.send()`, `res.end()`).

IIRC, in Hono, when a middleware sets a response directly on the context `(c.res = ...)`, it must also set `c.finalized = true` which tells the Hono engine that response is complete and to stop processing any further middleware or handlers and send this response immediately.

thhe missing `c.finalized` in `packages/connect/src/index.ts`, the code block is `if (res) { c.res = res; }`, which is missing `c.finalized = true;`. This bug might prevent a middleware from successfully ending the request chain.

I also noticed that you are overwriting the headers. 

```
const connectResponse = transformResponseToServerResponse(response);
        const preparedHeaders = (c.newResponse(null, 204, {})).headers;
        const connectHeaders = connectResponse.headers;

        for (const key of [...preparedHeaders.keys()]) {
          c.header(key, undefined);
        }

        for (const [key, value] of [...connectHeaders.entries()]) {
          c.header(key, value);
        }
```

Wouldn't this cause some middleware headers to be lost?

Caceresenzo:
Your branch is pretty old and still use a version of Hono that does not propery handle header clearing.

#928 (accidentally?) fixed it. ([link to patch](https://github.com/honojs/hono/commit/9cdfdf3c65affaec4d3682d66fbd1cffeede975d#diff-0cd594dddd6a9f2e3e26afebbb92716434437d7001c2416252aa7531e7f2b8d4R514-R518))

The faulty lines were:
```javascript
if (this.#headers) {
	this.#headers.delete(name)
} else if (this.#preparedHeaders) {
	delete this.#preparedHeaders[name.toLocaleLowerCase()]
}
```
By using a `else-if`, the headers were not deleted from `#preparedHeaders`, as they were stored at both variable at the same time. Removing the `else` (keeping just the `if`) fix your test.

You can also upgrade to the latest version of Hono, it also make the test pass.

---

There is also other issues, while trying to make a proof of concept using `express.static` middleware, I found out that the URL was not properly set. As if you are in a middleware, you only get it without the prefix.

Meaning that `transformRequestToIncomingMessage`, do not send it correctly:
```javascript
export function transformRequestToIncomingMessage(
	c: Context, // Yes context is needed here
	options?: RequestOptions,
) {
	const request = c.req.raw;
	const parsedURL = new URL(request.url, "http://localhost");

	// ...

	let middlewarePath = routePath(c);

	if (middlewarePath.at(-1) === '*') {
		middlewarePath = middlewarePath.slice(0, -1);
	}
	if (middlewarePath.at(-1) === '/') {
		middlewarePath = middlewarePath.slice(0, -1);
	}

	const message = createRequest({
		method: request.method.toUpperCase() as RequestMethod,
		url: parsedURL.pathname.substring(middlewarePath.length),
		headers: Object.fromEntries(request.headers.entries()),
	// ...

	return message;
}
```

I was unable to make the middleware works in the end as streaming require to use `hono/streaming` which could be tricky to intercept.

PR #936, opened 2025-01-16
fix(oidc-auth): Fixed type error

hnw:
## Summary

While running the oidc-oauth test, I encountered the following error. Therefore, I have fixed it.

The test worked in other environments, so I suspect that the issue was caused by the esbuild version update.

```
    src/index.ts:152:7 - error TS2322: Type 'JWTPayload' is not assignable to type 'Partial<OidcAuth>'.
      'string' index signatures are incompatible.
        Type 'unknown' is not assignable to type 'JsonValue | undefined'.

    152       auth = await verify(session_jwt, env.OIDC_AUTH_SECRET)
              ~~~~
    src/index.ts:248:69 - error TS2345: Argument of type '{} | null' is not assignable to parameter of type 'string'.
      Type 'null' is not assignable to type 'string'.

    248         const response = await oauth2.revocationRequest(as, client, auth.rtk)
                                                                            ~~~~~~~~
```

## Changes

- Fixed type error

## Tasks

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/eighty-trains-peel.md
+++ b/.changeset/eighty-trains-peel.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': patch
+---
+
+Fix type error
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@hnw 

Thank you. It's not better if the CI can't throw the type error but, it's okay to merge it now.

PR #942, opened 2025-01-21
fix(typebox-validator): Fix @hono/typebox-validator, when imported as esm, transiently importing typebox as cjs

### LINKED ISSUE #873: [typebox-validator]: typebox-validator Argument of 'TObject'  is not assignable to parameter of type 'TSchema'
I can't use the TObject schema in the validator due to a type mismatch. There is a problem in the source code that tbValidator esm compiled imports TSchema from cjs @sinclair/typebox.

Github repo with issue implementation:
https://github.com/FOZERY/typebox-validator-issue

Example with default app
image

cfraz89:
Adds package.json to the esm build folder with `"type": "module"`, so that the runtime recognises the type of the package as esm. Then it will import typebox as esm. Fixes #873

### DIFF
--- a/.changeset/purple-weeks-cross.md
+++ b/.changeset/purple-weeks-cross.md
@@ -1,0 +1,5 @@
+---
+'@hono/typebox-validator': patch
+---
+
+Fix transiently importing typebox as cjs even when typebox-validator imported as esm
--- a/packages/typebox-validator/package.json
+++ b/packages/typebox-validator/package.json
@@ -11,7 +11,7 @@
   "scripts": {
     "test": "jest",
     "build:cjs": "tsc -p tsconfig.cjs.json",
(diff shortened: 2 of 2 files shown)


.changeset/purple-weeks-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/typebox-validator': minor
```
yusukebe, line 2: This is a fix change. So you should replace `minor` with `patch`.

cfraz89: Ah I see sorry, have corrected it.

yusukebe: Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@cfraz89 

Thank you! I'll merge and release a new version.

PR #959, opened 2025-02-05
fix(node-ws): ws wasn't created when upgrade process is async

### LINKED ISSUE #954: Awaiting in the upgradeWebSocket callback prevents the websocket events from being registered correctly

4.6.15

Node 22

- Use the upgradeWebSocket helper on a route
- Await anything inside of it 
(Even this will cause the same result)
```
await new Promise((resolve) => {
          setTimeout(() => {
            resolve(`Delay`);
          }, 0);
        });
```

- The promises inside the callback should resolve and the websocket events should be registered normally

### What do you see instead?

- The connection is completed but events don't get registered (more details below)

### Additional information

Awaiting in the callback seems to mess with the order of operations. 
When not awaiting anything (whether the callback is async or not doesn't matter), the order of operation as seen in the debugger is this:
(the functions referenced here come from index.mjs in the @hono/node-ws package)
1. createEvents (the callback given to the upgradeWebSocket helper) is awaited
2. The callback returns
3. nodeUpgradeWebSocket is awaited
4. wss.handleUpgrade runs (so wss.emit 'connection' runs)
5. Events are created (starting with events.onOpen?.(new Event("open"), ctx))

While awaiting, the order changes and looks like this
1. createEvents (the callback given to the upgradeWebSocket helper) is awaited
2. The promise inside the callback is awaited (so the callback hasn't returned the events yet)
3. wss.handleUpgrade with 'emit connection' gets called right after, even if no events have been returned yet
4. Now the callback returns the events
5. nodeUpgradeWebSocket is awaited 
6. Events do not get created and attached to the webSocket, so the webSocket connection does complete, but with no events attached.

Moving the awaiting inside one of the events (onOpen, for example) or even earlier with a preceding middleware works fine.

nakasyou:
Fixes #954 

- [ ] Add tests
- [x] Run tests


yusukebe:
Hi @nakasyou 

Thank you for the PR. I tried it and can confirm the problem was solved with this fix. But can you add a test?

nakasyou:
Hi @yusukebe

I've just added a test for the problem, so could you check it?

yusukebe (approved):
LGTM!

.changeset/big-pillows-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/node-ws': minor
```
yusukebe, line 2: `minor` should be `patch`.

nakasyou:
@yusukebe I fixed it! Can you let me see this merged?

yusukebe:
@nakasyou Thank you!

PR #973, opened 2025-02-19
feat(node-ws): Reject unexpected WebSocket connections

dkulyk:
This PR addresses an issue in which WebSocket connections are permanently established even when the app does not expect them. As a result, these unwanted connections remain active indefinitely, potentially leading to resource leaks and other unintended behavior.

### Problem:

- The server currently accepts all WebSocket connection attempts without verifying if the app expects them.
- Unwanted WebSocket connections remain open, causing unnecessary resource consumption and potential security risks.

### Changes:

If the server detects a WebSocket connection that is not expected by the application, it immediately rejects the handshake and closes the connection.

This behaviour can be disabled by  ```createNodeWebSocket({ app, strict: false })```

### DIFF
--- a/packages/node-ws/src/index.ts
+++ b/packages/node-ws/src/index.ts
@@ -5,6 +5,7 @@ import { WebSocketServer } from 'ws'
 import type { IncomingMessage } from 'http'
 import type { Server } from 'node:http'
 import type { Http2SecureServer, Http2Server } from 'node:http2'
+import type { Duplex } from 'node:stream'
 import { CloseEvent } from './events'
 
 export interface NodeWebSocket {
@@ -15,6 +16,7 @@ export interface NodeWebSocketInit {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   app: Hono<any, any, any>
   baseUrl?: string | URL
+  strict?: boolean
 }
 
 /**
@@ -24,25 +26,25 @@ export interface NodeWebSocketInit {
  */
 export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
   const wss = new WebSocketServer({ noServer: true })
-  const waiter = new Map<IncomingMessage, (ws: WebSocket) => void>()
+  const waiter = new Map<IncomingMessage, { resolve: (ws: WebSocket) => void, response: Response }>()
 
   wss.on('connection', (ws, request) => {
     const waiterFn = waiter.get(request)
     if (waiterFn) {
-      waiterFn(ws)
+      waiterFn.resolve(ws)
       waiter.delete(request)
     }
   })
 
-  const nodeUpgradeWebSocket = (request: IncomingMessage) => {
+  const nodeUpgradeWebSocket = (request: IncomingMessage, response: Response) => {
     return new Promise<WebSocket>((resolve) => {
-      waiter.set(request, resolve)
+      waiter.set(request, { resolve, response })
     })
   }
 
   return {
     injectWebSocket(server) {
-      server.on('upgrade', async (request, socket, head) => {
+      server.on('upgrade', async (request, socket: Duplex, head) => {
         const url = new URL(request.url ?? '/', init.baseUrl ?? 'http://localhost')
         const headers = new Headers()
         for (const key in request.headers) {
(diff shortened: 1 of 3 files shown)


yusukebe:
@dkulyk Thank you for the PR.

@nakasyou Can you review this?

nakasyou:
Hi @yusukebe
Of course! However give me a day due to entrance exam.

yusukebe:
@nakasyou Good luck with the entrance exam!

packages/node-ws/src/index.ts
```diff
@@ -24,25 +26,25 @@ export interface NodeWebSocketInit {
  */
 export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
   const wss = new WebSocketServer({ noServer: true })
-  const waiter = new Map<IncomingMessage, (ws: WebSocket) => void>()
+  const waiter = new Map<IncomingMessage, { resolve: (ws: WebSocket) => void, response: Response }>()
 
   wss.on('connection', (ws, request) => {
     const waiterFn = waiter.get(request)
```
nakasyou, line 32: I think the name `waiterFn` should be changed because it is no longer function.

packages/node-ws/src/index.ts
```diff
@@ -15,6 +16,7 @@ export interface NodeWebSocketInit {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   app: Hono<any, any, any>
   baseUrl?: string | URL
+  strict?: boolean
```
nakasyou, line 19: I can't think of the way using `strict` to allow connecting WebSocket with unexpected path, IMO this is not needed. What do you think?

dkulyk: I left it for backward compatibility. If devs use it for some reason.

packages/node-ws/src/index.ts
dkulyk: I can rename the map to waiterMap and waiterFn to just waiter. I think it will be more consistent

nakasyou: It sounds good!

packages/node-ws/src/index.ts
nakasyou: I see but I can't think of the way to use (and this is a bug). How about adding `@deprecated`?

dkulyk: The wss is not allowed outside. It really has no sense. I'll remove it

nakasyou (approved):
Looks good!

packages/node-ws/src/index.ts
```diff
@@ -52,11 +53,25 @@ export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
...
-        await init.app.request(
+
+        const response = await init.app.request(
           url,
           { headers: headers },
           { incoming: request, outgoing: undefined }
         )
+
+        const waiter = waiterMap.get(request)
+        if (!waiter || waiter.response !== response) {
```
yusukebe, line 64: @dkulyk  @nakasyou 

The following test case will fail. Is this should be passed? The current implementation will not throw an error.

```ts
it('Should allow WebSocket upgrade even if response is modified after next()', async () => {
  app.get('*', async (c, next) => {
    await next()
    c.res = new Response('foo')
  })

  app.get(
    '/',
    upgradeWebSocket(() => ({}))
  )

  const ws = new WebSocket('ws://localhost:3030/')
  const mainPromise = new Promise<boolean>((resolve) => {
    ws.onopen = () => {
      resolve(true)
    }
    ws.onerror = () => {
      resolve(false)
    }
  })

  expect(await mainPromise).toBe(true)
})
```

nakasyou: I can't imagine when this code is necessary. From where I stand, there is no need to pass that test case.

yusukebe: @nakasyou 

Thank you for the response. It's a rare case, but some middleware uses `c.res = response` such as Trailing Slash Middleware. The following test will fail.

```ts
it('Should allow WebSocket upgrade even if response is modified after next()', async () => {
  app.use(trimTrailingSlash())

  app.get(
    '/',
    upgradeWebSocket(() => ({}))
  )

  const ws = new WebSocket('ws://localhost:3030/')
  const mainPromise = new Promise<boolean>((resolve) => {
    ws.onopen = () => {
      resolve(true)
    }
    ws.onerror = () => {
      resolve(false)
    }
  })

  expect(await mainPromise).toBe(true)
})
```

I think it's not good to compare the Response objects as following line:

```ts
waiter?.response === response
```

yusukebe: @nakasyou 

Sorry again.

I considered it. I also realized there is no case for `c.res = response` when the endpoint is applied to WebSockets. So, it is okay, though the behavior is slightly changed from the current implementation.

yusukebe (approved):
LGTM!

yusukebe:
Hi @dkulyk 

Looks good. Thank you for your contribution!

PatrickRoumanoff:
This looks like it is now rejecting a valid websocket connection if a middleware is creating a new Response object through `c.res = new Response`
In my case, middlewares have been setup to update the response before next for the http use case on a higher level, such as `route.use(myMiddleware)`.
Does this mean that after 1.0.8 I can't mingle my routes between http and websocket if I need to use a middleware that creates a new response before `next()`?

PR #976, opened 2025-02-23
Pass optional state as parameter, so we can link other state info to it

mansimransingh:
`state?: string` is an option in google oauth middleware. I think its generally useful to be able to pass custom state identifier since it allows us to attach arbitrary data to a request.


yusukebe:
@mansimransingh Thank you for the PR!

Hi @monoald, Can you review this?

PR #984, opened 2025-02-27
Add ability to exclude routes from OpenAPI docs

rmichalak:
Similar to what is available here: https://github.com/rhinobase/hono-openapi?tab=readme-ov-file#conditionaly-hiding-routes

### DIFF
--- a/.changeset/hot-moles-fix.md
+++ b/.changeset/hot-moles-fix.md
@@ -1,5 +1,5 @@
 ---
-'@hono/zod-openapi': patch
+'@hono/zod-openapi': minor
 ---
 
 Add ability to exclude specific routes from OpenAPI docs
--- a/.changeset/hot-moles-fix.md
+++ b/.changeset/hot-moles-fix.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Add ability to exclude specific routes from OpenAPI docs
--- a/packages/zod-openapi/test/index.test.ts
+++ b/packages/zod-openapi/test/index.test.ts
@@ -1923,3 +1923,48 @@ describe('Generate YAML', () => {
     expect(() => stringify(doc)).to.not.throw()
   })
 })
+
+describe('Hide Routes', () => {
+  it('Should hide the route', async () => {
+    const app = new OpenAPIHono()
+    app.openapi(
+      createRoute({
+        method: 'get',
+        hide: true,
+        path: '/books',
+        responses: {
+          200: {
+            description: 'Books',
+            content: {
+              'application/json': {
+                schema: z.array(
+                  z.object({
+                    title: z.string(),
+                  })
+                ),
+              },
+            },
+          },
+        },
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @rmichalak 

What use case is this feature? I think it's okay to use just `app.get()`:

```ts
app.get('/books', (c) => c.json([{ title: 'foo' }]))
```

rmichalak:
Hi @yusukebe,

> What use case is this feature? I think it's okay to use just `app.get()`:

Will I still have all the type-safety and validation coming from the openapi specification?

yusukebe:
@rmichalak 

The response can't be typed, but you can use `@hono/zod-validator` to validate and add types for requests.

rmichalak:
@yusukebe 

> The response can't be typed, but you can use `@hono/zod-validator` to validate and add types for requests.

Well, but that would mean I need to write more code and use a package which I currently don't use explicitly (although I know it's used behind the scenes). On the other hand, with my change I can use exactly the same patterns as with any other endpoint.

Plus it's easier to expose the endpoint dynamically in runtime, e.g. for specific environments:
```ts
hide: process.env.ENV === "test"
```

yusukebe:
@rmichalak 

I still can't understand why this feature is needed. Is the following not good?

```ts
const app = new OpenAPIHono()
if (condition) {
  app.openapi(
    createRoute({
      /** */
    }),
    (c) => c.json([{ title: 'foo' }])
  )
}
```

rmichalak:
@yoshikouki but I do want the endpoint to be available for requests, just hidden from the docs. Let's imagine a new endpoint which we want to test internally, but not announce publicly yet. With your approach, when the `condition` is met, the endpoint will be both available and listed in OpenAPI docs. When the `condition` is not met, the endpoint with not even be deployed.

I want to achieve something in the middle - deployed but not listed.

.changeset/hot-moles-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
```
yusukebe, line 2: This is a new feature. This should be `minor` instead of `patch`.

yusukebe:
@rmichalak 

Thank you for the explanation. I fully understand! The code is good, but I've added the comment.

packages/zod-openapi/test/index.test.ts
```diff
@@ -1923,3 +1923,48 @@ describe('Generate YAML', () => {
     expect(() => stringify(doc)).to.not.throw()
   })
 })
+
+describe('Hide Routes', () => {
+  it('Should hide the route', async () => {
```
yusukebe, line 1928: Can you add a test to test `/books` returns a 200 response correctly?

rmichalak: Addressed both comments

yusukebe:
@rmichalak 

Looks good! I'll merge and release a new version. Thank you for your contribution!

PR #985, opened 2025-02-27
RFC: Response cache

muningis:

- [ ] Add tests
- [ ] Run tests


packages/standard-validator/src/index.ts
```diff
@@ -1,6 +1,6 @@
+import type { StandardSchemaV1 } from '@standard-schema/spec'
```
muningis, line 1: These changes are already existing in honojs:main, will have to recreate branch to fix this.

muningis:
@yusukebe Hy, could you check this out please? :) Still very basic, no tests even yet, just to get feeling/idea if there would be interest to continue with this

yusukebe:
At first glance, I don't think this is needed. We already have the [built-in Cache Middleware](https://hono.dev/docs/middleware/builtin/cache). If you want to cache just "Response", it's enough to use it.

muningis:
As per my understanding, that middleware only works on cloudflare an deno, and uses specific cache mechanism. This allows little bit more freedom, for example, by using `redis` which has TTL (unlike Cache API).

Also, correct me if I'm wrong, but that Cache API helps cache'ing it on Cloudflare/DenoDeploy level, basically "edge/cdn" rather than on server, right?

muningis:
Forgot to mention - this allows to customize cache’ing keys. So for example, if you’re running A/B test, and store A/B tests list in cookie - you can use that cookie to allow to cache page rendered under specific A/B conditions.

Quite niche example, but solution is quite lightweight with provided flexibility.

yusukebe:
@muningis 

Thank you for the explanation! Makes sense.

> Also, correct me if I'm wrong, but that Cache API helps cache'ing it on Cloudflare/DenoDeploy level, basically "edge/cdn" rather than on server, right?

Yes. The Cache API is standardized, but you can only use it in particular runtimes like Cloudflare Workers and Deno.

This middleware looks good. Let's go ahead.

piotrcichosz:
redis - or for example Cloudflare KV..

@muningis are you planning on finishing this PR?

muningis:
Had completely forgoten about this. I will get back to this next week.

yusukebe:
@muningis 

If this is ready to review, please ping me.

muningis:
Hello @yusukebe, yes this is ready now.

packages/response-cache/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        return new Response(body, {
+          status,
+          headers
+        })
+      } else {
+        if (logging?.enabled) {logging.onMiss?.(key, c)}
+        await next()
+
+        const body = await c.res.clone().text()
```
yusukebe: I believe there are cases where the response body can't be parsed with `res.text()`.

packages/response-cache/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+interface CacheMiddlewareOptions {
+  store: CacheStore;
+  /**
+   * Function to generate a cache key from the request and context. If not provided, the request path will be used.
+   * @param req - The request object.
+   * @param c - The context object.
+   * @returns A string key for the cache.
+   */
+  keyFn?: (req: HonoRequest, c: Context) => string;
```
yusukebe: Is not okay the following spec?

```ts
keyFn?: (c: Context) => string
```

.changeset/khaki-tires-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/response-cache': major
```
yusukebe: Can you set it `minor` instead of `major`? This is like a taste of matter, but I prefer to release it first as `0.1.0`.

packages/response-cache/src/index.ts
muningis: Good point. Works for HTML/JSON/Text responses, otherwise might need different handling. Will use headers checking to make this better.

.changeset/khaki-tires-roll.md
muningis: Sure thing

packages/response-cache/src/index.ts
muningis: Yeah, considering `Context` already contains `HonoRequest`, we can adjust.

yusukebe:
@muningis 

Can you ping me if you fixed them?

packages/response-cache/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+const responseCache = ({ store, keyFn, logging }: CacheMiddlewareOptions) => {
+  return createMiddleware(async (c, next) => {
+    const key = keyFn ? keyFn(c.req, c) : c.req.path
+    try {
+      const cached = await store.get(key)
+      if (cached) {
+        if (logging?.enabled) {logging.onHit?.(key, c)}
+
+        const snapshot = JSON.parse(cached)
```
piotrcichosz, line 53: I wonder if we need JSON parse here - maybe this logic could be in store as it can depend on storage service?
I.e. cloudflare workers cache api vs kv api - cache api returns Response, KV can return json or stream.

muningis: @yusukebe what do you think? I'm inclined to agree, as this makes it more flexible, and perhaps even easier invalidation too.

muningis:
@yusukebe - adjusted accordingly.

yusukebe:
@muningis 

I believe we can't merge this PR into the main because the branch has inconsistencies. Can you create another PR? Plus, can you set the project to match the other projects? For example, the `package.json` settings differ from those of other projects.

BarryThePenguin:
@muningis If you get stuck, let me know 👋🏻

muningis:
https://github.com/honojs/middleware/pull/1652
Adjusted package.json, deno.json and tsdown instead of tsup. 
Checking why `yarn install` fails on new PR.
let me know if I missed anything else.

PR #998, opened 2025-03-02
chore: Migrate workspaces that use `jest` to `vitest`

### LINKED ISSUE #994: Investigate migrating workspaces that use `jest` to `vitest`
Is it worth replacing `@jest/globals` and [`vitest/globals`](https://vitest.dev/config/#globals) by importing the `vitest` module in tests? This aligns with Vitest defaults and requires less configuration.

I noticed that some workspaces put tests in a `test` folder, while others add `.tests.ts` files to the `src` directory. Is there a preference? My preference would be whichever requires the least configuration

It looks like most workspaces are using either v1 or v2, so I'll stick with v2 for now.

To use [vitest.workspace](https://vitest.dev/guide/workspace.html), all versions of `vitest` will need to match, so that can be done separately.

BarryThePenguin:
Fixes #994


packages/graphql-server/vitest.config.ts
```diff
@@ -0,0 +1,16 @@
...
+  test: {
+    exclude: [...defaultExclude, 'bun_test'],
+    globals: true,
+    restoreMocks: true,
+    unstubEnvs: true,
+    onConsoleLog(_, type) {
+      // Do not show `console.error` messages
+      if (type === 'stderr') {
+        return false
+      }
```
BarryThePenguin, line 13: I'm unsure how useful this is.. it's just replicating the `spyOn(console, 'error')` that was here previously

https://github.com/honojs/middleware/blob/0f8d2f3d055e7873c22960db6a4a10e20bd75d91/packages/graphql-server/test/index.test.ts#L13-L14

packages/clerk-auth/src/index.test.ts
```diff
@@ -1,36 +1,31 @@
...
 
 const EnvVariables = {
   CLERK_SECRET_KEY: 'TEST_API_KEY',
   CLERK_PUBLISHABLE_KEY: 'TEST_API_KEY',
 }
 
-const authenticateRequestMock = jest.fn()
+const authenticateRequestMock = vi.fn()
+
+vi.mock(import('@clerk/backend'), async (importOriginal) => {
```
BarryThePenguin, line 11: `vi.mock` also supports types, so I changed the implementation of the mocks slightly to include the original properties, along with the mocked properties

yusukebe:
Hi @BarryThePenguin !

Thank you for working on hard work!

You/we are using version `^2.1.8` of `vitest` but, the last version is `3.*`, so is it good to bump it to the `3.*`? I think it should not be right now, and you can do it after merging this.

BarryThePenguin:
Yes, that's right. Some packages were using v1 and others v2, so I stayed with v2 in this PR. I can take a look at upgrading to v3 next 👍🏻

packages/oauth-providers/vitest.config.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from 'vitest/config'
 
 export default defineConfig({
   test: {
+    environment: 'miniflare',
```
BarryThePenguin, line 5: Using `vitest-environment-miniflare` will make the `crypto` global accessible in tests, similar to what `jest-environment-miniflare` was doing previously.

The tests for `@hono/oauth-providers` use `msw` which won't work with `@cloudflare/vitest-pool-workers`. I did try that first, but it will require rewriting the `msw` handlers to use [`import { fetchMock } from "cloudflare:test"`](https://developers.cloudflare.com/workers/testing/vitest-integration/test-apis/#cloudflaretest-module-definition)

yarn.lock
```diff
@@ -2722,6 +2722,7 @@ __metadata:
     "@asteasolutions/zod-to-openapi": "npm:^7.1.0"
     "@cloudflare/workers-types": "npm:^4.20240117.0"
     "@hono/zod-validator": "npm:^0.4.1"
+    "@types/node": "npm:^20.10.4"
```
BarryThePenguin, line 2725: I'm unsure if it's intentional or not, but `@hono/zod-openapi` has it's own `yarn.lock` file which leads to weird build issues where a dependency exists in the root lock file, but not in `@hono/zod-openapi` 🤔 

Unsure if it's simple to remove it or not..

yusukebe: This is intended to specify `@hono/zod-validator` on `npm` instead of in on `workspace`.

The current dependencies specify like the following:

https://github.com/honojs/middleware/blob/0f8d2f3d055e7873c22960db6a4a10e20bd75d91/packages/zod-openapi/package.json#L55

But if you do `yarn add @hono/zod-validator` it will be `workspace:^`.

CleanShot 2025-03-04 at 08 14 33@2x

This will be okay in another general case (?), but it is not suitable for our project. Any good idea?

BarryThePenguin: I think `yarn add "@hono/zod-validator@npm:^0.4.1"` should be enough, though it doesn't look like the types from `0.4.3` are compatible with `@hono/zod-openapi`?

So either `@hono/zod-validator@npm:0.4.1"` or `@hono/zod-validator@npm:0.4.2"` might be better

yusukebe: Okay! Let's go with `yarn add "@hono/zod-validator@npm:0.4.2"`. It's the latest version which is compatible with `@hono/zod-openapi`.

BarryThePenguin: Done 👍🏻

yusukebe (approved):
LGTM!

PR #1004, opened 2025-03-05
test(oauth-providers): use node 20.x in CI

### LINKED ISSUE #1003: Remove `vitest-environment-miniflare`
To update to Vitest v3 we need to remove `vitest-environment-miniflare`. The main reason this exists is to provide the `crypto` global in the `@hono/oauth-providers` tests.

> The tests for `@hono/oauth-providers` use `msw` which won't work with `@cloudflare/vitest-pool-workers`. I did try that first, but it will require rewriting the `msw` handlers to use [`import { fetchMock } from "cloudflare:test"`](https://developers.cloudflare.com/workers/testing/vitest-integration/test-apis/#cloudflaretest-module-definition)
> 
> _Originally posted by @BarryThePenguin in https://github.com/honojs/middleware/pull/998#discussion_r1977267400_

Alternatively, I think we can upgrade CI to use node v20 which does have the global `crypto` variable. That way we won't need to remove `msw`

BarryThePenguin:
One step closer to Vitest v3 🎉 

Fixes #1003

### DIFF
--- a/.github/workflows/ci-oauth-providers.yml
+++ b/.github/workflows/ci-oauth-providers.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.x
+          node-version: 20.x
       - run: yarn install --frozen-lockfile
       - run: yarn build
       - run: yarn test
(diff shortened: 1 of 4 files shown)


yusukebe:
@BarryThePenguin Thank you! Merging.

PR #1006, opened 2025-03-05
fix: ArkType errors with Hono 4

FLchs:


packages/arktype-validator/src/index.test.ts
```diff
@@ -88,16 +88,21 @@ describe('Basic', () => {
         age: '20',
       }),
       method: 'POST',
+      headers: {
```
FLchs, line 91: Not specifying the content type made the test successful even with valid body...

PR #1010, opened 2025-03-09
feat(oidc-auth): allow setting audience for oidc-auth

wayofthepie:
This allows setting the audience with the `oidc-auth` middleware. Without it most providers just return an opaque token for the access token and it cannot be used for authorization. Example for Auth0 https://community.auth0.com/t/getting-access-token-without-audience/57871/3.


yusukebe:
@wayofthepie Thank you for the PR!

@hnw Can you review this?

wayofthepie:
Woops accidentally closed via a comment!

hnw:
Thank you for your contribution. I appreciate you taking the time to submit this pull request and I understand the intent behind your proposal.

Regarding the `audience` parameter you've added, it's my understanding that Auth0 is the only OIDC provider that utilizes this parameter for authorization server requests. RFC 8707 defines the `resource` parameter for a similar purpose, and this parameter seems to be implemented by multiple OIDC providers.

To prevent potential confusion for users working with OIDC providers other than Auth0, could you please add a note in the parameter's documentation explicitly stating that it is primarily intended for use with Auth0?

Thank you for your understanding and for making this adjustment.

wayofthepie:
Hi @hnw,, thanks for the response. I have updated with a note on the audience use with Auth0 vs other providers in the README. Let me know if that makes sense to you.

packages/oidc-auth/src/index.ts
```diff
@@ -308,6 +309,9 @@ const generateAuthorizationRequestUrl = async (
     authorizationRequestUrl.searchParams.set('access_type', 'offline')
     authorizationRequestUrl.searchParams.set('prompt', 'consent')
   }
+  if (env.OIDC_AUDIENCE) {
```
rajsite, line 312: fyi @wayofthepie, with the latest changes merged in main you will likely need a tweak to include this new OIDC variable in the config object: https://github.com/honojs/middleware/blob/e678c36193bd10f63448b08b6989873095722f61/packages/oidc-auth/src/index.ts#L84-L96

wayofthepie: Thanks! I can have a look at pulling in the latest changes soon, and update that.

wayofthepie:
I think that covers everything from the suggestions from @rajsite now.

rajsite (approved):

packages/oidc-auth/src/index.test.ts
```diff
@@ -374,6 +375,27 @@ describe('oidcAuthMiddleware()', () => {
...
+  })
+  test('Should not set audience if not defined', async () => {
+    const req = new Request('http://localhost/', {
+      method: 'GET',
+      headers: { cookie: `oidc-auth=${MOCK_JWT_EXPIRED_SESSION}` },
+    })
+    const res = await app.request(req, {}, {})
+    expect(res).not.toBeNull()
+    expect(res.status).toBe(302)
+    expect(res.headers.get('location')).not.toMatch(/audience=audience/)
```
rajsite, line 397: Maybe a search for just the url param would be a more general assert?

```suggestion
    expect(res.headers.get('location')).not.toMatch(/audience=/)
```

wayofthepie: Sure, that makes sense. Updated also.

yusukebe (approved):
LGTM!

PR #1021, opened 2025-03-17
ci(coverage): upload initial coverage to codecov

### LINKED ISSUE #1020: setup `test.coverage` for Vitest

BarryThePenguin:
Adding code coverage based on honojs/hono#2862

Using [Carryforward Flags](https://docs.codecov.com/docs/carryforward-flags) to collect coverage from individual paths in the monorepo

Fixes #1020

### DIFF
--- a/.github/workflows/ci-coverage.yml
+++ b/.github/workflows/ci-coverage.yml
@@ -1,0 +1,21 @@
+name: initial-coverage
+on:
+  pull_request:
+    branches: ['*']
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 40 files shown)


.github/workflows/ci-coverage.yml
```diff
@@ -0,0 +1,21 @@
+name: initial-coverage
```
BarryThePenguin, line 1: We need to upload some coverage as a baseline so there's something to compare carry forward reports with.

We can remove this after a report has been generated on `main`

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

I've never set up the coverage of so many projects before, but I think this looks good! Merging.

PR #1026, opened 2025-03-19

BarryThePenguin:
Trying out Bun, based on #561


yusukebe:
@BarryThePenguin 

Did you really want to close this? Or accidentally?

BarryThePenguin:
Yes I mean to close it. The tests appeared to work locally so I wanted to try it out on CI, but a handful of builds failed

It looks like Bun isn't resolving peer dependencies like yarn does. I'll spend some time cleaning up dependencies first before taking about look at Bun again 👍🏻

PR #1027, opened 2025-03-22
Move "default" entrypoint down in "typia-validator"

koteelok:
The "typia-validator" middleware just doesn't work in ESM builds. Has nobody used it before?

The entry point "default" should be the last option in the `package.json`:
https://nodejs.org/api/packages.html#conditional-exports


yusukebe (approved):
LGTM!

yusukebe:
@koteelok 

Thanks! I'll merge this.

PR #1030, opened 2025-03-23
build(ajv-validator): lint published package

BarryThePenguin:
I'm looking at aligning the `build`/`test`/`release` scripts in each workspace. This is the template that I'll aim to follow for each workspace. If we get this right, the rest will be pretty straightforward

### DIFF
--- a/.github/workflows/ci-ajv-validator.yml
+++ b/.github/workflows/ci-ajv-validator.yml
@@ -19,6 +19,7 @@ jobs:
           node-version: 20.x
       - run: yarn workspaces focus hono-middleware @hono/ajv-validator
       - run: yarn workspace @hono/ajv-validator build
+      - run: yarn workspace @hono/ajv-validator publint
       - run: yarn test --coverage --project @hono/ajv-validator
       - uses: codecov/codecov-action@v5
         with:
--- a/packages/ajv-validator/package.json
+++ b/packages/ajv-validator/package.json
@@ -9,10 +9,10 @@
     "dist"
   ],
   "scripts": {
-    "test": "vitest --run",
-    "build": "tsup ./src/index.ts --format esm,cjs --dts",
-    "publint": "publint",
-    "release": "yarn build && yarn test && yarn publint && yarn publish"
+    "build": "tsup ./src/index.ts",
+    "prepack": "yarn build",
+    "publint": "attw --pack && publint",
+    "test": "vitest"
   },
   "exports": {
     ".": {
@@ -33,7 +33,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/honojs/middleware.git"
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/ajv-validator"
   },
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
@@ -41,10 +42,11 @@
     "hono": ">=3.9.0"
   },
   "devDependencies": {
+    "@arethetypeswrong/cli": "^0.17.4",
     "ajv": ">=8.12.0",
     "hono": "^4.4.12",
-    "tsup": "^8.1.0",
-    "typescript": "^5.8.2",
+    "publint": "^0.3.9",
+    "tsup": "^8.4.0",
     "vitest": "^3.0.8"
   }
 }
--- a/packages/ajv-validator/tsconfig.json
+++ b/packages/ajv-validator/tsconfig.json
@@ -1,10 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "./src",
-    "outDir": "./dist",
-  },
-  "include": [
-    "src/**/*.ts"
-  ],
+    "outDir": "./dist"
+  }
 }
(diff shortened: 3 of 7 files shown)


packages/ajv-validator/package.json
```diff
@@ -9,10 +9,10 @@
...
   ],
   "scripts": {
-    "test": "vitest --run",
-    "build": "tsup ./src/index.ts --format esm,cjs --dts",
-    "publint": "publint",
-    "release": "yarn build && yarn test && yarn publint && yarn publish"
+    "build": "tsup ./src/index.ts",
+    "prepack": "yarn build",
+    "publint": "attw --pack && publint",
+    "test": "vitest"
```
BarryThePenguin, line 15: I'm looking to use these scripts as a template for other workspaces

- `build` - stays mostly the same. Referencing a shared config in the root, each workspace can [override config using cli arguments](https://tsup.egoist.dev/#using-custom-configuration)
- `prepack` - is run [anytime `yarn pack` is called](https://yarnpkg.com/advanced/lifecycle-scripts#prepack-and-postpack), and is triggered by both `attw` and `publint`
- `publint` - is just a convenient way to call both `attw` and `publint`. An alternative would be to have CI call the scripts individually
- `test` - isn't used by CI, it may just be handy to keep around

.github/workflows/ci-ajv-validator.yml
```diff
@@ -19,6 +19,7 @@ jobs:
           node-version: 20.x
       - run: yarn workspaces focus hono-middleware @hono/ajv-validator
       - run: yarn workspace @hono/ajv-validator build
+      - run: yarn workspace @hono/ajv-validator publint
```
BarryThePenguin, line 22: CI runs the new `publint` script

packages/ajv-validator/package.json
```diff
@@ -9,10 +9,10 @@
     "dist"
   ],
   "scripts": {
-    "test": "vitest --run",
-    "build": "tsup ./src/index.ts --format esm,cjs --dts",
-    "publint": "publint",
-    "release": "yarn build && yarn test && yarn publint && yarn publish"
```
BarryThePenguin, line 15: - [`vitest --run`](https://vitest.dev/guide/cli.html#run) disables watch mode, which is disabled by default in CI. The option is redundant, so I've removed it
- `tsup --format esm,cjs --dts` are both now referenced in the shared `tsup.config.ts` file, along with a few other options
- [`release`] - it doesn't look like this script is being used anywhere, Changesets doesn't look like it's configured to use it, so I thought is was best to remove it.

It looks like https://github.com/honojs/middleware/blob/main/.github/workflows/release.yml just runs `yarn build` and Changesets does the release. Using `prepack` should remove the need to build all workspaces before doing a publish, so only the packages being published are built

packages/ajv-validator/package.json
```diff
@@ -33,18 +33,20 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/honojs/middleware.git"
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/ajv-validator"
```
BarryThePenguin, line 37: `publint` pointed out these two improvements

See https://docs.npmjs.com/cli/v11/configuring-npm/package-json#repository

packages/ajv-validator/tsconfig.json
```diff
@@ -1,10 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "./src",
-    "outDir": "./dist",
-  },
-  "include": [
-    "src/**/*.ts"
-  ],
-}
```
BarryThePenguin, line 10: These options are both redundant, only `outDir` is really necessary

- [`rootDir`](https://www.typescriptlang.org/tsconfig/#rootDir)
- [`include`](https://www.typescriptlang.org/tsconfig/#include)

packages/ajv-validator/tsup.config.ts
```diff
@@ -0,0 +1 @@
+export { default } from '../../tsup.config'
```
BarryThePenguin, line 1: The alternative to this file is specifying `--config ../../tsup.config.ts` in the `build` script

packages/ajv-validator/package.json
yusukebe: What do you think of adding a Type check using the `tsc`? It now does not have some command doing `tsc` in Ajv validator middleware. But I think it's better to do a Type check, though I'm wondering if we should put it in the `test` phase or other places. If adding the `tsc`, that will fail in this Ajv validator, so we have to fix it. At the same, I think there are some projects we have to fix about Types of matter.

BarryThePenguin: I agree 👍🏻 I have plans look at using [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) to do type checking. That should allow us to use seperate tsconfig files to check individual workspaces, including any workspace dependencies. This would avoid the need to type check the entire monorepo for every build/pr/etc

I'll add it to my list 😅

yusukebe: @BarryThePenguin 

Awesome! Please do so.

yusukebe (approved):
LGTM!

PR #1033, opened 2025-03-24
build(arktype-validator): lint published package

BarryThePenguin:


packages/ajv-validator/tsup.config.ts
BarryThePenguin, line 1: Based on the log output, it looks like tsup automatically picks up on the root `tsup.config.ts`

> `[@hono/ajv-validator]: CLI Using tsup config: /Users/jonno/src/honojs/middleware/tsup.config.ts`

So there's no need to re-export it in the workspace

BarryThePenguin, line 1: Based on the log output, it looks like tsup automatically picks up on the root `tsup.config.ts`

> `[@hono/ajv-validator]: CLI Using tsup config: /home/runner/work/middleware/middleware/packages/ajv-validator/tsup.config.ts`

So there's no need to re-export it in the workspace

yusukebe (approved):
LGTM!

PR #1046, opened 2025-03-26
build(esbuild-transpiler): lint published package

BarryThePenguin:


packages/esbuild-transpiler/package.json
```diff
@@ -2,32 +2,48 @@
...
   "types": "dist/index.d.ts",
   "files": [
     "dist"
   ],
   "scripts": {
-    "test": "vitest run",
-    "build": "tsup ./src/*.ts ./src/transpilers/*.ts --format esm,cjs --dts --no-splitting --external esbuild-wasm,esbuild",
-    "publint": "publint",
-    "release": "yarn build && yarn test && yarn publint && yarn publish"
+    "build": "tsup ./src/*.ts ./src/transpilers/*.ts --no-splitting --external esbuild-wasm,esbuild",
```
BarryThePenguin, line 12: It looks like `--no-splitting` is actually doing something in this case, so I've left it in

packages/esbuild-transpiler/package.json
```diff
@@ -47,20 +63,23 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/honojs/middleware.git"
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "esbuild-transpiler"
```
yusukebe, line 67: This should be `packages/esbuild-transpiler`.

yusukebe (changes requested):

packages/esbuild-transpiler/package.json
BarryThePenguin: Nice pickup, thanks!

```suggestion
    "directory": "packages/esbuild-transpiler"
```

I did seperate PRs hoping it would be easier to review and pickup small things like this. Sorry there are so many! 😅

yusukebe: No problem!

yusukebe (approved):
LGTM!

PR #1053, opened 2025-03-26
build(oauth-providers): lint published package

BarryThePenguin:


packages/oauth-providers/package.json
```diff
@@ -2,96 +2,38 @@
...
       },
       "require": {
-        "types": "./dist/providers/discord/index.d.ts",
-        "default": "./dist/providers/discord/index.js"
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
       }
     },
-    "./twitch": {
+    "./*": {
```
BarryThePenguin, line 29: I also reduced all the seperate providers `exports` into a single [subpath pattern](https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-patterns), hopefully that makes it a little easier to add new providers in the future

yusukebe: I've confirmed that this works well. Thanks!

yusukebe (approved):
LGTM!

PR #1054, opened 2025-03-26
build(oidc-auth): lint published package

BarryThePenguin:


packages/oidc-auth/package.json
```diff
@@ -9,10 +9,10 @@
...
   ],
   "scripts": {
-    "test": "vitest",
-    "build": "tsup ./src/index.ts --format esm --dts",
-    "prerelease": "yarn build && yarn test",
-    "release": "yarn prerelease && yarn npm publish"
+    "build": "tsup ./src/index.ts --format esm",
+    "prepack": "yarn build",
+    "publint": "attw --pack --profile esm-only && publint",
+    "test": "vitest"
```
BarryThePenguin, line 15: This is the first package that I've noticed is esm only 👀

yusukebe: I think exporting only esm is not intended. So can you make it export cjs as well?

BarryThePenguin: Sure thing 👍🏻

PR #1067, opened 2025-03-26
build(typia-validator): lint published package

BarryThePenguin:


.github/workflows/ci-typia-validator.yml
```diff
@@ -19,6 +19,7 @@ jobs:
           node-version: 18.x
```
yusukebe, line 19: Ah, I've commented on other places, but we have the CI with Node.js version `18.x`. I think it's better to use a later version, `20.x`. So, after merging all `lint published package` PRs, could you create a PR to update the Node.js version?

yusukebe (approved):
LGTM!

PR #1068, opened 2025-03-26
build(valibot-validator): lint published package

BarryThePenguin:


.github/workflows/ci-valibot-validator.yml
```diff
@@ -19,6 +19,7 @@ jobs:
           node-version: 18.x
```
yusukebe, line 19: How about updating it to `20.x`?

yusukebe: Let's not change here in this PR for the reason of [the comment](https://github.com/honojs/middleware/pull/1067#pullrequestreview-2719920080).

yusukebe (approved):
LGTM!

PR #1069, opened 2025-03-26
build(zod-openapi): lint published package

BarryThePenguin:


packages/zod-openapi/vitest.config.ts
```diff
@@ -4,7 +4,7 @@ export default defineProject({
   test: {
     globals: true,
     typecheck: {
-      tsconfig: './tsconfig.vitest.json',
+      enabled: true,
```
BarryThePenguin, line 7: Note that `vitest` now runs as if `--typecheck` was used

packages/zod-openapi/vitest.config.ts
```diff
@@ -4,7 +4,8 @@ export default defineProject({
   test: {
     globals: true,
     typecheck: {
-      tsconfig: './tsconfig.vitest.json',
+      tsconfig: './tsconfig.json',
+      enabled: true,
```
BarryThePenguin, line 8: Note that `vitest` now runs as if `--typecheck` was used

yusukebe (approved):
LGTM!

PR #1076, opened 2025-03-27
ci: use node v20

BarryThePenguin:


BarryThePenguin:
Is it worth also updating [`engines.node`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) too? I'm unsure how useful that would be..

yusukebe:
@BarryThePenguin 

> Is it worth also updating [`engines.node`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) too? I'm unsure how useful that would be..

The Node.js version in CI should currently be a general version like `20.x` or `22.x`. But we don't have to change the `engines.node` since it depends on each package, and the current specifications are fine.

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Thanks! Merging now.

PR #1077, opened 2025-03-28
build: typescript project references

BarryThePenguin:


yusukebe:
Hi @BarryThePenguin 

I may be wrong, but even though it does not throw errors when I run the `yarn typecheck` command, it throws errors when I run `yarn build`. My understanding is that it should throw errors even with `yarn typecheck`. What do you think? The following is an example of `@hono/node-ws`.

CleanShot 2025-03-30 at 18 06 48@2x

BarryThePenguin:
`@hono/node-ws` is a tricky one as it uses types from `@types/ws`. From what I understand, the `typecheck` command and `tsup` might be resolving these types from different places which is why they give different results. Thanks for taking a look!

packages/ajv-validator/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist",
+    "tsBuildInfoFile": "dist/tsconfig.build.tsbuildinfo",
+    "emitDeclarationOnly": false
```
BarryThePenguin, line 7: We can change this as we use `tsup` to build the output.

One benefit of having `tcs` emit declaration files during the build is it can generate [declaration maps](https://www.typescriptlang.org/tsconfig/#declarationMap). They enable features like "Go to Definition" in tools like vscode. There are some conversations as to if these files are [worth publishing to npm](https://github.com/egoist/tsup/issues/885), `tsup` also [mentions it in their documentation](https://github.com/egoist/tsup/blob/main/docs/README.md#generate-typescript-declaration-maps-dtsmap).

There are also a few issues about this:
- https://github.com/egoist/tsup/issues/571
- https://github.com/egoist/tsup/issues/647

packages/ajv-validator/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
...
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": ["**/*.test.ts"],
```
BarryThePenguin, line 10: Only include src files in the build, while excluding `.test.ts` files.

The big benefit of this is it prevents test related globals from appearing as valid references in src files, eg; `describe`, `test`, `expect` from `vitest`

packages/ajv-validator/tsconfig.spec.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc/packages/ajv-validator",
+    "types": ["vitest/globals"]
```
BarryThePenguin, line 5: Allow `.test.ts` files to reference `vitest/globals`

packages/ajv-validator/tsconfig.spec.json
```diff
@@ -0,0 +1,13 @@
...
+  },
+  "include": ["**/*.test.ts"],
```
BarryThePenguin, line 7: Only include `.test.ts` files in this project

packages/ajv-validator/tsconfig.spec.json
```diff
@@ -0,0 +1,13 @@
...
+  "references": [
+    {
+      "path": "./tsconfig.build.json"
+    }
+  ]
```
BarryThePenguin, line 12: Along with a reference to the project that includes the src files

tsconfig.tsup.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "composite": false,
+    "jsx": "react",
+    "types": ["@cloudflare/workers-types", "node", "ws"]
```
BarryThePenguin, line 6: `tsup` struggles to understand composite projects, so it's simpler to just disable them

It also needs a reference to any types needed during build, though the `jsx` configuration can probably just go in `tsconfig.base.json` 🤔

BarryThePenguin (commented):
I'm unsure what is wrong with `@hono/node-ws`... it looks like it might be an existing issue? Though I can't reproduce it outside of `tsup` 😩 

I noticed there were a few older issues like honojs/hono#3327, might be related?

yusukebe:
Hey @BarryThePenguin 

The `@hono/node-ws` issue may be fixed by https://github.com/honojs/hono/pull/4036.

yusukebe:
Oops. Accidentally closed. Reopened.

yusukebe:
Hi @BarryThePenguin 

The `@hono/node-ws` issue has been resolved by https://github.com/honojs/middleware/pull/1094 (Thanks @nakasyou!)

BarryThePenguin:
Awesome, looks good now 👍🏻 Thanks both!

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Looks good! I'll merge this now. Thank you!

PR #1083, opened 2025-03-29
ci: run eslint in each workflow

BarryThePenguin:
Rather than running eslint over the entire monorepo, I think it makes sense just to lint the workspaces that changed


package.json
```diff
@@ -62,6 +62,7 @@
     "@changesets/cli": "^2.26.0",
     "@cloudflare/vitest-pool-workers": "^0.7.8",
     "@cloudflare/workers-types": "^4.20230307.0",
+    "@hono/eslint-config": "workspace:*",
```
BarryThePenguin, line 65: Need this here so that `yarn workspaces focus hono-middleware @hono/...` will also install `@hono/eslint-config`

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

> Rather than running eslint over the entire monorepo, I think it makes sense just to lint the workspaces that changed

I also think the same thing! It takes time to lint the entire monorepo. Thanks!

Issue #990, opened 2025-03-01
Meta: monorepo maintenance


BarryThePenguin:
Hi 👋🏻 would the Hono devs be interested in reviewing contributions to the configuration and maintenance of the monorepo? Some ideas I have include..

- [x] #994
- [x] #999
- [x] #1003
- [x] #1007
- [x] #1020
- [x] align `build`/`test`/`prerelease`/`release` scripts in workspaces
- [x] include [`@arethetypeswrong/cli`](https://github.com/arethetypeswrong/arethetypeswrong.github.io) in the publish/release process
- [x] #1090
- [x] add type checking
- [x] removing `@eslint/eslintrc` and the legacy `FlatCompat`
- [x] upgrading `eslint` plugins
- [x] [cleanup migration](https://typescript-eslint.io/packages/typescript-eslint/#migrating-from-legacy-eslintrc-configs) to `typescript-eslint`
- [x] cleanup duplicate dependencies between workspaces
- [x] review eslint rules
- [ ] #628

I'm happy to volunteer, I'll aim to keep changes small and focused. I'm just gauging the interest of others to review the changes, and if these kinds of changes would be useful for the project 👀

yusukebe:
Hi @BarryThePenguin !

I am super interested in your proposal! The maintainability will be greatly improved.

All 5 thins you picked up are good to do. Regarding monorepo, we were discussing moving package manager to Bun and using Bun's workspace. https://github.com/honojs/middleware/issues/628

It may be good to start with migration using Bun.

cc: @gunta @nakasyou

nakasyou:
That's a good idea! However to use Bun, we need to match all packages with the latest Hono.

yusukebe:
> However to use Bun, we need to match all packages with the latest Hono.

This will be a pain point. Ideally, all packages should pass the test using the latest version of Hono, but I think that's difficult.

It might be better to continue using `yarn`. We could use `pnpm`, but since we don't use it in any of the other projects in `honojs`, I think it's fine to use `yarn`.

BarryThePenguin:
Ok, I think sticking with `yarn` for the short term would work best. With a longer term goal of adopting `bun`, once some of the more pressing maintenance issue have been resolved.

> However to use Bun, we need to match all packages with the latest Hono.

I'll keep this in mind 👍🏻

yusukebe:
@BarryThePenguin Thanks! Can you proceed with it?

BarryThePenguin:
[...]

PR #1086, opened 2025-03-29
ci: run tests with hono resolution matrix

BarryThePenguin:
Run CI against multiple versions of hono using [matrix strategy](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow)

### DIFF
--- a/.github/workflows/ci-zod-openapi.yml
+++ b/.github/workflows/ci-zod-openapi.yml
@@ -12,13 +12,17 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        hono-version: ['hono@^4.0.0']
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 20.x
+      - run: yarn add ${{ matrix.hono-version }} --mode=update-lockfile
       - run: yarn workspaces focus hono-middleware @hono/zod-openapi
-      - run: yarn workspace @hono/zod-openapi build
+      - run: yarn workspaces foreach --topological --recursive --from @hono/zod-openapi run build
       - run: yarn workspace @hono/zod-openapi publint
       - run: yarn eslint packages/zod-openapi
       - run: yarn test --coverage --project @hono/zod-openapi
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@typescript-eslint/parser": "^8.7.0",
     "@vitest/coverage-istanbul": "^3.0.8",
     "eslint": "^9.17.0",
+    "hono": "*",
     "npm-run-all2": "^6.2.2",
     "prettier": "^2.7.1",
     "tsup": "^8.4.0",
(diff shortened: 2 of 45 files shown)


.github/workflows/ci-zod-openapi.yml
```diff
@@ -12,13 +12,17 @@ on:
...
+        hono-version: ['hono@^4.0.0']
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 20.x
+      - run: yarn add ${{ matrix.hono-version }} --mode=update-lockfile
       - run: yarn workspaces focus hono-middleware @hono/zod-openapi
-      - run: yarn workspace @hono/zod-openapi build
+      - run: yarn workspaces foreach --topological --recursive --from @hono/zod-openapi run build
```
BarryThePenguin, line 25: Recessively build the dependencies of the current workspace

This is specifically for `@hono/zod-openapi` as it has a dependency on `@hono/zod-validator` which needs building first

package.json
```diff
@@ -68,6 +68,7 @@
     "@typescript-eslint/parser": "^8.7.0",
     "@vitest/coverage-istanbul": "^3.0.8",
     "eslint": "^9.17.0",
+    "hono": "*",
```
BarryThePenguin, line 71: Setting a single version in the root is enough to satisfy the peer dependencies of each workspace. It also ensures that there's only a single version installed at any one time.

packages/ajv-validator/package.json
```diff
@@ -44,7 +44,6 @@
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.17.4",
     "ajv": ">=8.12.0",
-    "hono": "^4.4.12",
```
BarryThePenguin, line 47: Can remove any reference to hono in dev dependencies as the single reference in the root package.json is enough to satisfy the peer dependency

packages/zod-openapi/package.json
```diff
@@ -52,7 +51,7 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^7.1.0",
-    "@hono/zod-validator": "npm:0.4.2"
+    "@hono/zod-validator": "workspace:^"
```
BarryThePenguin, line 54: Now that CI builds dependencies of a workspace, the `workspace:` protocol can be used to reference `@hono/zod-validator` directly instead of from npm

BarryThePenguin (commented):
Sorry for the all of the CI failures earlier @yusukebe , I had meant to create this PR in my own fork first 😅 

If you think this PR will be useful, and a good direction to take, I can update the remaining workflows.

Due to existing issue with different workspaces using different versions of hono, I'm happy  to focus on the workspaces that work, and fix the failing types or tests in future PRs

yusukebe:
@BarryThePenguin 

In my opinion, this PR is unfortunately unnecessary because middleware basically only needs to support `4.x`. There is no need to support older major versions, though we do not have a clearly documented support policy. I think what we need to do is to use `4.x` in projects that use `3.x`.

BarryThePenguin:
Ah.. you're right. We even discussed it in the original issue https://github.com/honojs/middleware/issues/990#issuecomment-2692683615 😅 

I'll focus on `4.x` 👍🏻

yusukebe:
@BarryThePenguin 

Thank you so much for understanding!

PR #1089, opened 2025-03-30

BarryThePenguin:


BarryThePenguin:
Currently waiting on https://github.com/oven-sh/bun/issues/21088

Or need to find another way of syncing `version` from `package.json` with `deno.json` 🤔

PR #1092, opened 2025-03-30
chore(dev-deps): upgrade to hono v4

BarryThePenguin:
Updates the monorepo to use the root `package.json` to install `hono`. This ensures a single version is installed, in this case `v4.x`

I've been able to take what I've learned in other PRs and apply that here 🎉


.github/workflows/ci-zod-openapi.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           node-version: 20.x
       - run: yarn workspaces focus hono-middleware @hono/zod-openapi
-      - run: yarn workspace @hono/zod-openapi build
+      - run: yarn workspaces foreach --topological --recursive --from @hono/zod-openapi run build
```
BarryThePenguin, line 21: Recursively builds the dependencies of the current workspace

This is specifically for `@hono/zod-openapi` as it has a dependency on `@hono/zod-validator` which needs building first

package.json
```diff
@@ -8,41 +8,6 @@
...
-    "build:node-ws": "yarn workspace @hono/node-ws build",
-    "build:react-compat": "yarn workspace @hono/react-compat build",
-    "build:effect-validator": "yarn workspace @hono/effect-validator build",
-    "build:conform-validator": "yarn workspace @hono/conform-validator build",
-    "build:casbin": "yarn workspace @hono/casbin build",
-    "build:ajv-validator": "yarn workspace @hono/ajv-validator build",
-    "build:tsyringe": "yarn workspace @hono/tsyringe build",
-    "build:cloudflare-access": "yarn workspace @hono/cloudflare-access build",
-    "build:standard-validator": "yarn workspace @hono/standard-validator build",
-    "build:otel": "yarn workspace @hono/otel build",
```
BarryThePenguin, line 45: I'm 90% sure these aren't used anywhere 👀

packages/arktype-validator/src/index.ts
```diff
@@ -29,6 +29,7 @@ export const arktypeValidator = <
   schema: T,
   hook?: Hook<T['infer'], E, P>
 ): MiddlewareHandler<E, P, V> =>
+  // @ts-expect-error not typed well
```
BarryThePenguin, line 32: Looking at the other `@hono/*-validator` packages, this is a common issue that is fixed with this directive

packages/trpc-server/src/index.ts
```diff
@@ -44,7 +44,7 @@ export const trpcServer = ({
               return Reflect.get(t, p, t)
             },
           }),
-    }).then((res) => c.body(res.body, res))
+    }).then((res) => c.newResponse(res.body, res))
```
BarryThePenguin, line 47: The types for `c.body` don't allow the combination of..

```ts
(data: Data | null, init?: ResponseOrInit): Response;
```

..but `c.newResponse` does, both end up calling `#newResponse`

https://github.com/honojs/hono/blob/0bb821a4a10053f82a6afac09a1243fd2da62ad4/src/context.ts#L730-L738

https://github.com/honojs/hono/blob/0bb821a4a10053f82a6afac09a1243fd2da62ad4/src/context.ts#L707

package.json
yusukebe: You are correct. It's okay to remove them!

packages/trpc-server/src/index.ts
yusukebe: Both `c.body` and `c.newResponse` are okay, but `c.body` is better since `c.newResponse` is not often used as a public API. So how about adding `@ts-expected-error` like this (I think it's not a big problem adding `@ts-expect-error` or `@ts-ignore` internal implementation)?

```ts
 // @ts-expect-error c.body accepts both ReadableStream and null but is not typed well
```

BarryThePenguin: Great! Done 👍🏻

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Perfect! Let's go with this. Thank you!

PR #1093, opened 2025-04-01
chore: use bun

nakasyou:
This PR switches the package manager used in this project to Bun.

### Why migrate to Bun?

One reason for migrating is the faster installation speed, but that’s not the primary motivation. The main benefit of using Bun is that it helps save disk space. Yarn creates a `node_modules` directory for every workspace, even when packages share the same version. In contrast, Bun typically creates a `node_modules` directory only at the project root, significantly reducing disk space usage.

### Addressing Version Mismatch Issues

[There were issues where Bun caused version mismatches, leading to test failures.](https://github.com/honojs/middleware/issues/990#issuecomment-2692598450)

These failing tests have been resolved in this PR. Most test failures stemmed from type mismatches. For example, the current version of Hono includes a TypeScript property, `status`, to improve user experience, which didn’t exist in older versions. Adding such TypeScript properties caused issues when test code in this project used `Equal`, as shown below:

```typescript
type Actual = ExtractSchema<typeof route>
type Expected = {
  '/': {
    $post: {
      input: {}
      output: {}
    }
  }
}
type verify = Expect<Equals<Actual, Expected>>
```

In this PR, the issue is resolved by using `expectTypeOf<Actual>().toMatchObjectType<Expected>()`. This approach doesn’t require exact equality, only compatibility between types.

Additionally, with Bun, you can specify a particular version of Hono in `devDependencies`. If a different Hono version is listed in `devDependencies`, Bun creates a separate `node_modules` directory to maintain compatibility.

### The Author Should Do the Following, if Applicable

- [ ] Add tests
- [x] Run tests
- [ ] Run `yarn changeset` at the root of this repo and push the changeset


nakasyou:
Duplicate #628

PR #1098, opened 2025-04-02
feat(eslint-config): enable linting with type information

BarryThePenguin:
Updates the `@hono/eslint-config` to include `typescript-eslint` presets for [linting with type information](https://typescript-eslint.io/getting-started/typed-linting/)


eslint.config.mjs
```diff
@@ -1,6 +1,62 @@
...
-export default defineConfig(globalIgnores(['.yarn', '**/dist']), {
+export default defineConfig(globalIgnores(['.yarn', '**/coverage', '**/dist']), {
   extends: baseConfig,
+
+  languageOptions: {
+    parserOptions: {
+      projectService: true,
+      tsconfigRootDir: import.meta.dirname,
+    },
+  },
```
BarryThePenguin, line 12: This tells the `typescript-eslint` parser where to find the root `tsconfig.json`, and to use project references

eslint.config.mjs
```diff
@@ -1,6 +1,62 @@
...
+
+  linterOptions: {
+    reportUnusedDisableDirectives: 'error',
+    reportUnusedInlineConfigs: 'error',
+  },
```
BarryThePenguin, line 17: Additional [linter options](https://eslint.org/docs/latest/use/configure/configuration-files#configuring-linter-options)

`reportUnusedDisableDirectives` defaults to `'warn'` and `reportUnusedInlineConfigs` defaults to `'off'`

eslint.config.mjs
```diff
@@ -1,6 +1,62 @@
...
+    '@typescript-eslint/prefer-nullish-coalescing': 'off',
+    '@typescript-eslint/prefer-optional-chain': 'off',
+    '@typescript-eslint/prefer-regexp-exec': 'off',
+    '@typescript-eslint/prefer-return-this-type': 'off',
+    '@typescript-eslint/require-await': 'off',
+    '@typescript-eslint/restrict-plus-operands': 'off',
+    '@typescript-eslint/restrict-template-expressions': 'off',
+    '@typescript-eslint/unbound-method': 'off',
+    '@typescript-eslint/unified-signatures': 'off',
+  },
```
BarryThePenguin, line 61: I've initially disabled all the new rules to keep this PR small

I'd be happy to revisit each rule individually and decide if it's worth enabling or not. If a new rule is problematic, it can always be disabled in the shared eslint config

My personal preference is to have less configuration, and rely on tooling to make these decisions for me. But I understand that doesn't fit with how everyone works 😅

packages/eslint-config/index.js
```diff
@@ -7,7 +7,8 @@ import tseslint from 'typescript-eslint'
 export default [
   js.configs.recommended,
   nodePlugin.configs['flat/recommended'],
-  ...tseslint.configs.recommended,
+  ...tseslint.configs.strictTypeChecked,
+  ...tseslint.configs.stylisticTypeChecked,
```
BarryThePenguin, line 11: There's a few [shared configs](https://typescript-eslint.io/users/configs) to choose from.

Their suggestion is to enabling `recommended-type-checked` and `stylistic-type-checked` to start with. Then replacing  `recommended-type-checked` with `strict-type-checked` if you're comfortable working with TypeScript.

packages/graphql-server/tsconfig.bun.json
BarryThePenguin, line 1: Includes `bun_test` files in the project

tsconfig.repo-config-files.json
BarryThePenguin, line 1: Includes project configuration files

eslint.config.mjs
yusukebe: > I've initially disabled all the new rules to keep this PR small

I like this approach. Fixing lint errors is troublesome, and the diffs will be huge. We don't have to do it.

.changeset/fifty-aliens-rhyme.md
```diff
@@ -0,0 +1,10 @@
+---
+'@hono/eslint-config': major
```
yusukebe, line 2: Yes. This change includes breaking change, so this is a `major` version up.

yusukebe:
@BarryThePenguin 

Looks good! For this PR, let's just update `@hono/eslint-config` and fix the other lint errors later as you said.

.changeset/fifty-aliens-rhyme.md
BarryThePenguin: Feel free to ping me if you need help upgrading other projects that use `@hono/eslint-config`

PR #1117, opened 2025-04-09
ci(release): yarn config set npmAuthToken

BarryThePenguin:
Looks like `changesets/action` was creating an `.npmrc` file using `$NPM_TOKEN`

This adds an extra step to the release to add `$NPM_TOKEN` to the `.yarnrc.yml`

I tried including the following line in the `.yarnrc.yml` file

```
npmAuthToken: ${NPM_TOKEN}
```

but that causes local yarn commands to fail due to the environment variable not being present


BarryThePenguin:
Sorry for the trouble!

yusukebe:
@BarryThePenguin 

No problem! It succeeded. We can sleep in peace!

PR #1126, opened 2025-04-16
feat(oidc-auth): support empty OIDC client secret

tarasglek:
Hi,
I found that this library does not work with https://lastlogin.net/developers/

Would a change like this be ok?

### DIFF
--- a/.changeset/every-pugs-wave.md
+++ b/.changeset/every-pugs-wave.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': minor
+---
+
+Support empty OIDC_CLIENT_SECRET
--- a/packages/oidc-auth/src/index.ts
+++ b/packages/oidc-auth/src/index.ts
@@ -167,11 +167,18 @@ export const getClient = (c: Context): oauth2.Client => {
   const env = getOidcAuthEnv(c)
   let client = c.get('oidcClient')
   if (client === undefined) {
-    client = {
-      client_id: env.OIDC_CLIENT_ID,
-      client_secret: env.OIDC_CLIENT_SECRET,
-      token_endpoint_auth_method: 'client_secret_basic',
-    }
+    client =
+      env.OIDC_CLIENT_SECRET === ''
+        ? {
+            // No client secret provided, use 'none' auth method
+            client_id: env.OIDC_CLIENT_ID,
+            token_endpoint_auth_method: 'none',
+          }
+        : {
+            client_id: env.OIDC_CLIENT_ID,
+            client_secret: env.OIDC_CLIENT_SECRET,
+            token_endpoint_auth_method: 'client_secret_basic',
+          }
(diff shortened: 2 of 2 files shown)


yusukebe:
@tarasglek 

Can you run `yarn changeset` on the top of the project and push the generated file?

yusukebe:
Hi @hnw! Can you review this?

tarasglek:
I ran yarn changeset, hope I did this right

.changeset/every-pugs-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/oidc-auth': major
```
yusukebe, line 2: It will be a minor update. So, please replace `major` with `minor`.

tarasglek:
all good now?

tarasglek:
ping to get this landed

yusukebe:
@tarasglek 

Can you write a proper test?

hnw:
@tarasglek
Hi, thank you for the pull request! Sorry for the delay in reviewing.
I also appreciate you reporting the specific issue regarding integration with LastLogin.

I understand that your proposal aims to support the `token_endpoint_auth_method:'none'` (no client secret) case. I understand that such a configuration might be necessary for providers primarily targeting Public Clients (like web frontends or mobile apps), such as LastLogin.

On the other hand, the `oidc-auth` middleware is currently designed and optimized for **Confidential Clients**, which typically run on a server and can securely store a client secret. For Confidential Clients, security best practices for OAuth 2.0 and OpenID Connect strongly recommend using client authentication methods like `client_secret_basic` or `client_secret_post`. This is crucial to prevent client impersonation. Therefore, simply adding an option to forgo client authentication (`none`) in a middleware intended for Confidential Clients raises security concerns for us.

Regarding this point, I would also like to investigate further, for instance by checking LastLogin's documentation for their recommended settings when used as a Confidential Client. If you happen to know of any information (like documentation) from LastLogin regarding recommended settings for Confidential Clients or security considerations when using `none`, sharing it would be very helpful.

We would appreciate further discussion on whether there's a safe way to support `none` in this middleware, or how we should consider the scope of the middleware going forward.

tarasglek:
I am a little embarrassed to note, that lastlogin does work if I put in a random non-blank client_secret. Maybe @anderspitman, could change docs to put  in any value in client_secret and this is a lastlogin instruction bug.

anderspitman:
Yeah that makes sense @tarasglek. I opened an issue for it.

@hnw re: confidential clients. LastLogin is almost entirely stateless (other than rate limiting), so I'm not really sure of a good way to persist a client_secret. Since PKCE is considered good enough for public clients, I've always considered it good enough for all clients. Obviously a secret adds defense in depth, but I don't think it's strictly necessary. If you have anything I can read that contradicts that, I'd be happy to look at it.

PS - thanks for your work on hono!

PR #1133, opened 2025-04-22
fix(zod-openapi): Run validators before route middleware

kevin-dp:
We're using `zod-openapi` and found that route middleware runs before the validators. As a result, we don't have access to validated input (like `c.req.valid('param')`) inside route middleware, which was somewhat surprising to us.

I can imagine special use cases where one might want route middleware to run first (e.g. when dynamically adding parameters) but i would imagine that to be less common. If we really want we could make this configurable.

### DIFF
--- a/.changeset/some-friends-tie.md
+++ b/.changeset/some-friends-tie.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Run validators before route middleware.
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -550,8 +550,8 @@ export class OpenAPIHono<
     this.on(
       [route.method],
       route.path.replaceAll(/\/{(.+?)}/g, '/:$1'),
-      ...middleware,
       ...validators,
+      ...middleware,
       handler
     )
     return this


yusukebe:
@kevin-dp 

You can access it with the code like the following:

```ts
import { z, createRoute, OpenAPIHono } from '@hono/zod-openapi'

const createUserRoute = createRoute({
  method: 'post',
  path: '/user',
  request: {
    body: {
      content: {
        'application/json': {
          schema: z.object({
            name: z.string(),
          }),
        },
      },
    },
  },
  middleware: [
    async (c, next) => {
      await next()
      // @ts-expect-error not typed well
      console.log(c.req.valid('json')) // You can get the validated value
    },
  ] as const,
  responses: {
    200: {
      description: 'foo',
    },
  },
})

const app = new OpenAPIHono().openapi(createUserRoute, async (c) => {
  return c.json('foo')
})

export default app
```

kevin-dp:
Hi @yusukebe,

Our middleware is authenticating the request. Calling `next()` at the beginning of our middleware would allow us to access the validated inputs inside the middleware but would mean that the route handler is being executed before we actually authenticate the request.

yusukebe:
@kevin-dp 

Is that possible with your PR? I don't know what code you want to implement but you may be able it with the `hook`:

https://github.com/honojs/middleware/tree/main/packages/zod-openapi#handling-validation-errors

If you want to get more detailed help, please provide a concrete and minimal code.

yusukebe:
@kevin-dp 

Thank you for the explanation. I got it. But this will introduce a breaking change, although the test is not failing. And this is not a `fix` issue, but a new feature. We can't accept this PR as is now.

kevin-dp:
@yusukebe sure, i understand that. What is the reason for running the validators after the middleware? I was very surprised by that behavior.

yusukebe:
Plus, we have historical reasons. This Zod OpenAPI is based on Zod Validator or other validators. The old version of `hono` can't infer types in a handler when there is middleware after the validator.

CleanShot 2025-04-28 at 17 19 40@2x

So, reference implementations use middleware before the validator. So, this Zod OpenAPI is following it, though, as you said, in some cases it is convenient if we put middleware after validators.

yusukebe:
I think we can close this now.

PR #1136, opened 2025-04-24
fix(auth-js): Missing request body when verifying csrf token in production

Asweh-Debrej:
* refactored `reqWithEnvUrl` to be async and support reading request body
* added safe cloning using `req.text()` to allow downstream reuse
* updated all call sites to await the new async `reqWithEnvUrl`

## Why the change?

When handling requests with `x-www-form-urlencoded` content type, the request body may be prematurely closed before being fully read. This causes the body to be lost when cloning the original `Request` in `reqWithEnvUrl` before calling `req.text()`. This happens specificaly when I was trying to verify csrf token. I found this issue when deploying a server using `@hono/auth-js` on **Render**, while everything worked as expected in localhost. This confused me for 3 days which I found the root after dozens of log statements.

### Before
- Test Files  2 failed | 42 passed (44)
- Tests  447 passed (447)
- Type Errors  no errors

### After (no test added)
- Test Files  2 failed | 42 passed (44)
- Tests  447 passed (447)
- Type Errors  no errors

### DIFF
--- a/.changeset/plenty-plants-drop.md
+++ b/.changeset/plenty-plants-drop.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': minor
+---
+
+fix missing body in some production instances when using x-www-form-urlencoded
--- a/packages/auth-js/src/index.ts
+++ b/packages/auth-js/src/index.ts
@@ -36,7 +36,7 @@ export function setEnvDefaults(env: AuthEnv, config: AuthConfig) {
   coreSetEnvDefaults(env, config)
 }
 
-export function reqWithEnvUrl(req: Request, authUrl?: string) {
+export async function reqWithEnvUrl(req: Request, authUrl?: string) {
   if (authUrl) {
     const reqUrlObj = new URL(req.url)
     const authUrlObj = new URL(authUrl)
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @Asweh-Debrej 

Thank you for the PR. Can you share a test that is failing with the current implementation?

Asweh-Debrej:
The current PR may cause runtime errors. I tried replacing `req` with `{ ...req }` or `req,clone()` in `new Request(url, req)` below my changes inside `reqWithEnvUrl` it causes the test to fail.

Also I didn't do `content-type` checking in my current change, which may cause another problem.

I guess I have to use a workaround by using json instead for now

yusukebe:
@Asweh-Debrej 

Thank you for the comment. But it is difficult for me to reproduce it.

Can you write tests for this PR? This means it fails with the current version and will pass with this PR. We can't accept the PR without proper tests.

Asweh-Debrej:
I found the issue.
It’s not related to `@hono/auth-js`; the problem was with Bun not allowing reading ReadableStream after request cloning until version 1.1.27.

Render defaults to Bun 1.1.0, so I just needed to set the environment variable `BUN_VERSION` to `1.2.10`.

PR #1141, opened 2025-04-27
fix(node-ws): make adapter uncrashable

### LINKED ISSUE #1132: [@hono/node-ws] Crashes whole app on exception

@hono/node-ws

1.1.1

4.7.6

Node 23.1.0

throw Error or HTTPException somewhere inside `upgradeWebSocket`

Example is in **Additional information** section

Display an error and don't crash whole app

### What do you see instead?

_No response_

### Additional information

Even when I try to catch this error it doesn't catches

Example code:
```ts
app
  .use((c, next) => {
    // Try to catch 1.
    let response;
    try {
      response = next();
    } catch (e) {
      console.error(e);
    }

    return response;
  })
  .onError((err, c) => {
    // Try to catch 2.
    if (err instanceof HTTPException) {
      return c.text(err.message, err.status);
    }

    return c.text("Internal Server Error", 500);
  })
  .get(
    "/ws",
    upgradeWebSocket((c) => {
      throw new HTTPException(400); // Crashes whole app

      return {
        onOpen() {
          throw new Error("test"); // Crashes whole app
        },
      };
    })
  );
```

nakasyou:
fixes #1132 

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @nakasyou 

Thank you for the PR!

I've confirmed this PR enables the app not to crash when throwing an `Error` in `upgradeWebSocket`. But the app can't handle the error in the middleware of the app and `app.onError` described in https://github.com/honojs/middleware/issues/1132#issue-3009336599

```ts

app
  .use((c, next) => {
    let response
    try {
      response = next()
    } catch (e) {
      console.error(e) // not called
    }

    return response
  })
  .onError((err, c) => {
    // not called
    if (err instanceof HTTPException) {
      return c.text(err.message, err.status)
    }
  })
  .get(
    '/ws',
    upgradeWebSocket((c) => {
      throw new HTTPException(400)
    })
  )
```

So, should I write an `onError` event in `upgradeWebSocket`? If so, this means we can't handle the app in the middleware or `app.onError`?

.changeset/nice-rice-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/node-ws': patch
```
yusukebe, line 2: I've replaced `minor` with `patch`. But if you add a new option, it should be `minor`.

nakasyou:
Hi @yusukebe

Regarding:
> So, should I write an onError event in upgradeWebSocket? If so, this means we can't handle the app in the middleware or app.onError?

Exactly. As you mentioned, Hono's current `app.onError` is an error handler for HTTP requests and is designed to ultimately return an HTTP response. If an error occurs during the WebSocket upgrade process (during the handshake), even if `app.onError` could catch it, there's an issue where it's difficult, or there's no standard way, to properly convey the `Response` object returned by that handler to the client as a response indicating WebSocket connection failure. It needs to return an HTTP-level error response (e.g., 400 Bad Request), and the flexible response generation of `app.onError` doesn't necessarily fit this requirement.

Furthermore, if an error is thrown within event handlers *after* the WebSocket connection is established, such as `onOpen` / `onMessage` / `onClose` / `onError` (the WebSocket event `onerror`), these are executed outside of Hono's HTTP request/response processing cycle. Therefore, these errors cannot be propagated to Hono's middleware layer or `app.onError` for handling.

Consequently, error handling during the WebSocket upgrade and after the connection is established likely needs to be addressed separately from Hono's standard error handling mechanisms (try-catch in middleware or `app.onError`). This would probably need to happen within the callback passed to `upgradeWebSocket`, or potentially via specific options for `upgradeWebSocket` that might be introduced in the future (like an `onError` option, as you suggested).

.changeset/nice-rice-agree.md
nakasyou: Thank you for fixing!

yusukebe:
@nakasyou 

Thank you for your kind explanation. I could understand it completely. Okay! I'll merge this and release a new version.

PR #1151, opened 2025-05-07
feat(otel): Added support for W3C Trace Context format

nrdobie:
One of Open Telemetry's main feature is the ability to trace across micro services, this updates the `@hono/otel` package to check if `traceparent` is set and if so then set the context for the span to match accordingly.

### DIFF
--- a/packages/otel/src/index.ts
+++ b/packages/otel/src/index.ts
@@ -8,7 +8,7 @@ import {
   ATTR_URL_FULL,
   ATTR_HTTP_ROUTE,
 } from '@opentelemetry/semantic-conventions'
-import type { Env, HonoRequest, Input } from 'hono'
+import type { Env, Input } from 'hono'
 import { createMiddleware } from 'hono/factory'
 import metadata from '../package.json' with { type: 'json' }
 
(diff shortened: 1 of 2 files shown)


packages/otel/src/index.ts
```diff
@@ -8,7 +8,7 @@ import {
   ATTR_URL_FULL,
   ATTR_HTTP_ROUTE,
 } from '@opentelemetry/semantic-conventions'
-import type { Env, Input } from 'hono'
+import type { Env, HonoRequest, Input } from 'hono'
```
yusukebe, line 11: This is change is not necessary. `HonoRequest` is not used.

yusukebe:
@nrdobie Thank you for the PR.

Hi @dahlia, can you review this?

dahlia (approved):

yusukebe:
@dahlia Thanks!

packages/otel/src/index.ts
nrdobie: Oops 🙃

yusukebe (commented):
LGTM!

yusukebe:
Thank you @nrdobie ! I'll merge and release a new version.

PR #1153, opened 2025-05-08
feat(bun-compress): Created a compress middleware that supports Bun

nrdobie:
Bun does not currently support the [CompressionStream API](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream) so this middlware replicates the behavior of [`hono/compress`](https://hono.dev/docs/middleware/builtin/compress) using the Zlib library.

### DIFF
--- a/packages/bun-compress/package.json
+++ b/packages/bun-compress/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@hono/bun-compress",
+  "version": "0.0.0",
+  "description": "An example of third-party middleware for Hono",
+  "type": "module",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup ./src/index.ts",
+    "prepack": "yarn build",
+    "publint": "attw --pack && publint",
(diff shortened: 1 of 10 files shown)


yusukebe:
Interesting! I'll check it later.

packages/bun-compress/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/hello"
```
yusukebe, line 38: `packages/hello` should be `packages/bun-compress`.

yusukebe:
Hi @nrdobie 

This is great! I think the implementation is perfect. I commented on one thing.

packages/bun-compress/package.json
nrdobie: Done 👍

yusukebe (approved):
LGTM!

yusukebe:
@nrdobie 

Thanks! Looks good. Let's go with it!

PR #1173, opened 2025-05-18
feat(zod-validator): support Zod v4

yusukebe:
This PR enables the Zod Validator to support the new Zod v4. This also supports v3.

~~The important thing. You need to update to `v3.25.0`, which is included in `peerDependencies`, if you want to continue using v3.~~ (Maybe you don't have to do anything)

Fixes #1148

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `yarn changeset` at the top of this repo and push the changeset
- [ ] Follow [the contribution guide](https://github.com/honojs/middleware?tab=readme-ov-file#how-to-contribute)

PR #1174, opened 2025-05-18
feat(zod-validator): add support zod v4

chimame:
## Overview

Zod v4 will be released as part of Zod 3.25, so I created a new API that allows Zod v4 schemas to be used with Zod-Validator.

## Details
I created a new API for Zod v4 for the following two reasons, and designed it so that users of Zod v4 use the new API.

- The reason for separating the API for Zod v4 is that the types are different between v3 and v4. Therefore, if one API were to support both v3 and v4 types, the type definition would become too complicated, so I defined it as a separate API.
- Defining types for v3 and v4 in one API would make it impossible to use versions prior to Zod 3.25.

```ts
import { zValidator } from '@hono/zod-validator' // Or, if you use zod/v4 or zod/v4-mini, import `@hono/zod-validator/v4`.
```

## Other Changes

- I deleted some definitions that were unnecessary for testing.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @chimame!

Thank you for the PR! It's helpful. But I decided to go with #1173

I want to support both v3 and v4 using only `@hono/zod-validator`, not using something like `@hono/zod-validator/v4`. The PR #1173 supports it. Though the user should use `^3.25.0` in `peerDependencies`.

I'll close this PR after releasing the new version that includes #1173. I think your approach is good, but it doesn't fit in this case. Thank you so much!

PR #1178, opened 2025-05-29
feat: added @hono/mcp package

MathurAditya724:
Resolves https://github.com/honojs/hono/issues/4151

- [x] Add tests
- [x] Run tests


yusukebe:
@MathurAditya724 When it's ready, please ping me!

MathurAditya724:
@yusukebe you can start reviewing it.

The test `"should keep stream open after sending server notifications"` is failing if we are sending 2 notifications. Rest all of the official test for the HTTP streaming transport are passing.

yusukebe:
@MathurAditya724 

Okay! Before it, can you run `yarn format:fix` to format the code?

MathurAditya724:
Done @yusukebe

packages/mcp/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@hono/mcp",
+  "version": "0.1.0",
```
yusukebe: Can you set it to `0.0.0` since the changeset will bump the version number automatically.

packages/mcp/CHANGELOG.md
yusukebe, line 1: This file is unnecessary. Changeset will create it.

yusukebe:
Hey @MathurAditya724 

I think this is good! I tried this on Wrangler, Bun, and Node.js. These return expected values, but Wrangler throws the error, and Bun shows the message.

Wrangler:

```
[wrangler:info] POST /mcp 200 OK (4ms)
✘ [ERROR] Uncaught (in response) Error: The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/
```
https://github.com/user-attachments/assets/4b04a39a-e0c6-499b-8166-8be6019f260d

Bun:

```
yusuke $ bun run src/index.ts                                                                                                     [~/work/mcp/my-app-mcp]
Started development server: http://localhost:3000
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,612 @@
...
+  isJSONRPCResponse,
+  JSONRPCMessageSchema,
+} from '@modelcontextprotocol/sdk/types.js'
+import type { JSONRPCMessage, RequestId } from '@modelcontextprotocol/sdk/types.js'
+import type { Context } from 'hono'
+import { HTTPException } from 'hono/http-exception'
+import type { SSEStreamingApi } from 'hono/streaming'
+import { streamSSE } from './streaming'
+
+export class StreamableHTTPHonoTransport implements Transport {
```
yusukebe: I'm stumped, but I feel that `StreamableHTTPTransport` would be fine. It's the same as the name of the class that `@modelcontextprotocol/sdk/server/streamableHttp.js` is exporting, but the user specifies `@hono/mcp`, so I don't think users will be confused. What do you think?

MathurAditya724: Yeah, we should do that cause now the api of using it also similar to the official one.

yusukebe:
Hey @MathurAditya724 

Can't you really resolve the problem? https://github.com/honojs/middleware/pull/1178#issuecomment-2942410825?

Many people want this middleware and will use it after its release. Some will use it on Cloudflare Workers. If so, the throwing error is not good.

MathurAditya724:
I will be travelling today, can you give me till Sunday evening?

yusukebe:
@MathurAditya724 

Of course! Have a safe trip!

MathurAditya724:
Hey, so the streaming issue on a POST request is resolved but as the GET request is a stream, closing it causes test to fail.

MathurAditya724:
Should we also add SSE Transport? I already have the code in muppet, we can then have all the transport layers code in this package. While libs like muppet can just focus on building web standard friendly MCP server.

yusukebe:
@MathurAditya724 

> Should we also add SSE Transport? 

I thought we didn't have to have SSE, but the current Claude Desktop accepts only SSE? But it may support HTTPStreamable soon. Hmm.. Trends change quickly.

MathurAditya724:
Yeah, we can use [mcp-remote](https://www.npmjs.com/package/mcp-remote) package for the time being. I was suggesting this because I think in the long run, this package should include all the general/utils stuff. So irrespective of how you make your MCP server, you can use this package to take care of the common things you wanna do like auth, connect with transport, payments (maybe).

MathurAditya724:
Hey @yusukebe, did you got the chance to look at this?

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+        return this.handleDeleteRequest(ctx)
+      default:
+        return this.handleUnsupportedRequest(ctx)
+    }
+  }
+
+  /**
+   * Handles GET requests for SSE stream
+   */
+  private async handleGetRequest(ctx: Context) {
```
transitive-bullshit, line 87: would be nice to make these methods `protected` so we can subclass and tweak behavior

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+        stream.onAbort(() => {
+          this.#streamMapping.delete(this.#standaloneSseStreamId)
+        })
+      })
+    } catch (error) {
+      if (error instanceof HTTPException) {
+        throw error
+      }
+
+      this.onerror?.(error as Error)
```
transitive-bullshit, line 183: curious if you think hono's top-level `onError` error handler should interop with the `@hono/mcp` error handling?

it's a lil weird because the existing `hono.onError` expects to return a HTTP response, whereas within the context of the mcp handler, we wanna return a more specific jsonrpc HTTP response

yusukebe:
Okay. It's in the queue.

transitive-bullshit (approved):
this looks amazing @MathurAditya724 🔥

i was just coming to hono to investigate creating a `@hono/mcp` middleware, but you beat me to it and imho did an amazing job :)

a few high-level questions:

- thoughts on what mcp oauth support will look like? (def leave for a follow-up pr; it'll be great just to get this initial package published)
- i'm currently using the `McpAgent` from cloudflare's `agents/mcp` package, which has the pro of using a durable object to keep connections hot based on the `mcp-session-id`, but it has the downsides of 1) being tied to cloudflare and 2) using a bunch of websocket and partykit stuff under the hood which isn't really necessary to just run an mcp server from hono.
  - any thoughts on the tradeoffs between using `@hono/mcp` on CF vs using the `agents` package?
  - one thing I really don't like is that the `agents` package forces me to break out of the `hono` context and lose all the nice hono `ctx`:
```ts
    return DurableMcpServer.serve('/*', {
      binding: 'DO_MCP_SERVER'
    }).fetch(ctx.req.raw, ctx.env, executionCtx)
```
- it looks like cors is expected to be handled strictly upstream? just double checking, will `hono/cors` add the requisite cors headers to all responses that aren't handled here?
- it sucks, but i agree that this should probably support SSE in addition to streamable http due to MCP client fragmentation, at least for the time being. maybe keep this PR focused as-is and add support for SSE in a follow-on PR?

ok, that's enough of my rambling thoughts.

great work again! 🙏

packages/mcp/src/index.ts
transitive-bullshit: for my use case, for example, i'm using a `try/catch/finally` block around my mcp server's `setRequestHandler` handler, bypassing `hono.onError` which imho isn't ideal since it doesn't compose w/ the rest of the hono middleware

yusukebe:
@MathurAditya724 

> Should we also add SSE Transport?

It's a small thing, but recently I was able to use StreamableHTTP MCP in Claude Desktop (Ramen API too!), so we don't need to support SSE.

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+            error: {
+              code: -32000,
+              message: 'Conflict: Only one SSE stream is allowed per session',
+            },
+            id: null,
+          }),
+        })
+      }
+
+      return streamSSE(ctx, async (stream) => {
```
yusukebe, line 156: I found a solution for [the Wrangler errors I mentioned earlier](https://github.com/honojs/middleware/pull/1178#issuecomment-2942410825). Adding the below changes helps - sending periodic pings keeps the stream active and prevents the hanging Promise errors.

```diff
diff --git a/packages/mcp/src/index.ts b/packages/mcp/src/index.ts
index ea0a554..1c44bdd 100644
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -168,9 +168,17 @@ export class StreamableHTTPTransport implements Transport {
           ctx,
           stream,
         })
+        // Keep connection alive
+        const keepAlive = setInterval(() => {
+          if (!stream.closed) {
+            stream.writeSSE({ data: '', event: 'ping' }).catch(() => clearInterval(keepAlive))
+          }
+        }, 30000)
+
         // Set up close handler for client disconnects
         stream.onAbort(() => {
           this.#streamMapping.delete(this.#standaloneSseStreamId)
+          clearInterval(keepAlive)
         })
       })
     } catch (error) {
```

yusukebe:
Hi @transitive-bullshit !

Yes. @MathurAditya724 is doing a great thing.

I'm also thinking the same things. Cloudflare's `agents/mcp` is great, but if we use it, we can't handle the Context object well; it's not Hono-way. As you said (and Aditya may also think), it's a good idea to release the initial version first. The basic API is okay, it's neat.

packages/mcp/src/index.ts
MathurAditya724: Not a fan of this but it works 😆, at this point I just want this to be released. Adding it

transitive-bullshit: related upstream issue: https://github.com/cloudflare/workers-sdk/issues/7767

MathurAditya724:
Thank you @transitive-bullshit, for your kind words. Here are my thoughts on the questions you asked -

1. OAuth is still something I am exploring as everyone is having a different implementation, Clerk just came out with there version and we already have one from Cloudflare. My main idea about the API for this is to make it first of all in a hono-way and also flexible enough so that others can customise it for single use, tenants, etc. Still need to explore more before creating it but this is were I will be spending my next couple of weeks.
2. My idea is to add all the utility things/functions like transport, OAuth, Payments, etc in `hono/mcp`, so that irrespective of how the MCP Server is created it can run on hono. And update the `muppet` package to be an sdk for creating a MCP server and also have adapters so that it can be connected to different providers, like Cloudflare, Vercel, Netlify etc using `hono`.
3. This class has a `handleUnsupportedRequest` which respond with `'Method not allowed.'` which takes care of this.
4. I also think that having SSE would be cool, but people who will be using this will generally be building new MCP servers on HTTP Streaming, as that is the new standard and that's where they would wanna focus. But still if in future we see an interest for SSE transport, I already have a working transport in `muppet` which we can add here.

For the time-being I think let's first release a version and see how others use it or what's the general feedback, and then we can plan accordingly.

packages/mcp/src/index.ts
MathurAditya724: I don't mind, I was just following the official SDK for function naming and there states. @yusukebe what do you think? should I make it `protected`?

packages/mcp/src/index.ts
MathurAditya724: Right now, this package can only work with `hono` and I am using all hono things, just because it will make it easy for everyone to implement it in there current workflow. So for example, if you already have Sentry in `hono`, all the errors will start going to Sentry without any extra steps. And if you don't have sentry or something it will throw the error anyhow and will work with the MCP Client.

The Official SDK, is more generalised for using with nodejs and having a `try/catch` pattern makes sense. In terms of `hono/mcp` using things/functions which are already there makes more sense to me but sure if others also think the same, I will change it.

MathurAditya724:
@yusukebe Not sure why the tests are failing in the CI, they were all passing in the previous commit. Can you take a look?

MathurAditya724:
@yusukebe I think it's ready

packages/mcp/src/index.ts
yusukebe: > Not a fan of this but it works 😆

Haha. Thanks!

packages/mcp/package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/hello"
```
yusukebe: It's a nitpick, but can you replace this with `packages/mcp`?

packages/mcp/src/index.ts
yusukebe: Does this mean some will want to make a subclass of `StreamableHTTPTransport`? I think both are okay, but if it follows the official SDK, we can go with `private` now.

packages/mcp/src/index.ts
yusukebe: I think this is okay for the same reason as @MathurAditya724 .

packages/mcp/package.json
MathurAditya724: Done

MathurAditya724:
Looks like #1230 should resolve the failing test CI in this PR

BarryThePenguin:
> Looks like #1230 should resolve the failing test CI in this PR

Yup 👍🏻 if you merge `main` that should fix CI 🤞🏻

MathurAditya724:
@yusukebe Is there something which is blocking this from getting merged? I believe we have resolved all the issues in this or did I missed something

yusukebe:
> Is there something which is blocking this from getting merged?

No. It will be released at the same time as the release of `hono` v4.8.0.

yusukebe:
@MathurAditya724 Maybe tomorrow!

yusukebe (approved):
LGTM!

yusukebe:
Hi @MathurAditya724 

I've fixed the same issues, and it looks good! I'll merge and release an initial version. Thank you for the great work!

transitive-bullshit:
So stoked for this 🔥 

Thank you for all your great work @MathurAditya724 and @yusukebe 🙏

PR #1181, opened 2025-05-29
feat: introduce UA Blocker Middleware

finxol:
Create a new middleware for AI bots blocking.
Inspired by [ai.robots.txt](https://github.com/ai-robots-txt/ai.robots.txt).

This provides several options:
- Block all known AI bots
- Block all bots that we're not sure respect `robots.txt`
- Create a `robots.txt` file to be served (also offered as a middleware)

See README for details 

- [x] Add tests
- [x] Run tests


packages/ai-robots-txt/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@hono/ai-robots-txt",
+  "version": "0.1.0",
```
yusukebe: The Changeset will bump the version number automatically. So this number should be `0.0.0`.

.changeset/early-grapes-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/ai-robots-txt': major
```
yusukebe: `major` is also fine, but I think `minor` is better. It may be early to make it as version `1.x`.

packages/ai-robots-txt/.gitignore
```diff
@@ -0,0 +1 @@
+robots.json
```
yusukebe: It's difficult to decide whether to include `robots.json` or not. In my opinion, it's better to include `robots.json` in this repo. 

If you want to update `robots.json`, you will have to build and release a new version anyway (right?). It is not that much work to include `robots.json` in this project. Instead, versioning `robots.json` keeps track of the history, and allows users to track changes in `CHANGELOG.md` as well.

packages/ai-robots-txt/package.json
```diff
@@ -0,0 +1,54 @@
...
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/hello"
```
yusukebe: This should be `packages/ai-robots-txt`.

yusukebe:
Can you add a CI for this middleware, referring to https://github.com/honojs/middleware/blob/main/.github/workflows/ci-hello.yml?

yusukebe:
Hey @finxol 

This middleware is interesting! The idea is good. I've left some comments.

finxol:
@yusukebe Thanks for the feedback! I'll make the appropriate changes. Let me know if there's anything else.

packages/ai-robots-txt/.gitignore
finxol: Good points, this wasn't a strong decision, I'll remove it from gitignore.

With this, we can automatically open a PR here whenever `robots.json` changes in `ai-robots-txt/ai.robots.txt` to update it here and flag the need for a new release.

finxol: I've added a workflow to run the fetch script and open a PR to merge it if it has changed

packages/ai-robots-txt/tsconfig.json
BarryThePenguin, line 1: Make sure to include a reference for `packages/ai-robots` in the root `tsconfig.json` too

packages/ai-robots-txt/src/robots.ts
```diff
@@ -0,0 +1,21 @@
...
+  frequency: string
+  description: string
+}
+
+type BotsList = Record<keyof typeof robotsJson, BotEntry>
+
+/**
+ * Raw robots.json data, properly typed
+ */
+export const bots = robotsJson as Partial<BotsList>
```
BarryThePenguin: Is there a reason this needs to be `Partial<BotsList>`?

Would this be ok?

```suggestion
export const bots: BotsList = robotsJson
```

packages/ai-robots-txt/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @returns A string representing the PCRE-like regex group.
+ */
+function listToRegex(list: string[]): string {
+  const formatted = list.map((item) => escape(item.toUpperCase())).join('|')
+  return `(${formatted})`
+}
+
+export function blockAiBots(params = { allowRespecting: false }) {
+  return createMiddleware(async (c, next) => {
+    const userAgent = c.req.header('User-Agent')?.toUpperCase()
```
BarryThePenguin: You could avoid the cost of constructing the regular expression on each request by moving it out of the middleware handler

```suggestion
export function blockAiBots(params = { allowRespecting: false }) {
  let b

  if (params.allowRespecting) {
    b = Object.fromEntries(botslist.filter(([, value]) => value?.respect !== 'Yes'))
  } else {
    b = bots
  }

  const userAgentRegex = listToRegex(Object.keys(b))

  return createMiddleware(async (c, next) => {
    const userAgent = c.req.header('User-Agent')?.toUpperCase()
```

packages/ai-robots-txt/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup ./src/index.ts",
+    "prepack": "yarn build",
+    "publint": "attw --pack && publint",
+    "typecheck": "tsc -b tsconfig.json && echo 'Type checking completed'",
+    "test": "vitest",
+    "getrobotstxt": "node ./src/get-robots-txt.js"
```
BarryThePenguin: This package could be published with an executable ([`bin`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#bin)) that consumers call to generate their own `robots.json` file. The output would then be passed as an argument to `blockAiBots()`

That way this repository doesn't need to be concerned with keeping the contents of the file up to date. It just determines the contents of the file

yusukebe: > This package could be published with an executable ([`bin`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#bin)) that consumers call to generate their own `robots.json` file. The output would then be passed as an argument to `blockAiBots()`

Nice idea!

packages/ai-robots-txt/.gitignore
yusukebe: @finxol Thanks.

But, there is another idea. Can you check [the comment](https://github.com/honojs/middleware/pull/1181#discussion_r2134926296)?

packages/ai-robots-txt/package.json
finxol: @BarryThePenguin that’s a good idea, but then it would just be a user-agent based blocker.
Maybe we can export the generic as you suggest, but also export separately the `robots.json` file, which can then be passed to the generic blocker for easy setup?

BarryThePenguin: I'm not too worried either way. I guess it's a trade off in convenience for the package consumer vs the package maintainer. Using the scheduled workflow requires someone to review/approve/publish any changes. If the format of the robots.json file changes, that requires the maintainer to handle those changes. However, keeping those changes internal to the package is a plus for consumers as it means the API remains small.

packages/ai-robots-txt/src/index.ts
BarryThePenguin: This could be taken a step further by having the `getrobotstxt` script generate two files, one with a compiled regex using the `respect !== 'Yes'` condition, and the other with a complied regex using all the entries.

Having a static regex at build time would remove the cost of constructing the regex

packages/ai-robots-txt/package.json
finxol: I wouldn't mind the weekly update/publish for the `robots.json`. I'd prefer to keep the API simple for the user. I'll propose a new setup soon, I need to finish writing the tests.

finxol:
I've addressed all comments in the latest commit, the architecture and goal of the package is now broader than just an AI bots blocker.

The main export is now a generic User Agent based blocker, with a path export for the `robots.json` stuff that can be passed to the generic. See README file for examples.
The regex construction is moved out of the middleware, to the `uaBlocker` function. It isn't done at build time, but enables the consumer to pass any list of user agents to block.

finxol:
`publint` fails for `@hono/ua-blocker/ai-bots` on node10 because of the export path.
Do we need to support this `node10` module resolution, or can we ignore it?

BarryThePenguin:
> `publint` fails for `@hono/ua-blocker/ai-bots` on node10 because of the export path. Do we need to support this `node10` module resolution, or can we ignore it?

Yes, it can be ignored by adding `attw --pack --profile node16` to the `publint` script.

I included this change, along with some other suggestions, in finxol#1. Feel free to include them or not 😅

BarryThePenguin (approved):
Your Readme and tests are very comprehensive 👏🏻

yusukebe:
@finxol 

Wooow. Making this UA Blocker Middleware for general user agents is a great idea. Including `ai-bots` is also a good idea. If the CI passes, I'll review again, merge it, and release!

finxol:
@yusukebe Thank you for the feedback! I just want to make a small change first, I'll push it later today and I'll let you know. 

And thank you for the great work on Hono :)

finxol:
I've moved the generator script back out of `get-robots-txt.ts`. This allows us to generate from the version-controlled `robots.json` instead of the upstream file

finxol:
@yusukebe it's now ready for review :)

packages/ua-blocker/package.json
```diff
@@ -9,13 +9,13 @@
...
+    "prebuild": "yarn getrobotstxt",
     "build": "yarn prebuild && tsup ./src/index.ts ./src/ai-bots.ts",
     "prepack": "yarn build",
-    "publint": "yarn prebuild && attw --pack && publint",
-    "typecheck": "yarn prebuild && tsc -b tsconfig.json && echo 'Type checking completed'",
+    "publint": "attw --pack --profile node16 && publint",
+    "typecheck": "yarn prebuild && tsc -b tsconfig.json",
     "test": "yarn prebuild && vitest",
-    "getrobotstxt": "tsx ./src/script/get-robots-txt.ts"
+    "getrobotstxt": "node --experimental-strip-types ./script/get-robots-txt.ts"
```
yusukebe, line 18: Using the `node` command for TypeScript is amazing. But the CI fails because the workflow uses Node.js `20.x`.
https://github.com/honojs/middleware/blob/a7ee664c98b3a3089ac2f8c0df8c2b6f5470c64a/.github/workflows/ci.yml#L19

So, we have two options: 1. use `tsx` or 2. update Node.js in `ci` to `22. x`.

@BarryThePenguin What do you think of it?

.github/workflows/ci-ua-blocker.yml
BarryThePenguin, line 1: We can delete this now that #1212 is merged

packages/ua-blocker/.gitignore
BarryThePenguin, line 1: Can probably delete this? 👀

packages/ua-blocker/package.json
BarryThePenguin: We can remove the `yarn prebuild` script from `build`, `typecheck`, and `test`. That way `getrobotstxt` is only run in `ci-ua-blocker-sync.yml` with node `22.x`

packages/ua-blocker/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "description": "User agent-based blocker for Hono",
+  "type": "module",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "prebuild": "node --experimental-strip-types ./script/prebuild.ts",
+    "build": "yarn prebuild && tsup ./src/index.ts ./src/ai-bots.ts",
```
BarryThePenguin:
```suggestion
    "build": "tsup ./src/index.ts ./src/ai-bots.ts",
```

packages/ua-blocker/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "prepack": "yarn build",
+    "publint": "attw --pack --profile node16 && publint",
+    "typecheck": "yarn prebuild && tsc -b tsconfig.json",
```
BarryThePenguin:
```suggestion
    "typecheck": "tsc -b tsconfig.json",
```

packages/ua-blocker/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "test": "yarn prebuild && vitest",
```
BarryThePenguin:
```suggestion
    "test": "vitest",
```

BarryThePenguin (approved):

finxol:
Suggested changes are applied, ci should now pass

yusukebe:
@finxol

I tried using this UA Blocker middleware, and it felt good. The rest is making CI/test pass.

finxol:
Sorry for the force push, needed to rebase to have CI pass with `.prettierignore`

.prettierignore
```diff
@@ -1,3 +1,4 @@
 .changeset
 .vscode
-.yarn
\ No newline at end of file
+.yarn
+**/generated.ts
```
finxol, line 4: `generated.ts` should be ignored by Prettier, but having it in `packages/ua-blocker/.prettierignore` doesn't get picked up because of the `prettier --check . '!packages' packages/ua-blocker` path specification.

finxol:
@yusukebe Sorry for the ci and test issues. Everything should be fixed now.

yusukebe (approved):
LGTM!

yusukebe:
@finxol 

Thanks! Let's go with this. I'll merge and release an initial version.

PR #1188, opened 2025-06-02
feature(session): add session middleware

BarryThePenguin:
I've been replacing an authentication library in a Next.js application, and thought it would be a good contribution to make. This PR is only focusing on the session middleware.

It started out as a library to bridge the gap between the previous authentication library and a new library. It runs on Cloudflare Workers, so the library gradually evolved into a standalone worker. This latest iteration is now middleware for a Hono app. It's been in progress for a while now.

I've done a lot of reading and research along the way. Some key resources that I've used include:

- [The Copenhagen Book](https://thecopenhagenbook.com/)
- [Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html)
- [`openid-client`](https://github.com/panva/openid-client)
- [`jose`](https://github.com/panva/jose)

I also looked at using other similar libraries and frameworks including..

- [OpenAuth](https://openauth.js.org/)
- [Better Auth](https://www.better-auth.com/)
- [`@hono/oidc-auth`](https://github.com/honojs/middleware/tree/main/packages/oidc-auth)

..all of which do a lot more than I need, like account and database management, or don't expose the functionality I'm after. Specifically, an OIDC-client service to manage sessions and access/id/refresh tokens, while acting as an auth middleware  proxying requests to upstream services.

What I'm essentially after is [OAuth2 Proxy](https://oauth2-proxy.github.io/oauth2-proxy/) that I can run in Cloudflare Workers 😅 

There are a few other's out there [with similar needs](https://github.com/better-auth/better-auth/issues/2202), so I thought I'd share the end result.


packages/session/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "type": "git",
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/session"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "dependencies": {
+    "@panva/hkdf": "^1.2.1",
```
yusukebe: Oops, I looked at [the library](https://github.com/panva/hkdf), and then it was archived:

CleanShot 2025-06-23 at 16 07 01@2x

> This project is now archived. The functionality provided here is no longer necessary, as the Web Cryptography API is now widely supported across all major and modern JavaScript runtimes and you may just use the following

Should we use `Web Cryptography API` instead?

packages/session/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+   * @default process.env.AUTH_SECRET
+   */
+  secret?: string | EncryptionKey
+
+  deleteCookie?: typeof cookie.deleteCookie
+  getCookie?: typeof cookie.getCookie
+  setCookie?: typeof cookie.setCookie
+}
+
+type SessionEnv<Data> = Env & {
```
yusukebe: To enable type inference when the handlers are not chained, how about exporting `SessionEnv`?

```diff
diff --git a/packages/session/src/index.ts b/packages/session/src/index.ts
index 9795e67..c959de4 100644
--- a/packages/session/src/index.ts
+++ b/packages/session/src/index.ts
@@ -42,7 +42,7 @@ export interface SessionOptions<Data> extends SessionEvents<Data> {
   setCookie?: typeof cookie.setCookie
 }

-type SessionEnv<Data> = Env & {
+export type SessionEnv<Data = SessionData> = Env & {
   Bindings: {
     AUTH_SECRET?: string
   }
```

Usage: 

```ts
import { useSession, type SessionEnv } from '@hono/session'
import { Hono } from 'hono'

const app = new Hono<SessionEnv>()

app.use(useSession())

app.get('/put', async (c) => {
  await c.var.session.update({})
  return c.redirect('/')
})
```

Or you can pass generics:

```ts
import { useSession, type SessionEnv } from '@hono/session'
import { Hono } from 'hono'

const app = new Hono<
  SessionEnv<{
    message: string
  }>
>()

app.use(useSession())

app.get('/put', async (c) => {
  await c.var.session.update({
    message: 'hello' // Typed
  })
  return c.redirect('/')
})
```

packages/session/package.json
BarryThePenguin: Yes, we can. That's a very recent change 👀

packages/session/src/index.ts
BarryThePenguin: Looks good 👍🏻 I got a bit stuck trying to make this work well

packages/session/examples/cloudflare-kv.ts
```diff
@@ -0,0 +1,51 @@
...
+          // Optionally configure session data to expire some time after the session cookie expires.
+          expirationTtl: 2_592_000, // 30 days in seconds
+        })
+      )
+    },
+  })),
+  useSession({ secret })
+)
+
+app.get('/session', async (c) => {
```
yusukebe, line 36: I updated the examples to the "not chained" style. It's common for users. If you pass `SessionEnv` to `Hono` as a generic, you can write it like this.

yusukebe (approved):
LGTM!

yusukebe:
Hi @BarryThePenguin 

Looks good to me! I like this middleware. I'll create some examples using it later. Thank you!

packages/session/examples/cloudflare-kv.ts
BarryThePenguin: Makes sense, thank you!

PR #1212, opened 2025-06-10
ci: what changed

BarryThePenguin:
Removed the individual ci workflows for a single workflow using [`yarn workspaces list`](https://yarnpkg.com/cli/workspaces/list)

I've kept `@hono/bun-transpiler` in it's own workflow to keep things simple


packages/ajv-validator/package.json
```diff
@@ -44,7 +44,7 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.17.4",
-    "ajv": ">=8.12.0",
+    "ajv": "^8.12.0",
```
BarryThePenguin, line 47: I used this change to trigger CI

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+          node-version: 20.x
+      - name: Get changed packages
+        id: set-packages
+        run: |
+          exclude=(
+            "hono-middleware"
+            "@hono/bun-transpiler"
+          );
+          changed=$(yarn workspaces list --json --since | jq -nc --arg exclude "$exclude" '[inputs.name | select(inside($exclude) | not) | sub("@hono/"; "")]')
+          echo "packages=${changed}" >> $GITHUB_OUTPUT
```
BarryThePenguin, line 28: Uses [`yarn workspaces list`](https://yarnpkg.com/cli/workspaces/list), along with `--since`, to list workspaces that have changed since `origin/main`

```json
[
  {
    "location": ".",
    "name": "hono-middleware"
  },
  {
    "location": "packages/ajv-validator",
    "name": "@hono/ajv-validator"
  }
]
```

[`jq`](https://jqlang.org/) then selects each package `name`..

```json
[
  "hono-middleware",
  "@hono/ajv-validator"
]
```

..excludes the root workspace, `hono-middleware`, and `@hono/bun-transpiler` as it has it's on workflow that uses `bun`..

```json
[
  "@hono/ajv-validator"
]
```

..and substitutes `@hono/` with an empty string

```json
[
  "ajv-validator"
]
```

This output is [stored in context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs) to be used in the following jobs

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+      - run: yarn workspaces focus hono-middleware @hono/${{ matrix.package }}
+      - run: yarn workspaces foreach --topological --recursive --from @hono/${{ matrix.package }} run build
+      - run: yarn workspace @hono/${{ matrix.package }} publint
+      - run: yarn workspace @hono/${{ matrix.package }} typecheck
+      - run: yarn eslint packages/${{ matrix.package }}
+      - run: yarn prettier --check . !packages packages/${{ matrix.package }}
```
BarryThePenguin, line 48: Uses the output from `what-changed` to run CI scripts on just the packages that have changes

The output from `what-changed` is used to build a dynamic [matrix strategy](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow), which is referenced by each step to run a script in a specific workspace, eg;

```sh
$ yarn workspace @hono/${{ matrix.package }} typecheck
``` 

..or, run a script using a specific workspace directory, eg;

```sh
$ yarn eslint packages/${{ matrix.package }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+        id: test
+        if: ${{ matrix.package != 'qwik-city' }}
+      - uses: codecov/codecov-action@v5
+        if: ${{ matrix.package != 'qwik-city' }}
+        with:
+          fail_ci_if_error: true
+          directory: ./coverage
+          flags: ${{ matrix.package }}
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
BarryThePenguin, line 74: Does basically the same as the previous job, but runs tests using the same dynamic matrix strategy. I made this a seperate job just because tests can be a bit slower than the other steps

.github/workflows/codestyle.yml
BarryThePenguin, line 1: I dropped `.github/workflows/codestyle.yml` in favour of..

```sh
$ yarn eslint packages/${{ matrix.package }}
$ yarn prettier --check . !packages packages/${{ matrix.package }}
```

..which does the same thing, except in individual workspaces rather than the entire monorepo

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Awesome!! This is clever, smart, and neat! Thank you.

Issue #740, opened 2024-09-13
Also publish to JSR?


imcotton:
From the **Hono** document:

> If you want to use Third-party Middleware with the TypeScript Type inferences, you need to use the npm: specifier. [^deno]

For package like `npm:@hono/valibot-validator`, this holding my back from using **Hono** from **JSR** by:

- `jsr:@hono/hono`
- `jsr:@valibot/valibot`

Any plans to have middleware packages also available on JSR too?

[^deno]: https://hono.dev/docs/getting-started/deno#npm-specifier

yusukebe:
Hi @imcotton 

Supporting JSR is great. But there are some reasons I can't do it now:

* Making them support JSR will be hard. Especially removing slow types.
* Have to create deploy flow. I'm unsure if we can create a good flow to publicize them with `Changesets.`
* Maintenance costs will be increased. We(almost I) must also check the JSR things by adding npm packages.

imcotton:
I think slow types is OK-ish for just publish with `--allow-slow-types` if it cost too much of effort.

Flow-wise, I think it'd be done by:
- having one corresponding `jsr.json` file
- add `deno publish` or `npx jsr publish` right after the `npm publish` on its CI

Maintenance indeed increased, but if I'm not missing anything else, above only be one time setup per package and then automatically carried afterwards.

Not sure the `jsr:@hono/*` is the intended place to publish these middleware, but right now it rather empty since only hosting the `jsr:@hono/hono` and nothing else.

I understand this is low priority and still up to decisions, in the meantime I could vendor the TypeScript source code by coping into my own project, but thought it worth to brought up to discussion, thanks for the participation.

ingokpp:
+1 For JSR support. I really like the idea of not using npm at all in my deno project.

dmint789:
+1 too. Would be nice to have the `@hono/zod-validator` package on there.

azohra:
Took me longer than I care to admit to figure out these were not on JSR and that I had to use NPM. I am a newer Deno user and I can absolutely appreciate it may not be worth the effort to cross-publish. 

My small ask would be if updating the docs to indicate that the 3rd party portions are only available on NPM. Happy to make a PR if this would be a desired outcome.

riderx:
My type is broken because it uses 2 different types apparently, and I think it's linked to this:
[...]

PR #1216, opened 2025-06-11
refactor: add file extensions

BarryThePenguin:
As part of #740, [relative imports should use the correct extensions](https://jsr.io/docs/publishing-packages#relative-imports)

This can be checked using `deno publish --dry-run`, which I'll include in a future PR


tsconfig.base.json
```diff
@@ -11,6 +11,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "forceConsistentCasingInFileNames": true,
+    "rewriteRelativeImportExtensions": true,
```
BarryThePenguin, line 14: Also need to enable [`rewriteRelativeImportExtensions`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#path-rewriting-for-relative-paths) so exisiting tools continue to work

yusukebe:
Hey @BarryThePenguin 

Thank you for the PR. Regarding extensions, I want to use `sloppy-imports` to avoid adding extensions. `honojs/hono` uses it: https://github.com/honojs/hono/blob/6837649cd104b1886e6cb6653bf26d0700c4c3e6/jsr.json#L12

BarryThePenguin:
Sure thing, that looks simpler 👍🏻 

Is there a reason why you want to avoid adding extensions?

yusukebe:
@BarryThePenguin 

Thanks!

> Is there a reason why you want to avoid adding extensions?

It is not common for other developers to add extensions. We have not had any significant problems with `honojs/hono` using `sloppy-imports`.

PR #1217, opened 2025-06-11
refactor(zod-openai): enable isolated declarations

BarryThePenguin:
This is the last package to enable `isolatedDeclarations`

There is one issue where two methods use `ReturnType<typeof generator.generateDocument>`, which is too dynamic for JSR

https://github.com/honojs/middleware/blob/a7ee664c98b3a3089ac2f8c0df8c2b6f5470c64a/packages/zod-openapi/src/index.ts#L558-L574

Two solutions I can think of
1. Add `openapi3-ts` as a dependency and import the types directly
2. Update `@asteasolutions/zod-to-openapi` so it re-exports the necessary types from `openapi3-ts`

### DIFF
--- a/.changeset/petite-masks-melt.md
+++ b/.changeset/petite-masks-melt.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Add explicit return types
--- a/packages/zod-openapi/package.json
+++ b/packages/zod-openapi/package.json
@@ -53,7 +53,8 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^7.3.0",
-    "@hono/zod-validator": "workspace:^"
+    "@hono/zod-validator": "workspace:^",
+    "openapi3-ts": "^4.5.0"
   },
   "engines": {
     "node": ">=16.0.0"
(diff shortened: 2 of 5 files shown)


yusukebe:
@BarryThePenguin 

> 1. Add `openapi3-ts` as a dependency and import the types directly
> 2. Update `@asteasolutions/zod-to-openapi` so it re-exports the necessary types from `openapi3-ts`

I like `2`. But is the new version of `@asteasolutions/zod-to-openapi` exporting the expected type(`OpenAPIObject`?)?

BarryThePenguin:
> But is the new version of `@asteasolutions/zod-to-openapi` exporting the expected type(`OpenAPIObject`?)?

Not that I can find 👀

yusukebe:
@BarryThePenguin 

I also can't find the type. So, importing it from `openapi3-ts` is okay.

yusukebe:
Hey @BarryThePenguin 

Shall we go with `1`?

> 1. Add `openapi3-ts` as a dependency and import the types directly

BarryThePenguin:
Done! Thanks for the reminder

yusukebe:
@BarryThePenguin Thanks!

PR #1186, opened 2025-06-02
feat: Stytch Authentication middlewares

max-stytch:
## Stytch Authentication middleware

This PR adds a suite of pluggable middlewares for use with the Stytch Consumer and Stytch B2B projects. We're doing a _lot_ of [fun stuff](https://developers.cloudflare.com/agents/model-context-protocol/authorization/#3-bring-your-own-oauth-provider) with Hono (mostly MCPs on Cloudflare Workers), and I find myself copying and pasting the same middleware between a bunch of different projects.

~The README was written in a Claude Code induced haze - I'm going to come back tomorrow and reread it with fresh eyes~. 

### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn changeset` at the top of this repo and push the changeset
- [x] Follow [the contribution guide](https://github.com/honojs/middleware?tab=readme-ov-file#how-to-contribute)

PR #1222, opened 2025-06-13
chore: Prettier should operate on all files

max-stytch:
Changes the `yarn format:fix` command to run on all files, not just `ts/tsx` files. This matches what the CI command does: https://github.com/honojs/middleware/blob/44b1c24b95281211bdc796adda1da2734b25f49b/.github/workflows/ci.yml#L48

This came up in #1186 - I would change the README.md file, run the formatter locally, and then still experience errors in CI because the README would get skipped

N/A

- [ ] Add tests
- [ ] Run tests


yusukebe:
@max-stytch Thanks!

Hey @BarryThePenguin, can you review this?

package.json
```diff
@@ -17,8 +17,8 @@
     "test": "vitest",
     "lint": "eslint 'packages/**/*.{ts,tsx}'",
     "lint:fix": "eslint --fix 'packages/**/*.{ts,tsx}'",
-    "format": "prettier --check 'packages/**/*.{ts,tsx}'",
-    "format:fix": "prettier --write 'packages/**/*.{ts,tsx}'"
+    "format": "prettier --check 'packages/**/*'",
+    "format:fix": "prettier --write 'packages/**/*'"
```
BarryThePenguin:
```suggestion
    "format": "prettier --check .",
    "format:fix": "prettier --write ."
```
This has the inverse problem to what you experienced in your previous PR — if something on `main` isn't formatted, this may end up with unrelated changes in your branch.

Though that's better than the command not fixing something that's failing in CI 👍🏻

.prettierignore
```diff
@@ -1,3 +1,7 @@
 .changeset
 .vscode
-.yarn
\ No newline at end of file
+.yarn
+
+# Casbin
+*.conf
+*.csv
```
BarryThePenguin: Are these needed? I'm unsure if they need to be more specific, or they might be unrelated to the project?

max-stytch: They're from the casbin package - https://github.com/honojs/middleware/tree/main/packages/casbin/examples

but since Prettier can't speak `.csv` it makes sense to ignore _all_ csv files regardless of package

yusukebe (approved):
LGTM!

yusukebe:
@max-stytch @BarryThePenguin 

Thank you!

PR #1220, opened 2025-06-13
chore: update vitest to v3.2

BarryThePenguin:
[Vitest 3.2](https://vitest.dev/blog/vitest-3-2) deprecated `workspace`, and are replacing it with [`projects`](https://vitest.dev/guide/projects.html)

PR #1230, opened 2025-06-15
refactor: compsite build

BarryThePenguin:
Moving a few things around in an attempt to fix #1220

- The only `tsconfig.json` files that need [`composite`](https://www.typescriptlang.org/tsconfig/#composite) enabled are the `tsconfig.build.json` files, which is used by type checking between packages.

- We can enable [`emitDeclarationOnly`](https://www.typescriptlang.org/tsconfig/#emitDeclarationOnly) everywhere as we're using `tsup` to publish packages.


tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
-  "extends": "./tsconfig.base.json",
```
BarryThePenguin, line 2: Only `tsconfig.build.json` and `tsconfig.spec.json` files need to extend the base configuration. The `tsconfig.json` files are only used as project references

.github/workflows/ci.yml
```diff
@@ -41,8 +41,7 @@ jobs:
         with:
           node-version: 20.x
       - run: yarn workspaces focus hono-middleware @hono/${{ matrix.package }}
-      - run: yarn workspaces foreach --topological --recursive --from @hono/${{ matrix.package }} run build
-      - run: yarn workspace @hono/${{ matrix.package }} publint
+      - run: yarn workspaces foreach --topological --recursive --from @hono/${{ matrix.package }} run publint
```
BarryThePenguin, line 44: Was running into an issue where `tsup` was referencing `dist/*.d.ts` files from the previous `build` step when running `publint` 😩

packages/ua-blocker/examples/demo.ts
```diff
@@ -1,6 +1,6 @@
-import { uaBlocker } from '@hono/ua-blocker'
-import { nonRespectingAiBots, useAiRobotsTxt } from '@hono/ua-blocker/ai-bots'
```
BarryThePenguin, line 2: Referencing `@hono/ua-blocker` from within the `src` directory is causing test failures in both #1178 and #1220. I've moved `demo.ts` out of `src` into a new `examples` directory. There might be other ways to do this, but it should unblock others for now..

yusukebe: I see!

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Thank you for the hard work! Looks good.

PR #1242, opened 2025-06-19
chore(cr): publish pr to pkg.pr.new

### LINKED ISSUE #1227: Introduce `pkg.pr.new`
`pkg.pr.new` seems to be good: https://github.com/stackblitz-labs/pkg.pr.new

BarryThePenguin:
Add workflow to publish pull requests to https://pkg.pr.new/

Just need to install the [Github App](https://github.com/apps/pkg-pr-new) first

Fixes #1227


.github/workflows/cr.yml
```diff
@@ -0,0 +1,36 @@
+name: cr
+on:
+  push:
+    branches: [main]
+    tags: ['!**'] # Avoid publishing on tags
+  pull_request:
+    types: [opened, synchronize, labeled] # Run on PR creation, updates, and when labels are added
```
BarryThePenguin, line 7: Run the workflow when pull requests are opened/updated, or when a label is added

I used https://github.com/vitest-dev/vitest/blob/main/.github/workflows/cr.yml for reference

.github/workflows/cr.yml
```diff
@@ -0,0 +1,36 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }} # Concurrency group for each PR
+  cancel-in-progress: true # Cancel in progress builds for the same PR
```
BarryThePenguin, line 11: Cancels in-progress worflows when the current PR is updated

.github/workflows/cr.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  publish:
+    if: github.repository == 'honojs/middleware' && (github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'cr-tracked'))
```
BarryThePenguin, line 15: Only run the publish job if the PR has the `cr-tracked` label

.github/workflows/cr.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Build
+        run: yarn workspaces foreach --topological --parallel --since --no-private run build
```
BarryThePenguin, line 31: Only build packages changed in the PR

.github/workflows/cr.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Publish to StackBlitz
+        run: |
+          changed=$(yarn workspaces list --json --since --no-private | jq -sr '[.[].location] | @sh')
+          yarn pkg-pr-new publish --compact --no-template ${changed}
```
BarryThePenguin, line 36: Only publish packages that have changed

.github/workflows/cr.yml
yusukebe: Ah, using the `cr-tracked` label! I see.

yusukebe:
I installed the GitHub App and added the `cr-tracked` label. I think this will work. I'll merge.

yusukebe:
Oh, failed 👀 
https://github.com/honojs/middleware/actions/runs/15792065173/job/44519205750#step:6:7

BarryThePenguin:
Ah.. that's probably because there were no changed packages to publish 🤔 

Should CI just publish all packages on `main`? I guess that is the point of continuous releases isn't it..

PR #1244, opened 2025-06-20
lint: no-confusing-void-expression

BarryThePenguin:
Enables [`@typescript-eslint/no-confusing-void-expression`](https://typescript-eslint.io/rules/no-confusing-void-expression/)


packages/auth-js/src/client.ts
```diff
@@ -143,8 +143,12 @@ export function useOnline(): boolean {
     const abortController = new AbortController()
     const { signal } = abortController
 
-    const setOnline = () => setIsOnline(true)
-    const setOffline = () => setIsOnline(false)
```
BarryThePenguin, line 147: This is also the option to enable `ignoreArrowShorthand` to allow this style

yusukebe:
@BarryThePenguin 

This is a good rule. Thanks!

PR #1250, opened 2025-06-22
fix(ua-blocker-sync): format generated files

BarryThePenguin:
Looks like line ending of the generated `robots.json` is causing `git status --porcelain` to report changes when there are none.

This runs prettier to normalise line endings, before the change detection step runs

fixes #1247


yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Nice catch. Thanks!

finxol:
thanks for fixing it!

PR #1272, opened 2025-07-01
feat(session): enable accessing sid via `c.var.session.id`

yusukebe:
With this PR, you can get the current session id via `c.var.session.id`. I am also thinking the name is `c.var.session.sid` but simply `c.var.session.id`.

The original idea is by @craigsdennis

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/packages/session/src/session.ts
+++ b/packages/session/src/session.ts
@@ -27,6 +27,11 @@ export interface Session<Data> {
   readonly data: Data | null
 
   /**
+   * Current session ID.
+   */
+  readonly id: string | null
+
+  /**
    * Delete the current session, removing the session cookie and data from storage.
    */
   delete: () => void
@@ -121,6 +126,17 @@ export function getSession<Data extends SessionData>({
       }
 
       return session.data ?? null
+    },
+    get id() {
+      if (!session) {
+        throw new Error('Session not initialised. Call get() or update() first.')
+      }
+
+      if (session.action === 'destroy') {
+        throw new Error('Session has been destroyed.')
+      }
+
+      return session.payload?.sid ?? null
     },
     delete() {
       if (session) {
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @BarryThePenguin ! Can you review this?

BarryThePenguin (commented):
It would be helpful to understand the need for accessing the session id.

I originally wanted to avoid exposing `session.sid`. Simply based on the [Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html)

> The disclosure, capture, prediction, brute force, or fixation of the session ID will lead to session hijacking (or sidejacking) attacks, where an attacker is able to fully impersonate a victim user in the web application.

I interpreted that to mean the value of `sid`. On a second reading though, that could also just refer to the `sid=value` cookie pair.

This is an area I don't have a lot of experience in, so I'm open to any input and ideas!

packages/session/src/session.ts
```diff
@@ -122,6 +127,17 @@ export function getSession<Data extends SessionData>({
...
+    get id() {
+      if (!session) {
+        throw new Error('Session not initialised. Call get() or update() first.')
+      }
+
+      if (session.action === 'destroy') {
+        throw new Error('Session has been destroyed.')
+      }
+
+      return session.payload?.sid ?? null
```
BarryThePenguin, line 139: I had intended `session.payload` to remain internal to the middleware. While `session.data` was for applications to store additional information like a user id, or access/id/refresh tokens. `sid` is currently only used for session management with the storage middleware.

BarryThePenguin (approved):
After thinking about this for a bit, it's probably fine to expose this. The cookie value is what needs to remain secure

yusukebe:
@BarryThePenguin 

Thank you for reviewing. I also think it's okay to expose the `sid` if the cookie values are treated as secure. Merging this now.

PR #1273, opened 2025-07-02
refactor(release): single release command

BarryThePenguin:


.github/workflows/release.yml
```diff
@@ -42,8 +42,8 @@ jobs:
         id: changesets
         uses: changesets/action@v1
         with:
-          publish: yarn run publish && deno install --no-lock && deno publish
-          version: yarn run version:npm && yarn run version:jsr
+          publish: yarn run publish
+          version: yarn run version
```
BarryThePenguin, line 46: It looks like changesets expects these to be a single command, rather than multiple scripts. So I've combined these into individual `package.json` scripts

BarryThePenguin (commented):
I'm unsure if it's possible to verify the changes to the changeset action without actually running the action, so it might just be a matter of merge and let's see 🙈

.github/workflows/release.yml
yusukebe: I see! Thanks.

yusukebe (approved):

yusukebe:
Oops.

Unfortunately, it had a diff.

https://github.com/honojs/middleware/actions/runs/16014017988/job/45177157036#step:8:956

BarryThePenguin:
Ah.. that's specifically for `.yarnrc.yml`. I think I've seen that before, it's probably changing the formatting of the file

yusukebe:
> the formatting of the file

Ah, it's a bit of trouble.

BarryThePenguin:
I see now.. `yarn config set npmAuthToken ${NPM_TOKEN}` is writing to `.yarnrc.yml` which is causing the diff

PR #1274, opened 2025-07-02
refactor(release): use `YARN_NPM_AUTH_TOKEN`

BarryThePenguin:
This should prevent the release workflow from modifying `.yarnrc.yml`

From https://yarnpkg.com/configuration/yarnrc

> Finally, note that most settings can also be defined through environment variables (at least for the simpler ones; arrays and objects aren't supported yet). To do this, just prefix the names and write them in snake case: `YARN_CACHE_FOLDER` will set the cache folder (such values will overwrite any that might have been defined in the RC files - use them sparingly).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,11 +33,6 @@ jobs:
       - name: Build
         run: yarn build
 
-      - name: Set NPM Auth Token
-        run: yarn config set npmAuthToken ${NPM_TOKEN}
-        env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@v1
@@ -46,4 +41,4 @@ jobs:
           version: yarn run version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}


.github/workflows/release.yml
```diff
@@ -46,4 +41,4 @@ jobs:
           version: yarn run version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
yusukebe, line 44: So, should I set `YARN_NPM_AUTH_TOKEN` as a secret variable in this GitHub repo?

BarryThePenguin: This assigns the existing secret for `NPM_TOKEN` to the `YARN_NPM_AUTH_TOKEN` environment variable. So there's no need to change any existing secrets

BarryThePenguin: To provide a bit more context..

The [changesets action](https://github.com/changesets/action?tab=readme-ov-file#with-publishing) uses `env.NPM_TOKEN` to configure `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}` in `.npmrc`. Because the we're using yarn, the equivalent is `yarn config set npmAuthToken ${NPM_TOKEN}`.

We can avoid modifying `.yarnrc.yml` by setting `env.YARN_NPM_AUTH_TOKEN`, removing the need to use `yarn config set`.

yusukebe: Thank you for the explanation. Sorry for my misunderstanding. I can understand well!

yusukebe (approved):

yusukebe:
@BarryThePenguin Thank you. I'll merge this now.

yusukebe:
I've resolved some issues. This may be the last one. Hmm.

https://github.com/honojs/middleware/actions/runs/16039539202/job/45258963847#step:7:976

BarryThePenguin:
It looks like deno doesn't understand peer dependencies, so it fails when trying to publish a package with

```json
"peerDependencies": {
  "hono": "*"
}
```

I'll keep digging 👀

BarryThePenguin:
Hopefully #1276 solves that issue, it's difficult to verify as I can't reproduce the issue locally

yusukebe:
> It looks like deno doesn't understand peer dependencies

Ah, I see!

PR #1277, opened 2025-07-03
chore: add jsr specifier for hono

BarryThePenguin:
I've only just realised that `hono` is published as [`@hono/hono` on JSR](https://jsr.io/@hono/hono). So it may be worth adding an import specifier to point `hono` at `jsr:@hono/hono@^4.8.3`?


deno.jsonc
```diff
@@ -4,9 +4,9 @@
     "packages/ajv-validator",
     "packages/arktype-validator",
     // "packages/auth-js",
-    "packages/bun-compress",
+    // "packages/bun-compress",
```
BarryThePenguin, line 7: > Unknown export './utils/compress' for '@hono/hono@4.8.3'

Looks like hono is missing an export for `./utils/compress`. There is one for `./utils/*`, but I think Deno/JSR don't understand the wildcard

deno.jsonc
```diff
@@ -4,9 +4,9 @@
...
     // "packages/bun-transpiler",
-    "packages/casbin",
+    // "packages/casbin",
```
BarryThePenguin, line 9: > Unknown export './utils/basic-auth' for '@hono/hono@4.8.3'.

Likewise for `./utils/basic-auth`

deno.jsonc
```diff
@@ -19,7 +19,7 @@
     "packages/graphql-server",
     "packages/hello",
     // "packages/medley-router",
-    "packages/node-ws",
+    // "packages/node-ws",
```
BarryThePenguin, line 22: This one was a bit more complicated..

> TS2741 [ERROR]: Property '#private' is missing in type '{ binaryType: "arraybuffer"; close(code: number | undefined, reason: string | undefined): void; protocol: string; raw: WebSocket; readonly readyState: 0 | 2 | 1 | 3; send(source: string | ... 1 more ... | Uint8Array<...>, opts: SendOptions | undefined): void; url: URL; }' but required in type 'WSContext<WebSocket>'.
        const ctx: WSContext<WebSocket> = {
              ~~~
    at file:///Users/jonno/src/honojs/middleware/packages/node-ws/src/index.ts:117:15
>
>    '#private' is declared here.
      #private!: unknown;
      ~~~~~~~~
        at https://jsr.io/@hono/hono/4.8.3/src/helper/websocket/index.ts:65:35
>
> error: Type checking failed.

BarryThePenguin:
I'm wondering if it's worth opening an issue at https://github.com/denoland/deno before we try anything else

yusukebe:
@BarryThePenguin

I shared this problem with my friend working at Deno. He will forward it to another person familiar with it. So, we can wait.

yusukebe:
Ooooh. https://github.com/honojs/middleware/pull/1279 will fix it.

yusukebe:
@BarryThePenguin 

https://github.com/honojs/middleware/pull/1279 did not fix the problem. But I talked with my friend at Deno.

The problem may occur since each package was looking up the npm `hono` on the root. This PR can fix it by updating each package's lookup JSR `hono`. We can try it. I'll merge this PR now.

deno.jsonc
yusukebe: This package is for Node.js. So we may not need to publish it to JSR.

deno.jsonc
yusukebe: Ah, I see. It's a little bit of a hassle to export all utils from `hono`. But it's worth considering.

yusukebe:
@BarryThePenguin

It works!

https://github.com/honojs/middleware/actions/runs/16064045104/job/45335180349

CleanShot 2025-07-04 at 10 52 36@2x

Thank you for your work! I think it will be the biggest project that publishing packages on a monorepo to JSR.

BarryThePenguin:
Amazing, that's great news! Still a few packages remain, but I can chip away at them over time 😁

PR #1278, opened 2025-07-03
feat: add essential plugins package for Hono SSG

3w36zj6:
close https://github.com/honojs/hono/issues/2125

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/packages/ssg-plugins-essential/README.md
+++ b/packages/ssg-plugins-essential/README.md
@@ -1,0 +1,26 @@
+# Essential plugins for Hono SSG
+
+[![codecov](https://codecov.io/github/honojs/middleware/graph/badge.svg?flag=ssg-plugins-essential)](https://codecov.io/github/honojs/middleware)
+
+Essential plugins for [Hono SSG](https://hono.dev/docs/helpers/ssg).
+This plugin collection provides standard and lightweight plugins essential for static site generation.
+
+- Sitemap plugin
+- robots.txt plugin
+- RSS/Atom plugin
+
+## Usage
+
+### Sitemap plugin
+
+### robots.txt plugin
+
+### RSS/Atom plugin
+
+## Author
+
+3w36zj6 <https://github.com/3w36zj6>
+
+## License
+
+MIT
--- a/packages/ssg-plugins-essential/package.json
+++ b/packages/ssg-plugins-essential/package.json
@@ -1,0 +1,51 @@
+{
+  "name": "@hono/ssg-plugins-essential",
+  "version": "1.0.0",
+  "description": "Essential plugins for Hono SSG",
+  "type": "module",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup ./src/index.ts",
+    "prepack": "yarn build",
+    "publint": "attw --pack && publint",
+    "typecheck": "tsc -b tsconfig.json",
+    "test": "vitest"
+  },
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/ssg-plugins-essential"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": "*"
+  },
+  "devDependencies": {
+    "@arethetypeswrong/cli": "^0.17.4",
+    "publint": "^0.3.9",
+    "tsup": "^8.4.0",
+    "typescript": "^5.8.2",
+    "vitest": "^3.2.4"
+  }
+}
--- a/packages/ssg-plugins-essential/tsconfig.build.json
+++ b/packages/ssg-plugins-essential/tsconfig.build.json
@@ -1,0 +1,5 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {},
+  "references": []
+}
--- a/packages/ssg-plugins-essential/tsconfig.json
+++ b/packages/ssg-plugins-essential/tsconfig.json
@@ -1,0 +1,12 @@
+{
+  "files": [],
+  "include": [],
+  "references": [
+    {
+      "path": "./tsconfig.build.json"
+    },
+    {
+      "path": "./tsconfig.spec.json"
+    }
+  ]
+}
--- a/packages/ssg-plugins-essential/tsconfig.spec.json
+++ b/packages/ssg-plugins-essential/tsconfig.spec.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "../../dist/packages/ssg-plugins-essential",
+    "types": ["vitest/globals"]
+  },
+  "references": []
+}
--- a/packages/ssg-plugins-essential/vitest.config.ts
+++ b/packages/ssg-plugins-essential/vitest.config.ts
@@ -1,0 +1,7 @@
+import { defineProject } from 'vitest/config'
+
+export default defineProject({
+  test: {
+    globals: true,
+  },
+})


3w36zj6:
Hi @yusukebe,

I'm currently implementing an RSS Plugin for Hono SSG.
I want to collect `text/html` pages detected in `afterResponseHook` and generate an RSS XML file.
However, the `url` property of the `Response` object is always an empty string.

Below is my minimal implementation (except for embedding the correct URL):

```ts
import type { AfterGenerateHook, AfterResponseHook, SSGPlugin } from 'hono/ssg'
import { DEFAULT_OUTPUT_DIR } from 'hono/ssg'
import path from 'node:path'

const extractTitleFromHtml = (html: string): string => {
  const match = html.match(/<title>(.*?)<\/title>/i)
  return match ? match[1] : ''
}

const extractDescriptionFromHtml = (html: string): string | undefined => {
  const match = html.match(/<meta\s+name=["']description["']\s+content=["'](.*?)["']/i)
  return match ? match[1] : undefined
}

type FeedType = 'rss2'

type FeedItem = {
  title: string
  link: string
  description?: string
}
/**
 * RSS plugin options.
 *
 * @property baseURL - The base URL of the site, used to generate full URLs in the RSS feed.
 * @property feedTitle - The title of the RSS feed.
 * @property feedDescription - The description of the RSS feed.
 * @property feedType - The type of RSS feed to generate. Default is RSS 2.0.
 */
export type RssPluginOptions = {
  baseURL: string
  feedTitle: string
  feedDescription: string
  feedType?: FeedType
}

/**
 * RSS plugin for Hono SSG.
 *
 * Generates an RSS feed file in the output directory.
 *
 * @param options - Options for the RSS plugin.
 * @returns A SSGPlugin to create the feed.
 */

export const rssPlugin = (options: RssPluginOptions): SSGPlugin => {
  const feedItems: FeedItem[] = []

  const afterResponseHook: AfterResponseHook = async (response) => {
    const contentType = response.headers.get('Content-Type') ?? ''
    if (contentType.includes('text/html')) {
      const html = await response.text()
      const title = extractTitleFromHtml(html)
      const description = extractDescriptionFromHtml(html)
      const url = response.url // 👈 `url` property is always an empty string.
      feedItems.push({ title, link: url, description })
      return new Response(html, response)
    }
    return response
  }

  const afterGenerateHook: AfterGenerateHook = async (_result, fsModule, ssgOptions) => {
    const outputDir = ssgOptions?.dir ?? DEFAULT_OUTPUT_DIR
    const fileName = 'rss.xml'
    const filePath = path.join(outputDir, fileName)
    const normalizedBaseURL = options.baseURL.endsWith('/')
      ? options.baseURL
      : `${options.baseURL}/`
    const feedLink = `${normalizedBaseURL}${fileName}`

    const rssFeed = `<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0">
<channel>
<title>${options.feedTitle}</title>
<link>${feedLink}</link>
<description>${options.feedDescription}</description>
${feedItems
  .map(
    (item) => `<item>
<title>${item.title}</title>
<link>${item.link}</link>
${item.description ? `<description>${item.description}</description>` : ''}
</item>`
  )
  .join('\n')}
</channel>
</rss>
`
    await fsModule.writeFile(filePath, rssFeed)
  }

  return {
    afterResponseHook,
    afterGenerateHook,
  }
}

```

To move forward, what kind of extension or improvement do you think would be ideal for `hono`, `hono/ssg`, or the RSS plugin regarding this issue?

yusukebe:
Hi @3w36zj6 

In conclusion, the `url` property will always be `undefined`.

The `response` is returned from `app.request()` and is generated by `new Response()`. A newly generated Response object does not have a value for the `url` property.

The Response object returned as the result of the actual HTTP request made by `fetch` does have a `url` property, but in this case, it will always be `undefined`.

So, we have to find another way.

3w36zj6:
Thank you for your advice.

After reconsidering, I realized that since the order of `beforeRequestHook` and `afterResponseHook` calls for each endpoint is consistent within the plugin, I can simply map the request URL to the response body. 

Could you please review this approach?
Let me know if you have any questions or need any changes.

packages/ssg-plugins-essential/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@hono/ssg-plugins-essential",
+  "version": "1.0.0",
```
yusukebe: The version number will be increased by Changeset automatically. So, please specify it as `0.0.0`.

.changeset/quiet-stars-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/ssg-plugins-essential': major
```
yusukebe: I think `major` is okay, but in some modules, we specify it as `minor` for the first release. If you set this `major`, it will be released `1.0.0`. I think we don't need to release as `v1` first.

yusukebe:
Hey @BarryThePenguin! Can you also review this PR?

yusukebe:
@3w36zj6 Please don't mind that CI is falling. https://github.com/honojs/middleware/actions/runs/16687457843/job/47239503657?pr=1278

packages/ssg-plugins-essential/README.md
```diff
@@ -0,0 +1,137 @@
...
+      content: string | Promise<string>,
+      head: { title: string; description?: string }
+    ): Response | Promise<Response>
+  }
+}
+
+const app = new Hono()
+
+app.use('/*', async (c, next) => {
+  c.setRenderer((content, head) => {
```
yusukebe: This is like nitpicking, but do you know you can use `title` and `head` features as in React 19?

```tsx
import { Hono } from 'hono'

const app = new Hono()

app.use('*', async (c, next) => {
  c.setRenderer((content) => {
    return c.html(
      <html>
        <head></head>
        <body>{content}</body>
      </html>
    )
  })
  await next()
})

app.get('/about', (c) => {
  return c.render(
    <>
      <title>About Page</title>
      <meta name="description" content="This is the about page." />
      about page content
    </>
  )
})

export default app
```

packages/ssg-plugins-essential/README.md
```diff
@@ -0,0 +1,137 @@
...
+        { userAgent: 'Googlebot', allow: ['/'] },
+      ],
+      sitemapUrl: 'https://example.com/sitemap.xml',
+      extraLines: ['# This is a comment.'],
+    }),
+  ],
+})
+```
+
+### RSS/Atom plugin
```
yusukebe, line 118: If you have the following application:

```tsx
import { Hono } from 'hono'

const app = new Hono()

app.get('/foo', (c) => {
  return c.html(
    <html>
      <body>
        <h1>Foo</h1>
      </body>
    </html>
  )
})

export default app
```

The RSS that is generated by this RSS/Atom plugin is below:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0">
<channel>
<title>My Blog</title>
<link>https://example.com/rss.xml</link>
<description>Latest updates from my blog.</description>
<item>
<title></title>
<link>https://example.com/foo</link>
<description></description>
</item>
</channel>
</rss>
```

The `link` content is `https://example.com/foo`, not `https://example.com/foo.html`. Is this expected behavior?

I aslo tried to use Sitemap plugin and the result is below:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
<url><loc>https://example.com/foo.html</loc></url>
</urlset>
```

The `<url><loc>` is `https://example.com/foo.html` which is direct from the above.

3w36zj6: In Hono SSG, when you use `/foo`, it generates `/foo.html`, and when you use `/foo/`, it generates `/foo/index.html`[^1].

From my understanding, the URLs in RSS and Sitemap should be canonicalized. Other SSGs like Astro, as well as many well-known sites, also serve canonical URLs.

So, in conclusion, I believe the current format of the RSS plugin is correct, and the Sitemap plugin should not include `.html` in its URLs.

[^1]: https://github.com/honojs/hono/blob/v4.8.0/src/helper/ssg/ssg.ts

yusukebe: I want to confirm the correct behaviors.

### 1. `/foo`

If you have the following app:

```tsx
// src/index.tsx
import { Hono } from 'hono'

const app = new Hono()

app.get('/foo', (c) => {
  return c.html(<h1>Foo</h1>)
})

export default app
```

And the code for building:

```ts
// src/build.ts
import app from './'
import { toSSG } from 'hono/ssg'
import fs from 'fs/promises'
import { rssPlugin } from '@hono/ssg-plugins-essential/rss'
import { sitemapPlugin } from '@hono/ssg-plugins-essential/sitemap'

toSSG(app, fs, {
  plugins: [
    sitemapPlugin({
      baseUrl: 'https://example.com'
    }),
    rssPlugin({
      baseUrl: 'https://example.com',
      feedDescription: 'foo',
      feedTitle: 'foo'
    })
  ]
})
```

It generates `static/foo.html`. 

The results are:

RSS:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0">
<channel>
<title>foo</title>
<link>https://example.com/rss.xml</link>
<description>foo</description>
<item>
<title></title>
<link>https://example.com/foo</link>
<description></description>
</item>
</channel>
</rss>
```

Sitemap:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
<url><loc>https://example.com/foo.html</loc></url>
</urlset>
```

This means:

* Path in app: `/foo`
* RSS: `https://example.com/foo`
* Sitemap: `https://example.com/foo.html`

### 2. `/foo/`

And if you have an app with the following.

```tsx
app.get('/foo/', (c) => {
  return c.html(<h1>Foo</h1>)
})
```

It generates `static/foo/index.html`. The results are:

* Path in app: `/foo/`
* RSS: `https://example.com/foo/`
* Sitemap: `https://example.com/foo/index.html`

Are these correct behaviors?

I think belows are correct behaviors:

`/foo`:

* RSS: `https://example.com/foo.html`
* Sitemap: `https://example.com/foo.html`

`/foo/`:

* RSS: `https://example.com/foo/`
* Sitemap: `https://example.com/foo/`

Or, if the platform you deploy is removing the trailing slash (I noticed that you may add the option for the user to specify it):

* RSS: `https://example.com/foo`
* Sitemap: `https://example.com/foo`

3w36zj6: Thank you for your detailed feedback and examples.

Currently, the RSS plugin canonicalizes URLs by default, so `/foo` in your app will be output as `https://example.com/foo` in the RSS feed, while the Sitemap plugin outputs the actual file path, such as `https://example.com/foo.html`.

The current issues to address are as follows:

- The URL representation is inconsistent between the sitemap plugin and the RSS plugin.
- While `/foo/index.html` can always be safely canonicalized to `/foo/`, canonicalizing `/foo.html` to `/foo` depends on the implementation and configuration of the hosting service.

To resolve these, I have extracted path canonicalization into a shared logic and added an option (`canonicalize`) to control the canonicalization of URLs without trailing slashes. After considering the default setting, I referred to [slorber/trailing-slash-guide](https://github.com/slorber/trailing-slash-guide), which shows that many major hosting services allow access to `/file` for `/file.html`, so I set the default to `true`.

How does this revision look to you?

packages/ssg-plugins-essential/README.md
3w36zj6: This is really nice! I didn't know about this, so it's very educational for me.

By the way, is this feature documented anywhere? I couldn't find it myself.

yusukebe: > By the way, is this feature documented anywhere? I couldn't find it myself

Ah, not documented. I think it's a good idea to put that into this page: https://hono.dev/docs/guides/jsx

packages/ssg-plugins-essential/README.md
```diff
@@ -0,0 +1,156 @@
...
+Simply add them to the `plugins` array when calling `toSSG`.
+Each plugin will automatically generate the corresponding file (e.g., `sitemap.xml`, `robots.txt`, or `rss.xml`) during the static site generation process.
+
+```tsx
+// index.tsx
+import { Hono } from 'hono'
+import { toSSG } from 'hono/ssg'
+import fs from 'fs/promises'
+
+declare module 'hono' {
```
yusukebe: I think this definition is not needed.

packages/ssg-plugins-essential/src/plugins/robots-txt.ts
```diff
@@ -0,0 +1,74 @@
...
+    afterGenerateHook: async (_result, fsModule, ssgOptions) => {
+      const outputDir = ssgOptions?.dir ?? DEFAULT_OUTPUT_DIR
+      const filePath = path.join(outputDir, 'robots.txt')
+
+      const lines: string[] = []
+
+      if (options.rules && options.rules.length > 0) {
+        for (const [i, rule] of options.rules.entries()) {
+          lines.push(`User-agent: ${rule.userAgent}`)
+          if (rule.allow) rule.allow.forEach((p) => lines.push(`Allow: ${p}`))
```
yusukebe: Although we should specify the style with the linter, currently, we don't use a one-line if block style. Therefore, the style shown in this patch is acceptable.

```diff
diff --git a/packages/ssg-plugins-essential/src/plugins/robots-txt.ts b/packages/ssg-plugins-essential/src/plugins/robots-txt.ts
index 47b7c25..4a69af9 100644
--- a/packages/ssg-plugins-essential/src/plugins/robots-txt.ts
+++ b/packages/ssg-plugins-essential/src/plugins/robots-txt.ts
@@ -42,9 +42,15 @@ export const robotsTxtPlugin = (options: RobotsTxtPluginOptions): SSGPlugin => {
       if (options.rules && options.rules.length > 0) {
         for (const [i, rule] of options.rules.entries()) {
           lines.push(`User-agent: ${rule.userAgent}`)
-          if (rule.allow) rule.allow.forEach((p) => lines.push(`Allow: ${p}`))
-          if (rule.disallow) rule.disallow.forEach((p) => lines.push(`Disallow: ${p}`))
-          if (rule.extraLines) lines.push(...rule.extraLines)
+          if (rule.allow) {
+            rule.allow.forEach((p) => lines.push(`Allow: ${p}`))
+          }
+          if (rule.disallow) {
+            rule.disallow.forEach((p) => lines.push(`Disallow: ${p}`))
+          }
+          if (rule.extraLines) {
+            lines.push(...rule.extraLines)
+          }
           if (i !== options.rules.length - 1) {
             lines.push('')
           }
```

packages/ssg-plugins-essential/src/plugins/rss.ts
```diff
@@ -0,0 +1,124 @@
...
+          description: meta?.description ?? '',
+          link: url,
+        }
+      })
+
+    const rssFeed = `<?xml version="1.0" encoding="UTF-8"?>
+<rss version="2.0">
+<channel>
+<title>${options.feedTitle}</title>
+<link>${feedLink}</link>
```
yusukebe, line 103: I think it's better to escape the variable strings. For example, if the string is `My Blog & <News>`, it will break the HTML tags, and it's dangerous.

So, you can create escaping logic by yourself, or you can use `html` helper: https://hono.dev/docs/helpers/html

packages/ssg-plugins-essential/src/utils/canonicalizeFilePath.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (routePath.endsWith('/index.html')) {
+    routePath = routePath.slice(0, -'index.html'.length) || '/'
+  } else if (routePath.endsWith('.html')) {
+    routePath = routePath.slice(0, -'.html'.length)
+  }
+  const canonicalBaseUrl = baseUrl.endsWith('/') ? baseUrl : `${baseUrl}/`
+  let url: string
+  if (cleanedFile.endsWith('index.html')) {
+    const dir = cleanedFile.slice(0, -'index.html'.length)
+    url = `${canonicalBaseUrl}${encodeURI(dir)}`
```
yusukebe: Simply, how about using `URL`?

```ts
url = new URL(dir, canonicalBaseUrl).toString()
```

packages/ssg-plugins-essential/src/plugins/rss.ts
```diff
@@ -0,0 +1,124 @@
...
+      ? options.baseUrl
+      : `${options.baseUrl}/`
+    const feedLink = `${normalizedBaseURL}${fileName}`
+
+    const items = result.files
+      .filter((file) => file.endsWith('.html'))
+      .map((file) => {
+        const { routePath, url } = canonicalizeFilePath(
+          file,
+          outputDir,
```
yusukebe, line 87: I think we should care about Windows not POISX. Imagine the `outputDir` on Windows, it can be passed like `dist\\foo\\bar` as a file path.

yusukebe: In such cases, it is safer and more reliable to use `node:path`.

packages/ssg-plugins-essential/README.md
3w36zj6: I’ve added documentation for this feature here: https://github.com/honojs/website/pull/710

yusukebe:
Hi @3w36zj6 

The last request. Can you update the `deno.jsonc` at the top?

yusukebe (approved):
LGTM!

yusukebe:
@3w36zj6 

Looks good to me. Thank you for your hard work! I'll merge and release the initial version.

PR #1279, opened 2025-07-04
chore: pin Deno to 2.3.7 in release workflow

kt3k:
The latest Deno (2.4.0) seems to have an issue with module resolution. Let's try to use 2.3.7 for releasing for now.

related:
- https://github.com/jsr-io/jsr-npm/issues/129
- https://github.com/jsr-io/jsr-npm/pull/130


yusukebe:
@kt3k 

Thanks!

Issue #1241, opened 2025-06-19
Use Standard Schema Validator instead of separate ones


yusukebe:
[Standard Schema](https://github.com/standard-schema/standard-schema) provides a unified way to work with different validator libraries. Many libraries already support it. We have Standard Schema validator in this `honojs/middleware` repo.

* [Standard Schema validator](https://github.com/honojs/middleware/tree/main/packages/standard-validator)

With Standard Schema Validator, we may no longer rely on individual validators:

* [ArkType validator](https://github.com/honojs/middleware/tree/main/packages/arktype-validator)
* [Effect Schema validator](https://github.com/honojs/middleware/tree/main/packages/effect-validator)
* [Valibot validator](https://github.com/honojs/middleware/tree/main/packages/valibot-validator)
* [Zod validator](https://github.com/honojs/middleware/tree/main/packages/zod-validator)

Issue #1235 specifically concerns ArkType validator, but by focusing on Standard Schema Validator, we can avoid handling such individual cases. We also won't need to add support for new validators when they appear.

I've been aware of the benefits of Standard Schema, but I was hesitant to commit to it fully. Now, I believe it's time to deprecate the separate validators and encourage the use of Standard Schema validator.

This proposal is based on [the comment](https://github.com/honojs/middleware/issues/1235#issuecomment-2979871560) by @MonsterDeveloper.

Please share any concerns you have about using Standard Schema validator. Otherwise, let's move forward!

cc: Authors of each validator

@MonsterDeveloper 
@gunta
@ZerNico
@muningis

MonsterDeveloper:
Thank you! Exactly what I was thinking about.

As for `arktype-validator`, the only custom logic piece apart from the validation itself is stripping out sensitive fields from the error return value (i.e. `cookies` header), implemented as per #1135 

Preliminary digging shows, that custom logic can be easily implemented by checking the `schema['~standard'].vendor === "arktype"`:

```ts
console.log(schema['~standard']);

// Output
{
  vendor: "arktype",
  version: 1,
  validate: [Function: validate],
}
```

I can add a PR for the standard schema validator for seamless migration.

Issue #1135, opened 2025-04-24
validating headers with arktype-validator can lead to http-only cookies being leaked to client


NVolcz:
### Which middleware has the bug?

@hono/arktype-validator

### What version of the middleware?

2.0.0

### What version of Hono are you using?

4.7.7

### What runtime/platform is your app running on? (with version if possible)

Bun

### What steps can reproduce the bug?

Applications somewhat commonly utilizes HTTP-only cookies in order to mitigate the security impact of XSS. This protection can be bypassed if the application validates headers with arktype-validator since the validator will respond with the headers, including the cookie header, if there is an error.

Steps to reproduce:
1. Define endpoints:
```
app.get("/"), async (c) => {
   setCookie(c, "SECRET", "123", { httpOnly: true });
   c.text("OK");
}

app.get("/validate"), arktypeValidator("header", type({ "User-Agent": "string" })), async (c) => {
  // Will error since the casing is incorrect :-/
   c.text("OK");
}
```

2. Go to `/` to set the secret cookie.
3. Go to `/validate` to leak the secret cookie.

Example response:
```
{
  "success": false,
  "errors": [
    {
      "data": {
        "accept": "application/json, text/plain, */*",
        "accept-encoding": "gzip, deflate, br, zstd",
        "accept-language": "en-US,en;q=0.5",
        "cache-control": "no-cache",
        "connection": "close",
        "content-length": "37",
        "content-type": "application/json",
        "cookie": "SECRET=123",
        "host": "localhost:8081",
        "origin": "http://localhost:8080",
        "pragma": "no-cache",
        "priority": "u=0",
        "referer": "http://localhost:8080/validate",
        "sec-fetch-dest": "empty",
        "sec-fetch-mode": "cors",
        "sec-fetch-site": "same-origin",
        "user-agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0"
      },
      "path": ["User-Agent"],
      "code": "required",
      "missingValueDescription": "a string",
      "meta": {},
      "expected": "a string",
      "actual": "missing",
      "problem": "must be a string (was missing)",
      "message": "value at [\"User-Agent\"] must be a string (was missing)"
    }
  ]
}
```

### What is the expected behavior?

The cookies not to be in the response.

### What do you see instead?

I see a couple of options:
1. Do nothing and document this.
2. Filter the cookie header from being included in the response on error.
3. Don't include any headers at all in the response.

[...]

PR #1282, opened 2025-07-04
[standard-validator]: Fix arktype leaking headers 

MonsterDeveloper:
## Overview

As discussed #1241 and #1135, this PR fixes arktype headers leak in `standard-validator`. This is a step toward deprecating arktype-specific validator and relying solely on standard schema validator.

### Checks

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/standard-validator/__schemas__/arktype.ts
+++ b/packages/standard-validator/__schemas__/arktype.ts
@@ -26,7 +26,12 @@ const querySortSchema = type({
   order: "'asc'|'desc'",
 })
 
+const headerSchema = type({
+  'User-Agent': 'string',
+})
+
 export {
+  headerSchema,
   idJSONSchema,
   personJSONSchema,
   postJSONSchema,
--- a/packages/standard-validator/src/index.ts
+++ b/packages/standard-validator/src/index.ts
@@ -21,6 +21,10 @@ type Hook<
   c: Context<E, P>
 ) => TOrPromiseOfT<Response | void | TypedResponse<O>>
 
+const RESTRICTED_DATA_FIELDS = {
+  header: ['cookie'],
+}
+
 const sValidator = <
   Schema extends StandardSchemaV1,
   Target extends keyof ValidationTargets,
@@ -71,7 +75,32 @@ const sValidator = <
     }
 
     if (result.issues) {
-      return c.json({ data: value, error: result.issues, success: false }, 400)
+      let processedIssues = result.issues
+      
+      // Strip sensitive data for arktype schemas
+      if (schema['~standard'].vendor === 'arktype' && target in RESTRICTED_DATA_FIELDS) {
+        const restrictedFields = RESTRICTED_DATA_FIELDS[target as keyof typeof RESTRICTED_DATA_FIELDS] || []
+        
+        processedIssues = result.issues.map((issue) => {
+          if (
+            issue &&
+            typeof issue === 'object' &&
+            'data' in issue &&
+            typeof issue.data === 'object' &&
+            issue.data !== null &&
+            !Array.isArray(issue.data)
+          ) {
+            const dataCopy = { ...(issue.data as Record<string, unknown>) }
+            for (const field of restrictedFields) {
+              delete dataCopy[field]
+            }
+            return { ...issue, data: dataCopy }
+          }
+          return issue
+        }) as readonly StandardSchemaV1.Issue[]
+      }
+      
+      return c.json({ data: value, error: processedIssues, success: false }, 400)
     }
 
     return result.value as StandardSchemaV1.InferOutput<Schema>
--- a/packages/standard-validator/src/index.ts
+++ b/packages/standard-validator/src/index.ts
@@ -76,11 +76,12 @@ const sValidator = <
 
     if (result.issues) {
       let processedIssues = result.issues
-      
+
       // Strip sensitive data for arktype schemas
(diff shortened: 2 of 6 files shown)


packages/standard-validator/__schemas__/arktype.ts
```diff
@@ -26,7 +26,12 @@ const querySortSchema = type({
   order: "'asc'|'desc'",
 })
 
+const headerSchema = type({
+  'User-Agent': 'string',
```
yusukebe: Are you intentionally using a `User-Agent` that includes capital letters? Because it contains capital letters, indeed, it will cause a validation error when validating the value.

However, I think this is confusing. So, I think adding a comment or comparing a completely different key would be great.

packages/standard-validator/src/index.ts
```diff
@@ -21,6 +21,10 @@ type Hook<
   c: Context<E, P>
 ) => TOrPromiseOfT<Response | void | TypedResponse<O>>
 
+const RESTRICTED_DATA_FIELDS = {
```
yusukebe: This is one of my ideas.

How about having a more generic `ValidatorConfig`? That way, [the line 81](https://github.com/honojs/middleware/pull/1282/files#diff-433e08d30aa0d8aad5b194e037d3faf25ea854cf1c4b125ffb298a7dbbc71a29R81) `schema['~standard'].vendor === 'arktype'`, can be made more general instead of specific to arktype.

Like this:

```diff
diff --git a/packages/standard-validator/src/index.ts b/packages/standard-validator/src/index.ts
index cd24666..5df0476 100644
--- a/packages/standard-validator/src/index.ts
+++ b/packages/standard-validator/src/index.ts
@@ -21,8 +21,22 @@ type Hook<
   c: Context<E, P>
 ) => TOrPromiseOfT<Response | void | TypedResponse<O>>

-const RESTRICTED_DATA_FIELDS = {
-  header: ['cookie'],
+type ValidatorConfig = Record<
+  string,
+  {
+    restrictedDataFields: {
+      header: string[]
+      //...
+    }
+  }
+>
+
+const VALIDATOR_CONFIGS: ValidatorConfig = {
+  arktype: {
+    restrictedDataFields: {
+      header: ['cookie'],
+    },
+  },
 }

 const sValidator = <
@@ -77,10 +91,11 @@ const sValidator = <
     if (result.issues) {
       let processedIssues = result.issues

-      // Strip sensitive data for arktype schemas
-      if (schema['~standard'].vendor === 'arktype' && target in RESTRICTED_DATA_FIELDS) {
+      // Strip sensitive data based on validator configuration
+      const vendor = schema['~standard'].vendor
+      if (vendor in VALIDATOR_CONFIGS && target in VALIDATOR_CONFIGS[vendor].restrictedDataFields) {
         const restrictedFields =
-          RESTRICTED_DATA_FIELDS[target as keyof typeof RESTRICTED_DATA_FIELDS] || []
+          VALIDATOR_CONFIGS[vendor].restrictedDataFields[target as 'header'] || []

         processedIssues = result.issues.map((issue) => {
           if (
```

packages/standard-validator/__schemas__/arktype.ts
MonsterDeveloper: Right.. I've just pushed a fix where it's now `user-agent` everywhere, and also the tests mention it explicitly:

```ts
const req = new Request('http://localhost/headers', {
  headers: {
    // Not passing the User-Agent header to trigger the validation error
    Cookie: 'SECRET=123',
  },
})
```

packages/standard-validator/src/index.ts
MonsterDeveloper: It's a great idea, but the issue is that the shape of an error is different in each validation library. Thus the processing logic to strip out the sensitive data will be unique for each library as well.

When I analyzed the errors I've actually discovered that Valibot also includes sensitive data in error output—will fix it as well.

ArkType:
```ts
{
    ' arkKind': 'error',
    path: [ 'user-agent' ],
    data: {},
    nodeConfig: {},
    input: {
      code: 'required',
      missingValueDescription: 'a string',
      relativePath: [ 'user-agent' ]
    },
    code: 'required',
    missingValueDescription: 'a string',
    relativePath: [ 'user-agent' ]
  }
```

Zod:
```ts
{
    code: 'invalid_type',
    expected: 'string',
    received: 'undefined',
    path: [ 'user-agent' ],
    message: 'Required'
  }
```

Valibot:
```ts
{
    kind: 'schema',
    type: 'string',
    expected: 'string',
    received: 'undefined',
    message: 'Invalid type: Expected string but received undefined',
    path: [
      {
        type: 'object',
        origin: 'value',
        input: { cookie: 'SECRET=123' },
        key: 'user-agent'
      }
    ]
  }
```

Can't think of a workaround for now, apart from manually adding different checks for each of the [Standard Schema supported libraries](https://standardschema.dev/#what-schema-libraries-implement-the-spec)...

yusukebe: > but the issue is that the shape of an error is different in each validation library.

Ah, it's a pain point that Standard Schema couldn't resolve.

packages/standard-validator/src/sanitize-issues.ts
yusukebe, line 1: Creating this `src/sanitize-issues.ts` is a good idea!

yusukebe (approved):
LGTM!

yusukebe:
@MonsterDeveloper 

Sorry for being late. Looks good! Thank you!

PR #1287, opened 2025-07-07
fix(valibot-validator): import valibot from jsr

BarryThePenguin:
I'd probably encourage everyone to use `@hono/standard-validator`, but this should still be useful

> I noticed that `jsr:@hono/valibot-validator` currently depends on `npm:valibot`. Would it be possible to switch the dependency to `jsr:@valibot/valibot` instead? In addition as changes of [PR 1277](https://github.com/honojs/middleware/pull/1277/files#diff-02cdcfb2199788f089f47ead4befb17cdf868b406375d9d0ac555ab5194772ef)?
>  

 _Originally posted by @imcotton in [#740](https://github.com/honojs/middleware/issues/740#issuecomment-3037495198)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "version": "yarn version:npm && yarn version:jsr",
     "version:npm": "changeset version",
     "version:jsr": "yarn workspaces foreach --all --no-private --parallel run version:jsr",
-    "version:set": "tmp=$(mktemp) && jq --arg version $0 '.version = $version' $INIT_CWD/deno.json > $tmp && mv $tmp $INIT_CWD/deno.json",
+    "version:set": "tmp=$(mktemp) && jq --arg version $0 '.version = $version' $INIT_CWD/deno.json > $tmp && mv $tmp $INIT_CWD/deno.json && prettier --write $INIT_CWD/deno.json",
     "typecheck": "yarn tsc --build",
     "typecheck:clean": "yarn tsc --build --clean",
     "typecheck:watch": "yarn tsc --build --watch",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "version": "yarn version:npm && yarn version:jsr",
     "version:npm": "changeset version",
     "version:jsr": "yarn workspaces foreach --all --no-private --parallel run version:jsr",
-    "version:set": "tmp=$(mktemp) && jq --arg version $0 '.version = $version' $INIT_CWD/deno.json > $tmp && mv $tmp $INIT_CWD/deno.json",
+    "version:set": "tmp=$(mktemp) && jq --arg version $0 '.version = $version' $INIT_CWD/deno.json > $tmp && mv $tmp $INIT_CWD/deno.json && prettier --write $INIT_CWD/deno.json",
```
BarryThePenguin, line 19: I noticed that prettier doesn't like the output formatting of `jq`, so I've updated this to run `prettier` on the changed `deno.json`

imcotton:
Thank you for the quick follow-up fix, much appreciated!

Also, thanks for pointing out `jsr:@hono/standard-validator`. I'll definitely give it a try later. If I may make one more suggestion: it would be great if it used `jsr:@standard-schema/spec` instead of `npm:@standard-schema/spec` as it currently does in ([v0.1.2](https://jsr.io/@hono/standard-validator@0.1.2)) :eyes:.

yusukebe:
@BarryThePenguin Thanks!

PR #1289, opened 2025-07-07
test: add `config.test.dir`

BarryThePenguin:
In Vitest v4 the default exclude pattern is going to be simplified, see https://github.com/vitest-dev/vitest/pull/6287

The recommendation is to use `config.test.dir` instead, as it's faster.


yusukebe (approved):
LGTM!

yusukebe:
It was simplified. Good!

PR #1297, opened 2025-07-09
test: vitest include

BarryThePenguin:
Similar to #1289, the default exclude pattern in Vitest v4 will change, so this PR adds explicit an `test.include` configuration.

I've been working on vitest-dev/vitest-ecosystem-ci#42, which runs our tests against the Vitest `main` branch. Merging this PR means our tests should pass with the latest version of Vitest


vitest.config.ts
```diff
@@ -9,7 +9,6 @@ export default defineConfig({
         autoUpdate: true,
       },
     },
-    dir: 'src',
```
BarryThePenguin, line 12: Vitest does some `cwd` magic when running test workers, so this didn't do what I thought it would..

vitest.config.ts
```diff
@@ -9,7 +9,6 @@ export default defineConfig({
...
-    projects: ['packages/*', '!packages/bun-transpiler'],
+    projects: ['packages/*/vitest.config.ts'],
```
BarryThePenguin, line 12: Only include projects that actually specify a `vitest.config.ts`

packages/esbuild-transpiler/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineProject } from 'vitest/config'
+
+export default defineProject({
+  test: {
+    globals: true,
+    include: ['src/**/*.test.ts'],
+  },
+})
```
BarryThePenguin, line 8: This project was already included, adding `vitest.config.ts` just ensures it is still included with these changes

packages/eslint-config/package.json
```diff
@@ -29,8 +29,5 @@
     "eslint-plugin-import-x": "^4.1.1",
     "eslint-plugin-n": "^17.10.2",
     "typescript-eslint": "^8.27.0"
-  },
-  "devDependencies": {
-    "hono": "^4.8.4"
```
BarryThePenguin, line 34: A little cleanup 🧹

packages/graphql-server/vitest.config.ts
```diff
@@ -1,9 +1,9 @@
-import { defaultExclude, defineProject } from 'vitest/config'
+import { defineProject } from 'vitest/config'
 
 export default defineProject({
   test: {
-    exclude: [...defaultExclude, 'bun_test'],
```
BarryThePenguin, line 5: Using `include` removes the need for this `exclude`

BarryThePenguin (commented):
I have thought about adding a shared vitest config at some point, but I think it's better that the config for each project remains seperate for now

yusukebe (approved):
LGTM!

yusukebe:
> I've been working on vitest-dev/vitest-ecosystem-ci#42, which runs our tests against the Vitest `main` branch. Merging this PR means our tests should pass with the latest version of Vitest

Cool. That is a nice idea! 

This PR looks good to me. Merging. Thanks!

PR #1301, opened 2025-07-11
fix(session): types for SessionOptions cookie methods

BarryThePenguin:
Changes the types for the `deleteCookie`, `getCookie`, and `setCookie` methods to match their usage

### DIFF
--- a/packages/session/src/index.ts
+++ b/packages/session/src/index.ts
@@ -37,9 +37,9 @@ export interface SessionOptions<Data> extends SessionEvents<Data> {
    */
   secret?: string | EncryptionKey
 
-  deleteCookie?: typeof cookie.deleteCookie
-  getCookie?: typeof cookie.getCookie
-  setCookie?: typeof cookie.setCookie
+  deleteCookie?: (c: Context, name: string, opt?: CookieOptions) => void
+  getCookie?: (c: Context, name: string) => string | undefined
+  setCookie?: (c: Context, name: string, value: string, opt?: CookieOptions) => void
 }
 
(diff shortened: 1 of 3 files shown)


packages/session/src/index.ts
```diff
@@ -37,9 +37,9 @@ export interface SessionOptions<Data> extends SessionEvents<Data> {
    */
   secret?: string | EncryptionKey
 
-  deleteCookie?: typeof cookie.deleteCookie
-  getCookie?: typeof cookie.getCookie
-  setCookie?: typeof cookie.setCookie
+  deleteCookie?: (c: Context, name: string, opt?: CookieOptions) => void
+  getCookie?: (c: Context, name: string) => string | undefined
+  setCookie?: (c: Context, name: string, value: string, opt?: CookieOptions) => void
```
BarryThePenguin, line 42: The [`GetCookie` type from `hono`](https://github.com/honojs/hono/blob/530ab09ae10caf33903dfb677dff239df01d5ded/src/helper/cookie/index.ts#L10-L14) makes it difficult use the `getCookie` option. For example, if I want to set `prefixOptions`..

```ts
useSession({
  getCookie(c, key) {
    return getCookie(c, key, 'secure')
  },
})
```

..I see the following Typescript error..

```
Type '(c: Context<any, any, {}>, key: string) => string | undefined' is not assignable to type 'GetCookie'.
  Target signature provides too few arguments. Expected 2 or more, but got 1.
```

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Indeed, this should be fixed. Thanks!

PR #1324, opened 2025-07-20
feat(auth-js): Support async ConfigHandler

jamestalmage:
This allows the config file to be generated async. 

Rationale / need: See (for example) the [split config](https://authjs.dev/guides/edge-compatibility#split-config) described in auth-js docs, where a database is used but not in all environments or not for all paths. Allowing an async config enables you to import the config that does / doesn't use the database based on whatever conditions you desire.

It also allows you to use things like cloudflare's [secret store](https://developers.cloudflare.com/secrets-store/integrations/workers/#call-get-on-the-binding-variable)(the api for which is exclusively async) for configuration variables.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/good-otters-press.md
+++ b/.changeset/good-otters-press.md
@@ -1,0 +1,5 @@
+---
+'@hono/auth-js': minor
+---
+
+Allow async authjs Config
--- a/packages/auth-js/src/index.test.ts
+++ b/packages/auth-js/src/index.test.ts
@@ -54,7 +54,7 @@ describe('Config', () => {
     app.use(
       '/*',
       initAuthConfig(async () => {
-        await new Promise(resolve => setTimeout(resolve, 1));
+        await new Promise((resolve) => setTimeout(resolve, 1))
         return {
           basePath: '/api/auth',
           providers: [],
--- a/packages/auth-js/src/index.test.ts
+++ b/packages/auth-js/src/index.test.ts
@@ -47,6 +47,27 @@ describe('Config', () => {
     expect(res.status).toBe(200)
   })
 
+  it('Should allow async ConfigHandler', async () => {
+    globalThis.process.env = { AUTH_SECRET: 'secret' }
+    const app = new Hono()
+
+    app.use(
+      '/*',
+      initAuthConfig(async () => {
+        await new Promise(resolve => setTimeout(resolve, 1));
+        return {
+          basePath: '/api/auth',
+          providers: [],
+        }
+      })
+    )
+
+    app.use('/api/auth/*', authHandler())
+    const req = new Request('http://localhost/api/auth/signin')
+    const res = await app.request(req)
+    expect(res.status).toBe(200)
+  })
+
   it('Should return 401 is if auth cookie is invalid or missing', async () => {
     const app = new Hono()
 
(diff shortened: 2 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@jamestalmage 

That's a good improvement. Thank you!

jamestalmage:
Thanks @yusukebe - I've opened [a pull request](https://github.com/nextauthjs/next-auth/pull/13128) on auth.js that would allow secrets to be configured as other than strings... I'm thinking secret managers like [Cloudflare Secrets Store](https://developers.cloudflare.com/secrets-store/). When/if that is merged, are you open to a pull request that allows you to bind cloudflare secret stores using the same naming convention (i.e. `AUTH_<provider-id>_SECRET`)?

yusukebe:
Hi @jamestalmage 

Honestly, I'm not an actual user for this middleware, so I want to hear @divyam234's opinion, but adding the feature to change the name to look up for the auth, depending on the environment, is a good idea. I can't judge until I see your implementation, so please create a PR.

Alternatively, my other idea is to add a function for setting the auth secret value depending on a condition like this (I might be saying something out of line):

```ts
app.use(
  '/*',
  initAuthConfig({
    setAuthSecret: (c) => {
      if (c.env.MODE === 'PROD') {
        return c.env.AUTH_SECRET
      } else if (c.env.MODE === 'STAGE') {
        return c.env.AUTH_STAGE_SECRET
      }
      return c.env.AUTH_DEV_SECRET
    },
  })
)
```

jamestalmage:
Gotcha,
Looking forward to @divyam234's feedback.

Just to clarify a misunderstanding (my fault for not explaining better): The PR wouldn't be to change the names of any environment variables.

Auth.js collects a number of different environment variables (up to four per configured OAuth provider). You end up with a list like `AUTH_SECRET`, `AUTH_GITHUB_ID`, `AUTH_GITHUB_SECRET`, `AUTH_GOOGLE_SECRET`... etc. If you use the specified naming convention, you don't have to manually fetch those values from the environment and inject them into the provider (a nice feature, as it eliminates a bunch of boilerplate). However, when Auth.js finds a value, it simply assumes the value is a string, and injects it without inspecting it or processing it.

Cloudflare allows you to bind more than just strings to the environment. One of the options is [secrets](https://developers.cloudflare.com/secrets-store/). Cloudflare secrets are objects with a `secret.get()` method that returns `Promise<string>`. Without a change, using the cloudflare secret store means giving up all of the convenience offered by the Auth.js naming conventions. You have to fetch all those values and inject them manually.

My PR in the Auth.js repo separates the steps of scanning for environment variables and for injecting them into the OAuth provider objects. This would allow implementers of secret-manager plugins to scan the environment easily: They don't have to determine the configured providers, or which variable names to lookup. They just provide the logic to convert the secrets to strings. For cloudflare the check would just be

```js
forEachSecret(secret => {
   if (typeof secret === 'string') {
      return secret;
   }
   if (typeof secret?.get === 'function') {
      return secret.get()
   }
})
```

yusukebe:
@jamestalmage 

Thank you for clarifying. I can understand well. So, if [your PR](https://github.com/nextauthjs/next-auth/pull/13128) is merged, you can create a PR for this middleware, and I can review it.

PR #1328, opened 2025-07-28
feat(otel): Support selective HTTP header capture

tsuyuni:
## Summary

  - Add `captureRequestHeaders` and `captureResponseHeaders` options to control which HTTP
  headers are collected in OpenTelemetry spans
  - Prevent sensitive information (cookies, authorization tokens) from being logged by default
  - The 'Traceparent' and 'Tracestate' headers are always captured from requests

## Changes

  - Added `captureRequestHeaders?: string[]` option to specify which request headers to capture
  - Added `captureResponseHeaders?: string[]` option to specify which response headers to capture

  - Modified header collection logic to only capture explicitly specified headers
  - Added comprehensive test coverage for the new functionality

## Breaking Changes

  **⚠️ Breaking Change**: By default, no headers are captured anymore. Previously, all headers
  were captured automatically.

  To maintain the previous behavior, explicitly specify the headers you want to capture:

  ```typescript
  app.use(otel({
    captureRequestHeaders: ['content-type', 'user-agent'],
    captureResponseHeaders: ['content-type', 'cache-control']
  }))
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/shiny-views-grin.md
+++ b/.changeset/shiny-views-grin.md
@@ -1,0 +1,5 @@
+---
+'@hono/otel': patch
+---
+
+Add `captureRequestHeaders` and `captureResponseHeaders` options for selective header collection.
--- a/packages/otel/src/index.ts
+++ b/packages/otel/src/index.ts
@@ -64,8 +64,9 @@ export const otel = (options: OtelOptions = {}): MiddlewareHandler => {
       activeContext,
       async (span) => {
         options.captureRequestHeaders?.push('Traceparent', 'Tracestate')
+        const captureRequestHeaders = options.captureRequestHeaders?.map((h) => h.toLowerCase())
         for (const [name, value] of Object.entries(c.req.header())) {
-          if (options.captureRequestHeaders?.includes(name)) {
+          if (captureRequestHeaders?.includes(name)) {
             span.setAttribute(ATTR_HTTP_REQUEST_HEADER(name), value)
           }
         }
@@ -76,8 +77,9 @@ export const otel = (options: OtelOptions = {}): MiddlewareHandler => {
           span.updateName(`${c.req.method} ${c.req.routePath}`)
           span.setAttribute(ATTR_HTTP_ROUTE, c.req.routePath)
           span.setAttribute(ATTR_HTTP_RESPONSE_STATUS_CODE, c.res.status)
+          const captureResponseHeaders = options.captureResponseHeaders?.map((h) => h.toLowerCase())
           for (const [name, value] of c.res.headers.entries()) {
-            if (options.captureResponseHeaders?.includes(name)) {
+            if (captureResponseHeaders?.includes(name)) {
               span.setAttribute(ATTR_HTTP_RESPONSE_HEADER(name), value)
             }
           }
--- a/packages/otel/src/index.ts
+++ b/packages/otel/src/index.ts
@@ -9,6 +9,7 @@ import {
   ATTR_HTTP_ROUTE,
 } from '@opentelemetry/semantic-conventions'
 import type { MiddlewareHandler } from 'hono'
+import type { RequestHeader, ResponseHeader } from 'hono/utils/headers'
 import { createMiddleware } from 'hono/factory'
 import metadata from '../package.json' with { type: 'json' }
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: .changeset/shiny-views-grin.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/otel': patch
```
dahlia: I think it should be `minor` since you introduced new APIs.

```suggestion
'@hono/otel': minor
```

dahlia (commented):
Looks good in general except for a versioning thing.

packages/otel/src/index.ts
```diff
@@ -60,8 +63,11 @@ export const otel = (options: OtelOptions = {}): MiddlewareHandler => {
       },
       activeContext,
       async (span) => {
+        options.captureRequestHeaders?.push('Traceparent', 'Tracestate')
         for (const [name, value] of Object.entries(c.req.header())) {
-          span.setAttribute(ATTR_HTTP_REQUEST_HEADER(name), value)
+          if (options.captureRequestHeaders?.includes(name)) {
+            span.setAttribute(ATTR_HTTP_REQUEST_HEADER(name), value)
+          }
```
sushichan044: When using `Object.entries(c.req.header())`, the `name` in L67 will always be in **lowercase**.
Therefore, it might be a good idea to convert all values in `options.captureRequestHeaders` to lowercase beforehand to ensure accurate comparison. This is also helpful because users might unintentionally input header names that are not in lowercase.

packages/otel/src/index.ts
```diff
@@ -71,7 +77,9 @@ export const otel = (options: OtelOptions = {}): MiddlewareHandler => {
           span.setAttribute(ATTR_HTTP_ROUTE, c.req.routePath)
           span.setAttribute(ATTR_HTTP_RESPONSE_STATUS_CODE, c.res.status)
           for (const [name, value] of c.res.headers.entries()) {
-            span.setAttribute(ATTR_HTTP_RESPONSE_HEADER(name), value)
+            if (options.captureResponseHeaders?.includes(name)) {
+              span.setAttribute(ATTR_HTTP_RESPONSE_HEADER(name), value)
+            }
```
sushichan044: same comment as https://github.com/honojs/middleware/pull/1328/files#r2235939458

.changeset/shiny-views-grin.md
tsuyuni: Thank you for the review. I fixed it.

sushichan044 (commented):
I left a comment because I was concerned about how user input for ‎`captureRequestHeaders` is handled.
If you decide to make changes, the following kind of test might be helpful.

```ts

  it('Should handle not lowercased value in captureRequestHeaders and captureResponseHeaders', async () => {
    const app = new Hono()
    app.use(
      otel({
        tracerProvider,
        captureRequestHeaders: ['Content-Type', 'X-Custom-Header'],
        captureResponseHeaders: ['Content-Type', 'X-Response-Header'],
      })
    )
    app.get('/foo', (c) => {
      c.header('X-Response-Header', 'response-value')
      return c.text('foo')
    })

    memoryExporter.reset()
    await app.request('http://localhost/foo', {
      headers: {
        'Content-Type': 'application/json',
        'X-Custom-Header': 'custom-value',
      },
    })

    const spans = memoryExporter.getFinishedSpans()
    const [span] = spans
    expect(span.attributes[ATTR_HTTP_REQUEST_HEADER('content-type')]).toBe('application/json')
    expect(span.attributes[ATTR_HTTP_REQUEST_HEADER('x-custom-header')]).toBe('custom-value')
    expect(span.attributes[ATTR_HTTP_RESPONSE_HEADER('content-type')]).toBe(
      'text/plain; charset=utf-8'
    )
    expect(span.attributes[ATTR_HTTP_RESPONSE_HEADER('x-response-header')]).toBe('response-value')
  })
```

tsuyuni:
@sushichan044 
Thank you for the feedback! You're completely right.
Since Hono's header types use uppercase, I've updated the code to convert the capture arrays to lowercase for proper matching.

yusukebe:
@tsuyuni 

Thank you for the cool feature and the nice implementation!

@dahlia @sushichan044 Thank you for the quick reviews!

I'll merge this and release a new version.

PR #1336, opened 2025-07-29
chore(clerk-auth): deno publish

### LINKED ISSUE #1335: publish @hono/clerk-auth to jsr

@hono/clerk-auth

Hello sorry to be annoying, just wondering if we could get the jsr version published pretty please ?

Our main hono import comes from jsr and we would like to not mix npm with jsr hono versions.

BarryThePenguin:
Publish `@hono/clerk-auth` to JSR

Fixes #1335

### DIFF
--- a/.changeset/five-wings-join.md
+++ b/.changeset/five-wings-join.md
@@ -1,0 +1,5 @@
+---
+'@hono/clerk-auth': patch
+---
+
+Publish to JSR
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -8,7 +8,7 @@
     // "packages/bun-transpiler",
     // "packages/casbin",
     "packages/class-validator",
-    // "packages/clerk-auth",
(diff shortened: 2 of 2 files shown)


yusukebe:
Hey @BarryThePenguin 

I updated the files to avoid the following error blocking `deno publish --dry-run` with [the same method of `hono/request-id`](
https://github.com/honojs/hono/blob/main/src/middleware/request-id/index.ts).

```
found an ambient module, which is a global augmentation, which are not unsupported
```

What do you think of this change?

BarryThePenguin (commented):
That's a really smart idea! So only npm users will get the ambient module 👍🏻

This approach will unblock a lot of other packages too 🎉

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Okay! Let's go.

PR #1355, opened 2025-08-05
chore: add dependabot.yml

### LINKED ISSUE #1288: Automate dependency updates

_No response_

Would the Hono team be open to adding a tool to manage dependency updates?

- [Dependabot](https://github.com/dependabot)
- [Renovate](https://github.com/apps/renovate)

are the two tools that I'm familiar with.

Both have support for multiple package ecosystems like Bun, npm, and Yarn, while only Renovate appears to have support for Deno.

Dependabot has [some configuration](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference) eg; grouping and scheduling.

Renovate has [a lot of configuration options](https://docs.renovatebot.com/config-overview/). So many that they have [config presets](https://docs.renovatebot.com/config-presets/)

Automating dependency updates in this way does create more Pull Requests to review, which requires the effort and time of maintainers to manage.

BarryThePenguin:
Closes #1288

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: npm
+    allow:
+      - dependency-type: production
+    directory: /
+    schedule:
+      interval: weekly
+    versioning-strategy: widen
+
+  - package-ecosystem: npm
+    allow:
+      - dependency-type: development
+    directory: /
+    schedule:
+      interval: weekly
+    versioning-strategy: increase


.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
```
BarryThePenguin, line 7: Updates github actions on a weekly schedule

.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+
+  - package-ecosystem: npm
+    allow:
+      - dependency-type: production
+    directory: /
+    schedule:
+      interval: weekly
+    versioning-strategy: widen
```
BarryThePenguin, line 15: Updates `dependencies` on a weekly schedule using the `widen` version strategy

> Widen the allowed version requirements to include both the new and old versions, when possible. Typically, this only increases the maximum allowed version requirement.

This should only open a PR when there is a new major version of a dependency. There is also configuration to add a [`cooldown`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-) to delay the PR by a configurable number of days.

I think this is optional, as major versions typically require code changes too..

.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+
+  - package-ecosystem: npm
+    allow:
+      - dependency-type: development
+    directory: /
+    schedule:
+      interval: weekly
+    versioning-strategy: increase
```
BarryThePenguin, line 23: Updates `devDependencies` on a weekly schedule using the `increase` version strategy

> Leave the constraint if the original constraint allows the new version, otherwise, bump the constraint.

yusukebe:
@BarryThePenguin 

Looks good to me! This will be helpful. Thank you!

PR #1356, opened 2025-08-05
chore: remove overlapping dependabot config

BarryThePenguin:
Looks like there was an issue with the initial configuration 😩 

https://github.com/honojs/middleware/runs/47402472825

```
Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'. Ecosystem 'npm' has overlapping directories.
```

It should now only update `devDependencies`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,14 +8,6 @@ updates:
 
   - package-ecosystem: npm
     allow:
-      - dependency-type: production
-    directory: /
-    schedule:
-      interval: weekly
-    versioning-strategy: widen
-
-  - package-ecosystem: npm
-    allow:
       - dependency-type: development
     directory: /
     schedule:


.github/dependabot.yml
```diff
@@ -6,14 +6,6 @@ updates:
     schedule:
       interval: weekly
 
-  - package-ecosystem: npm
```
yusukebe, line 9: This means we can't set both `dependencies` and `devDependencies`?

BarryThePenguin: We can allow both by removing..

```
    allow:
      - dependency-type: development
```

In that case, I'd just sticking with `auto` for the [versioning strategy](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--)

yusukebe: Ah, I see. So, setting only `dependency-type: development` is okay.

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Thanks! Merging.

PR #1366, opened 2025-08-07
chore: consolidate tool versions

BarryThePenguin:
This consolidates the tool versions from different CI workflows into a single `.tool-versions` file

I noticed in #1357 that [`oven-sh/setup-bun@v2`](https://github.com/oven-sh/setup-bun) has an option for `bun-version-file: .tool-versions`. Both [`denoland/setup-deno@v2`](https://github.com/denoland/setup-deno) and [`actions/setup-node@v4`](https://github.com/actions/setup-node) have similar options too.

I think the `.tool-versions` concept originated in the [asdf](https://asdf-vm.com/manage/configuration.html#tool-versions) version manages, but has since been adopted by other tools too.

Including a `.tool-versions` should only affect CI, and if contributor already knows about `.tool-versions` files, it will ensure they have the correct tools installed. I personally use https://mise.jdx.dev/

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version-file: .tool-versions
       - run: yarn install
       - run: yarn format:fix
       - run: yarn lint:fix
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,4 @@
+bun 1.1.32
+# TODO: Update to v2.x when https://github.com/jsr-io/jsr-npm/issues/129 is resolved
+deno 2.3.7
+node 20.19.4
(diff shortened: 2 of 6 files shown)


.tool-versions
```diff
@@ -0,0 +1,4 @@
+bun 1.1.32
+# TODO: Update to v2.x when https://github.com/jsr-io/jsr-npm/issues/129 is resolved
+deno 2.3.7
+node 20.19.4
```
BarryThePenguin, line 4: Also note that `ci-ua-blocker-sync.yml` is using Node v22, so I plan to look into upgrading Node next 👀 

https://github.com/honojs/middleware/blob/0c0a99d47a8dbd2228b5cdc1182312ee1525ed2f/.github/workflows/ci-ua-blocker-sync.yml#L22

yusukebe: > I plan to look into upgrading Node next 👀

Good. We should do it.

.github/workflows/autofix.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version-file: .tool-versions
```
yusukebe, line 25: I didn't know `.tool-versions`. Looks good!

yusukebe:
@BarryThePenguin Thanks!

Issue #1373, opened 2025-08-10
Codecov CI issues


yusukebe:
Codecov running on CI needs improvement. This result shows it.

https://github.com/honojs/middleware/pull/1360/checks

### 1. `codecov/project` fails

<img width="1214" height="338" alt="Image" src="https://github.com/user-attachments/assets/4b116a71-85c9-4cf9-9100-7b880a88e7c6" />

### 2. All packages are covered

Ideally, only related packages should be covered.

<img width="1364" height="1418" alt="Image" src="https://github.com/user-attachments/assets/b18f7b58-c306-4c48-beac-023e2b75ab00" />

yusukebe:
Hi @BarryThePenguin 

Can you investigate it?

BarryThePenguin:
Sure thing 👍🏻

I've been waiting for vitest v4 to be released which should help somewhat.. #1265

It's now in beta, so maybe we can merge that?

yusukebe:
Ah, I see. It's a Vitest side issue! So, only `v4` can fix it? Or we can still use `v3` and fix it?

BarryThePenguin:
Yes, it's something to do with how Vitest generates coverage reports for each project. #1265 is passing using `v4.0.0-beta.2`, but failing with newer beta releases

[honojs/middleware](https://github.com/honojs/middleware) is regularly tested against `main` branch of Vitest. You can see what's failing https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/16928252703/job/47968182844

It looks to be specifically related to `@cloudflare/vitest-pool-workers`

yusukebe:
Ah, `vitest-dev/vitest-ecosystem-ci`! It's great that this repo is in it. So, is it a good idea to wait for the official release of `v4`?

BarryThePenguin:
Yes, I think it's a good idea to wait. Jumping in the beta will just create more maintenance headaches 🫨

PR #1382, opened 2025-08-12
core(ci): what changed since base ref

BarryThePenguin:
Looking into #1373, I realised we aren't uploading any coverage from the `main` branch, just from PRs. This leads to strange coverage reports that say "Report is 123 commits behind head on main", or comparing coverage from over a month ago.

This PR does a couple of things that should fix this..

- Use [`nrwl/nx-set-shas`](https://github.com/nrwl/nx-set-shas) action to get the base ref
- Use the base ref when running `what-changed` with the [`--since`](https://yarnpkg.com/cli/workspaces/list#options) flag
- Seperate `ci` workflow into a second `pr` workflow
  -  `ci` runs when a new commit is pushed to `main`
  - `pr` only runs for PRs

Using `nrwl/nx-set-shas` we can get the commit of the last _successful_ build. This makes sure we run tests and generate coverage for all the packages changed since that commit. Currently only pull requests are compared with `main`. While commits to `main` do not cause any tests to run. Using the commit of the last _successful_ build, we should start to see coverage uploaded from the `main` branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       actions: 'read'
       pull-requests: 'read'
     steps:
+      - uses: actions/checkout@v4
       - uses: nrwl/nx-set-shas@v4
         id: setSHAs
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,6 +17,7 @@ jobs:
       actions: 'read'
       pull-requests: 'read'
     steps:
+      - uses: actions/checkout@v4
       - uses: nrwl/nx-set-shas@v4
         id: setSHAs
 


.github/actions/what-changed/action.yml
```diff
@@ -0,0 +1,32 @@
+name: What Changed
+description: Get a list of changed packages
```
BarryThePenguin, line 2: Pulled this out into an action as it's used in both the `ci` and `pr` workflows

.github/workflows/ci-bun-transpiler.yml
```diff
@@ -29,5 +29,3 @@ jobs:
           fail_ci_if_error: true
           directory: ./packages/bun-transpiler/coverage
           flags: bun-transpiler
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
BarryThePenguin, line 33: These aren't used any more 🗑️

.github/workflows/ci.yml
```diff
@@ -1,62 +1,40 @@
...
 
 jobs:
   what-changed:
     runs-on: ubuntu-latest
     outputs:
       packages: ${{ steps.set-packages.outputs.packages }}
+    permissions:
+      contents: 'read'
+      actions: 'read'
+      pull-requests: 'read'
```
BarryThePenguin, line 15: `nrwl/nx-set-sha` needs these permissions to get the state of previous workflows/actions

.github/workflows/ci.yml
```diff
@@ -1,62 +1,40 @@
...
-        with:
-          cache: true
-          deno-version-file: .tool-versions
-      - run: deno install --no-lock
-      - run: deno publish --dry-run
+      - run: yarn
+      - run: yarn publint
+      - run: yarn typecheck
+      - run: yarn lint
+      - run: yarn format
```
BarryThePenguin, line 37: These tasks will be run in each workspace for a commit to `main`

.github/workflows/pr.yml
```diff
@@ -0,0 +1,83 @@
+name: pr
```
BarryThePenguin, line 1: This is pretty much what `ci.yml` was previously, but now just for pull requests

.github/workflows/pr.yml
```diff
@@ -0,0 +1,83 @@
...
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}
+  cancel-in-progress: true
```
BarryThePenguin, line 8: Allows cancelling in progress builds for the same pull request

BarryThePenguin (commented):
An alternative would be to use something like https://nx.dev/ or https://turborepo.com/ which are purpose built for running monorepo tasks for workspaces with changes

yusukebe:
@BarryThePenguin 

Thank you for your investigation and creating the PR. Looks good.

yusukebe:
Ooops: https://github.com/honojs/middleware/actions/runs/16908773232/job/47904722311#step:4:513

PR #1420, opened 2025-08-27
feat(zod-openapi): support method chaining in `OpenAPIHono`

DanielleHuisman:
Support method chaining in `OpenAPIHono`:

```ts
new OpenAPIHono()
    .onError(errorHandler)
    .notFound(notFoundHandler)
    .use(middleware)
    .openapi(createRoute({}))
```

Currently, these methods have `Hono` as return type, even though the actual instance returned by `return this` is `OpenAPIHono`. This PR overrides the return types for these methods from `Hono` to `OpenAPIHono`.

I had to copy and modify the `MiddlewareHandlerInterface` type from `hono` in order to make this work. If anyone can think of a better solution, please let me know.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/loud-lines-try.md
+++ b/.changeset/loud-lines-try.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
+feat: support method chaining in `OpenAPIHono`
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @DanielleHuisman 

Thank you for the PR. Regarding this, I've never intentionally implemented it because we have to copy the long lines from [`types.ts`](https://github.com/honojs/hono/blob/main/src/types.ts) as you did. The type definitions are super verbose and are possible to change. So, I did not use them to reduce the maintenance costs.

I know there is an inconsistency between the actual behavior and type definitions, but I can't accept this PR easily.

PR #1423, opened 2025-08-31
chore: add turbo and build cache

BarryThePenguin:
Adds [`turbo`](https://turborepo.com/) and build caching in github actions

This should provide a transition path to using `bun` by reducing our reliance on [`yarn workspace`](https://yarnpkg.com/cli/workspace) commands


packages/ua-blocker/script/prebuild.ts
```diff
@@ -71,6 +71,14 @@ ts.addSyntheticLeadingComment(
   ' This file is generated by scripts/get-robots-txt.ts. Do not edit manually.'
 )
 
+for (const statement of statements) {
+  ts.addSyntheticLeadingComment(
```
yusukebe, line 75: Ah, this is a great solution!

yusukebe (approved):
LGTM!

PR #1436, opened 2025-09-03
fix(zod-validator): correct fallback behavior when using `coerce`

### LINKED ISSUE #1370: Using `z.coerce` breaks the client types

@hono/zod-validator

0.7.2

4.8.12

Node 20.19.4

Minimal repro:

```ts
const app = new Hono().get(
    "/",
    zValidator(
        "query",
        z.object({
            tenant: z.enum(["abba", "baab"]),
            page: z.coerce.number(),
        }),
    ),
    (c) => {
        const query = c.req.valid("query");
        return c.json({ query });
    },
);

const client = hc<typeof app>("http://localhost:3000");
const index = client.index.$get({
    query: {
        tenant: "abba",
        page: "1",
    },
});
```

The expected type for parameter query of index is `{ tenant: "abba" | "baab", page: string; }`, but when using `z.coerce.number()` it becomes `query: { tenant: string | string[]; page: string | string[]; }`. This is especially dangerous because it's very easy to not notice the issue with the types on the client side, causing one to pass wrong values and TypeScript not noticing it.

The types do work correctly within the request handler.

### Additional information

Changing from `z.coerce.number()` to e.g. `z.string().transform((value) => parseInt(value, 10))` fixes the issue but is less than optimal.

yusukebe:
Fixes #1370

This PR enables preventing enum types from falling back to `string | string[]` when `coerce` is used.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/wise-masks-say.md
+++ b/.changeset/wise-masks-say.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-validator': patch
+---
+
+fix: correct fallback behavior when using `coerce`
--- a/packages/zod-validator/src/index.ts
+++ b/packages/zod-validator/src/index.ts
@@ -42,12 +42,20 @@ export const zValidator = <
       ? {
           [K in Target]?: In extends ValidationTargets[K]
             ? In
-            : { [K2 in keyof In]?: ValidationTargets[K][K2] }
+            : {
+                [K2 in keyof In]?: In[K2] extends ValidationTargets[K][K2]
+                  ? In[K2]
+                  : ValidationTargets[K][K2]
+              }
         }
       : {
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @xremming

I think this is good! Wit this fix, the type for parameter query of `index` in [your comment](https://github.com/honojs/middleware/issues/1370#issue-3303524301) will be `{ tenant: "abba" | "baab", page: string | string[]; }` This expected type, because the `tenant` is not fallbacked to `string | string[]` and the `page` type is okay since `z.coerce.number()` can not define the input type (it will be unknown).

This is the best solution we can do now without breaking changes. What do you think of this?

yusukebe:
There is no breaking change. I think we can merge this.

Issue #1463, opened 2025-09-17
Json body validation stops working when bundling with esbuild


eric-poitras:
### Which middleware has the bug?

@hono/zod-openapi

### What version of the middleware?

0.19.6

### What version of Hono are you using?

4.7.8

### What runtime/platform is your app running on? (with version if possible)

NodeJs 22 + esbuild 0.25.5 

### What steps can reproduce the bug?

1 - Build a project that uses validation on the Body
2 - Run it using a `tsx` and everything works correctly.
3 - Bundle it using `esbuild` -> everything almost works but the json validator no longer register.

```
esbuild --sourcemap --sources-content=false --bundle --outfile=./dist/index.js --platform=node --target=node22 ./src/index.ts
```

### What is the expected behavior?

Bundled version should works as the unbundled version.

### What do you see instead?

Bundled version does not works as the validator is not registred.

### Additional information

Instanceof operator here fails under esbuild as the class functions does not seems to be the same:
```
https://github.com/honojs/middleware/blob/main/packages/zod-openapi/src/index.ts#L509
```

I previously saw that when you have 2 versions of the lib loaded in the same repository but I checked and it's not the case in our case. It seems as far as I see a side effect of the bundling and the way it emulates CommonJS.

Projects usually use a typeguard instead of `instanceof` operator validate type. This usually provide more robusts solution. I will take a look and provide a PR if needed.

yusukebe:
Hi @eric-poitras 

Thank you for raising the issue.

> Projects usually use a typeguard instead of `instanceof` operator validate type. This usually provide more robusts solution. I will take a look and provide a PR if needed.

Makes sense. We'd happy you create a PR.

yusukebe:
Hi @eric-poitras 

Thank you for the PR. I've merged, but I confirmed this issue again, and I can't reproduce your problem with your instruction. Can you share the minimal project to reproduce it? I want to confirm this is an actual bug and your PR has been fixed.

As a comment https://github.com/honojs/middleware/pull/1465#issuecomment-3306371112, using `typeof` many times could be slower than `instanceof`.

eric-poitras:
Hi @yusukebe,

I confirm, as @shahradelahi  says, that the step to reproduce is not as simple as using `esbuild`.  It seems to be related to the way our specific project is structured that cause `esbuild` to generate artifacts that breaks the prototype chain.

[...]

PR #1465, opened 2025-09-17
fix(zod-openapi): use a type-guard for zod to avoid issues with esbuild

eric-poitras:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/giant-cups-grab.md
+++ b/.changeset/giant-cups-grab.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Use a typeguard for zod to avoid issues when bundling with esbuild.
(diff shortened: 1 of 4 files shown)


yusukebe:
@eric-poitras 

Awesome. Thank you for the contribution!

shahradelahi:
Hi @yusukebe

I was wondering if you actually reviewed this PR and whether you agree these changes are unnecessary. **I’m not a hater—please don’t take this personally.** I just don’t understand the reasoning.

I read issue #1463, and I’m currently using `esbuild` to bundle my production API package and couldn’t reproduce the reported problem. Could the author be using bad practices? Using the `typeof` expression five times instead of `instanceof` could be significantly slower.

yusukebe:
@shahradelahi 

Sorry. I can't reproduce the problem. I'm working on it again. Thank you for pointing that out.

PR #1491, opened 2025-09-30
fix(oidc-auth): ensure revokeSession clears domain-scoped cookies with Domain attribute (closes #1490)

### LINKED ISSUE #1490: oidc-auth revokeSession doesn’t clear domain-scoped cookie in Chrome (behind proxy), cookie persists

@hono/oidc-auth

1.7.x (also observed on 1.6.x)

4.7.x

AWS Lambda (Node.js) behind CloudFront (Function URL origin), custom domain

### Description
When using @hono/oidc-auth behind a reverse proxy (CloudFront) and setting a domain-scoped session cookie (`OIDC_COOKIE_DOMAIN`), calling `revokeSession` does not reliably clear the cookie in Chrome. The `Set-Cookie` sent on logout lacks the `Domain` attribute, so Chrome retains the original domain-scoped cookie.

This appears related to proxy domain/host handling discussed in a previous issue about incorrect redirect/continue URL behind proxies, and manifests as a cookie-clearing problem in Chrome when cookies are scoped to a parent or custom domain. See related issue: [oidc-auth don't work behind proxy. Incorrect redirect on success #1215](https://github.com/honojs/middleware/issues/1215).

### Steps to reproduce
1. Run Hono with @hono/oidc-auth behind a proxy (e.g., CloudFront) on a custom domain, and set:
   - `OIDC_COOKIE_NAME=oidc-auth`
   - `OIDC_COOKIE_PATH=/`
   - `OIDC_COOKIE_DOMAIN=dev.gartenkreis.de` (or `.example.com`)
2. Complete OIDC login so the cookie is set as domain-scoped (`Domain=.example.com`).
3. Hit `GET /api/auth/logout` (or `POST`) calling `revokeSession` and redirect.

Observed logout response headers include `Set-Cookie` like:

```http
Set-Cookie: oidc-auth=; Max-Age=0; Path=/
```

Note: `Domain` is missing.

4. In Chrome DevTools → Application → Cookies → <your domain>, the cookie still persists. Navigating again shows you’re still authenticated.

- The session cookie should be fully removed after `revokeSession`:
  - `Set-Cookie` should match the original attributes (`Domain` and `Path`).
  - Chrome should delete the cookie immediately.

### What do you see instead?

- `Set-Cookie` lacks `Domain`, so the previously set domain-scoped cookie is not cleared in Chrome.
- User remains logged in until cookie naturally expires or is explicitly cleared with matching `Domain`.

### Minimal code
Logout route:

```ts
app.get('/api/auth/logout', async (c) => {
  await revokeSession(c)
  return c.redirect('/')
})
```

Library behavior (compiled dist) shows `revokeSession` deletes only with path:

```js
// node_modules/@hono/oidc-auth/dist/index.js (excerpt)
var revokeSession = async (c) => {
  const env2 = getOidcAuthEnv(c);
  const session_jwt = getCookie(c, env2.OIDC_COOKIE_NAME);
  if (session_jwt !== void 0) {
    deleteCookie(c, env2.OIDC_COOKIE_NAME, { path: env2.OIDC_COOKIE_PATH });
    // ...
  }
  c.set("oidcAuth", null);
};
```

### Workaround
After `revokeSession`, explicitly clear both host-only and domain-scoped variants and include an absolute `Expires` date:

```ts
deleteCookie(c, cookieName, { path: cookiePath }) // host-only
deleteCookie(c, cookieName, { path: cookiePath, domain: cookieDomain }) // domain-scoped
setCookie(c, cookieName, '', {
  path: cookiePath,
  domain: cookieDomain,
  secure: true,
  httpOnly: true,
  sameSite: 'Lax',
  maxAge: 0,
  expires: new Date(0),
})
```

Additionally set response no-cache headers to avoid `304 Not Modified` and ensure `Set-Cookie` isn’t bypassed:
- `Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate`
- `Pragma: no-cache`
- `Expires: 0`

### Suggested fix
- In `revokeSession`, call `deleteCookie` with both `path` and `domain` when `OIDC_COOKIE_DOMAIN` is set:
  - `deleteCookie(c, name, { path: OIDC_COOKIE_PATH, domain: OIDC_COOKIE_DOMAIN })`
- Optionally, also emit a `Set-Cookie` with empty value, `Max-Age=0` and `Expires=epoch` that includes `Domain` and `Path` to satisfy Chrome’s stricter behavior.
- Consider clearing any auxiliary cookies (e.g., “continue”) with matching `Domain/Path`.
- Document that deletion must mirror original cookie attributes.

### Why this is likely a library bug
- The cookie is initially set as domain-scoped when `OIDC_COOKIE_DOMAIN` is configured, but `revokeSession` deletes only with `path`, omitting `Domain`. Browsers (notably Chrome) require matching `Domain/Path` to remove cookies. This mismatch leads to persistent cookies after logout.

### Related
- Proxy host/cookie domain handling likely related: [oidc-auth don't work behind proxy. Incorrect redirect on success #1215](https://github.com/honojs/middleware/issues/1215)

SergKam:
This PR fixes the issue https://github.com/honojs/middleware/issues/1490 where revokeSession did not clear domain-scoped cookies when OIDC_COOKIE_DOMAIN is set. Now, the cookie is deleted with both path and domain

### DIFF
--- a/.changeset/eager-papers-pick.md
+++ b/.changeset/eager-papers-pick.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': patch
+---
+
+fix(oidc-auth): ensure revokeSession clears domain-scoped cookies with Domain attribute
--- a/packages/oidc-auth/src/index.test.ts
+++ b/packages/oidc-auth/src/index.test.ts
@@ -611,6 +611,28 @@ describe('RevokeSession()', () => {
       new RegExp(`oidc-auth=; Max-Age=0; Path=${MOCK_COOKIE_PATH}($|,)`)
     )
   })
+  test('Should revoke domain-scoped session cookie with Domain attribute', async () => {
+    const MOCK_COOKIE_DOMAIN = 'example.com'
+    process.env.OIDC_COOKIE_DOMAIN = MOCK_COOKIE_DOMAIN
+
+    const req = new Request('http://localhost/logout', {
+      method: 'GET',
+      headers: { cookie: `oidc-auth=${MOCK_JWT_ACTIVE_SESSION}` },
+    })
+
+    const res = await app.request(req)
+    expect(res.status).toBe(200)
+
+    const setCookieHeader = res.headers.get('set-cookie')
+    expect(setCookieHeader).toBeDefined()
+
+    const cookieStrings = setCookieHeader?.split(', ')
+    const domainCookie = cookieStrings?.find((c) => c.includes(`Domain=${MOCK_COOKIE_DOMAIN}`))
+
+    expect(domainCookie).toBeDefined()
+    expect(domainCookie).toContain('Max-Age=0')
+    expect(domainCookie).toContain('Path=/')
+  })
 })
 describe('initOidcAuthMiddleware()', () => {
   test('Should error if not called first in context', async () => {
--- a/packages/oidc-auth/src/index.ts
+++ b/packages/oidc-auth/src/index.ts
@@ -290,7 +290,15 @@ export const revokeSession = async (c: Context): Promise<void> => {
   const env = getOidcAuthEnv(c)
   const session_jwt = getCookie(c, env.OIDC_COOKIE_NAME)
   if (session_jwt !== undefined) {
+    // Always delete host-only cookie
     deleteCookie(c, env.OIDC_COOKIE_NAME, { path: env.OIDC_COOKIE_PATH })
+    // If domain is set, also delete domain-scoped cookie
+    if (env.OIDC_COOKIE_DOMAIN) {
+      deleteCookie(c, env.OIDC_COOKIE_NAME, {
+        path: env.OIDC_COOKIE_PATH,
+        domain: env.OIDC_COOKIE_DOMAIN,
+      })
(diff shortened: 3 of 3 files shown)


yusukebe:
@SergKam Thank you for the issue and PR!

Hey @hnw, can you review this?

hnw:
LGTM. Thanks for the solid PR.

This looks good. It effectively addresses the reported issue, and the new test case is a great addition for maintainability.

I also agree with the current approach of using two separate `deleteCookie` calls. It's a robust way to handle the edge case of a mid-session configuration change, which I've detailed below for context.

<details>
<summary>Note on the implementation</summary>

I initially considered suggesting that the two `deleteCookie` calls could be consolidated. However, I realized your approach is more robust because it correctly handles the edge case where the server configuration changes from host-only to domain-scoped during a user's session. A single call based on the new configuration would fail in that scenario. This defensive implementation is well-thought-out.
</details>

yusukebe:
@SergKam @hnw 

Thank you! I'll merge this and release a new version.

PR #1502, opened 2025-10-09
fix(zod-validator): Fix error type inference for Zod v4

BThero:
I have come across a small issue when using `zValidator` in combination with Zod v4. 

image
 
The error type is not properly deduced and fallbacks to `unknown`. The bug is as follows: 

- `z4.$ZodError<T>` expects a generic type `T` which should be the output type of the schema. 
- The generic type was not supplied.

I decided not to bother creating a separate issue. However, if you would still prefer that I create an issue, I will happily do it.

- [x] Add tests
- [x] Run tests (I only updated type tests, so there was no need to "run" them)


yusukebe:
@BThero Thank you! Merging and releasing a new version.

PR #1513, opened 2025-10-16
feat(otel): Align otel middleware with opentelemetry's semantic conventions

mrlorentx:
## What
This pull request does some relatively dramatic refactoring of the OpenTelemetry middleware in order to align it with [OpenTelemetry's semantic conventions](https://opentelemetry.io/docs/specs/semconv/).

## Why
The main reasoning here is to standardize telemetry in a way that fits into the conventions. While many projects are not following them, I feel like there's big value in being able to provide developers with a standard middleware that can take away a solid blob of static but in many cases necessary complexity.

## How
By replacing the legacy `otel` middleware with the new `httpInstrumentationMiddleware` which provides slightly more robust tracing and metrics instrumentation, using sane defaults but allowing configuration. The mission here is not to provide an abstraction level on top of _all telemetry_, but simply instrument the Hono layer in a very approachable way.

## Notes
While it may feel like a drastic rewrite, I took this path in order to make the middleware more robust. I've left the middleware open for extension in case there are future needs and considerations to handle - such as convention changes, or standard metrics to implement.

While some middleware instrument things like `http.server.request.body.size` I've kept it to a minimum, at least for now. Since most of this is based on usage I have at my place of work - a lot decisions have been driven of what I've seen a need for! I'm of course open to suggestions :)

- [X] Add tests
- [X] Run tests
- [X] `yarn changeset` at the top of this repo and push the changeset
- [X] Follow [the contribution guide](https://github.com/honojs/middleware?tab=readme-ov-file#how-to-contribute)

### DIFF
--- a/packages/otel/README.md
+++ b/packages/otel/README.md
@@ -29,7 +29,7 @@ export default app
 
 Since @opentelemetry/sdk-node is not supported on [Cloudflare Workers](https://workers.cloudflare.com/), you need to use [@microlabs/otel-cf-workers](https://github.com/evanderkoogh/otel-cf-workers) instead.
 
-So far I have not been able to see solid support for metrics in Cloudflare Workers. Any ideas here are very weolcome! Until then - the tracing instrumentation can be used!
+So far I have not been able to see solid support for metrics in Cloudflare Workers. Any ideas here are very welcome! Until then - the tracing instrumentation can be used!
 
 The following example shows how to use @microlabs/otel-cf-workers with [Honeycomb](https://www.honeycomb.io/):
 


mrlorentx:
It seems to me that the CI failures is due to other projects, if there's anything I can do here to help out just tell :)

Regarding the codecov report - ~~is that something you want me to do something about?~~ I've updated the PR!

yusukebe:
@mrlorentx Thank you for the PR!


Hey @dahlia @mrlorentx, can you review this?

packages/otel/README.md
```diff
@@ -30,6 +29,8 @@ export default app
 
 Since @opentelemetry/sdk-node is not supported on [Cloudflare Workers](https://workers.cloudflare.com/), you need to use [@microlabs/otel-cf-workers](https://github.com/evanderkoogh/otel-cf-workers) instead.
 
+So far I have not been able to see solid support for metrics in Cloudflare Workers. Any ideas here are very weolcome! Until then - the tracing instrumentation can be used!
```
dahlia: A nitpick.

```suggestion
So far I have not been able to see solid support for metrics in Cloudflare Workers. Any ideas here are very welcome! Until then - the tracing instrumentation can be used!
```

mrlorentx: Thanks, fixed now!

dahlia (approved):

yusukebe:
@mrlorentx @dahlia Thanks! I'll merge and release a new version.

PR #1522, opened 2025-10-21
fix(otel): header capture performance

IvanMicai:
## Summary
This PR optimizes header capture performance in the OpenTelemetry middleware

**Changes:**
- Converts header lookup from Array.includes() to Set.has() for both request and response headers
- Ensures traceparent and tracestate headers are always captured for proper trace propagation

## Problem

**Performance Issue:**
  - The previous implementation used Array.includes() for header matching, which has O(n) time complexity
  - On every request/response, the middleware iterates through all headers and performs array lookups
  - This becomes inefficient with large header lists or high request volumes

## Solution

**Performance Optimization:**
  - Create Set instances from header arrays during middleware initialization (one-time cost)
  - Use Set.has() for O(1) constant-time lookups instead of O(n) array scans
  - Headers are normalized to lowercase once during Set creation

**Fix Memory Leak:**
  - Move header array creation outside the request handler to middleware initialization
  - Only add traceparent and tracestate headers once during Set creation
  - Only create the Set when captureRequestHeaders is explicitly provided (opt-in behavior)

## The author should do the following, if applicable
- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/otel/src/index.ts
+++ b/packages/otel/src/index.ts
@@ -54,9 +54,7 @@ export const otel = (options: OtelOptions = {}): MiddlewareHandler => {
   const otelMetrics = createOtelMetrics(meter)
   const captureRequestHeaders = options.captureRequestHeaders
     ? new Set(
-        [...options.captureRequestHeaders, 'traceparent', 'tracestate'].map((h) =>
-          h.toLowerCase()
-        )
+        [...options.captureRequestHeaders, 'traceparent', 'tracestate'].map((h) => h.toLowerCase())
       )
     : undefined
   const captureResponseHeaders = options.captureResponseHeaders


yusukebe:
@IvanMicai Thank you for the PR.

Hi @dahlia @mrlorentx, sorry for pinging again. Can you review this?

mrlorentx (approved):
Nice catch! 

I think this looks good 👍

IvanMicai:
Thank you for the review.

After reviewing [PR #1513](https://github.com/honojs/middleware/pull/1513), which was recently merged (1.0.0), I can confirm that the refactoring already implemented the optimization this PR originally proposed.

Since the optimization is already in place, I'll close this PR. Thanks for the review @mrlorentx!

PR #1528, opened 2025-10-25
refactor: replace tsup with tsdown

BarryThePenguin:
It looks like [`tsup`](https://github.com/egoist/tsup) is no longer actively maintained, and they're recommending people migrate to [`tsdown`](https://github.com/rolldown/tsdown/) instead

### DIFF
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -505,7 +505,7 @@ export class OpenAPIHono<
         if (isJSONContentType(mediaType)) {
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           // @ts-ignore we can ignore the type error since Zod Validator's types are not used
-          const validator = zValidator('json', schema, hook)
+          const validator = zValidator('json', schema, hook) as MiddlewareHandler
           if (route.request?.body?.required) {
             validators.push(validator)
           } else {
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-bun-transpiler.yml
```diff
@@ -18,8 +18,7 @@ jobs:
         with:
           bun-version-file: .tool-versions
       - run: yarn workspaces focus hono-middleware @hono/bun-transpiler
-      - run: yarn workspace @hono/bun-transpiler build
-      - run: yarn workspace @hono/bun-transpiler publint
+      - run: yarn build --filter bun-transpiler
```
BarryThePenguin: `tsdown` also integrates `arethetypeswrong` and `publint` so we can remove the `publint` step ✨

packages/ajv-validator/tsdown.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'tsdown'
+
+export default defineConfig({
+  attw: true,
```
yusukebe, line 4: Ohh, tsdown includes the arethetypeswrong and publint?? It's cool!

packages/ajv-validator/tsdown.config.ts
yusukebe, line 1: Can all `tsdown.config.ts`s in the packages look for `tsdown.config.ts` in the top, and can we remove the tsdown.config.ts`s?

BarryThePenguin: Yes, that would be good except

- `@hono/zod-openapi` having a dependency on `@hono/zod-validator`, as CI needs to build `@hono/zod-validator` before building `@hono/zod-openapi`
- Each workspace has slightly different config, eg `entry`, `attw.profile`, etc..

I'll do my best to remove any duplicate configuration, but I think we'll still need to include most of them

yusukebe: I got it. Thank you for the explanation!

yusukebe:
Hi @BarryThePenguin 

I've updated to follow the latest main. This looks good to me! Let's go with this. Thanks!

PR #1534, opened 2025-10-28
docs(otel): Updates README.md

mrlorentx:
# Why
This PR updates the README for our new `otel` middleware to give more guidance to new users.

#What
Adds more in-depth information regarding configuration of the middleware and advanced options that could be used.

## Note
I've ignored pushing a changeset for this change, as it is a pure docs update. If you disagree with this - just ping and I'll push a patch change!

### DIFF
--- a/packages/otel/README.md
+++ b/packages/otel/README.md
@@ -31,33 +31,37 @@ export default app
 ```
 
 ## Configuration
+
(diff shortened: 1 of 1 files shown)


yusukebe:
@mrlorentx 

Thank you for the PR! No need to add a changeset. Looks good to me!

PR #1318, opened 2025-07-19
hono/mcp v0.2

MathurAditya724:
Resolves #1259 

- [x] Updated the `StreamableHTTPTransport` class with tests
- [ ] Auth Support

PR #1535, opened 2025-10-28
fix(mcp): unref keepAlive timer to allow server shutdown

BYK:
When there are open timers, it is impossible to shutdown a server in a graceful manner (at least in NodeJS as it keeps the process running). This patch calls `unref()` on the `keepAlive` timer so it does not block the process from shutting down.

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -17,7 +17,6 @@ import {
 } from '@modelcontextprotocol/sdk/types.js'
 import type { JSONRPCMessage, RequestId } from '@modelcontextprotocol/sdk/types.js'
 import type { Context } from 'hono'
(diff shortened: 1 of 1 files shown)


MathurAditya724:
I will be migrating these changes to #1318

MathurAditya724:
I have added the required changes in this branch. if it's gonna take more time for #1318 to be merged, we can just merge this simple fix for @BYK.

yusukebe:
@MathurAditya724

We have to go with the smallest possible PR. So, what about merging this PR first? If so, should any tests be added?

MathurAditya724:
Yes, we can merge this first. No test is need for this

yusukebe:
@MathurAditya724 Okay! And also, I think releasing a patch version is okay.

@BYK Thank you for the PR. I'll merge this.

PR #1545, opened 2025-11-05
ci(release): enable trusted publisher and provenance

BarryThePenguin:
I went through and double checked that everything is setup to support trusted publishing

It looks like enabling publish provenance is pretty straightforward too 🥳 

See https://github.com/honojs/hono/issues/4501


.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write # Used by changests to create releases
-      id-token: write # The OIDC ID token is used for authentication with JSR.
+      id-token: write # Used for authentication with both npm and jsr registries
```
BarryThePenguin, line 16: Looks like we already had the correct permissions to create an OIDC ID token

package.json
```diff
@@ -54,5 +54,5 @@
     "typia": "^9.7.2",
     "vitest": "^3.2.4"
   },
-  "packageManager": "yarn@4.9.2"
+  "packageManager": "yarn@4.10.3"
```
BarryThePenguin, line 57: The current version of yarn _should_ work, I've updated as there may be a few fixes that might be relevant, eg; https://github.com/yarnpkg/berry/pull/6911

.yarnrc.yml
```diff
@@ -2,4 +2,6 @@ nmHoistingLimits: workspaces
 
 nodeLinker: node-modules
 
-yarnPath: .yarn/releases/yarn-4.9.2.cjs
+npmPublishProvenance: true
```
yusukebe, line 5: `provenance`! This is good!

yusukebe:
@BarryThePenguin 

Thank you! I'll merge and set up the npm repository side.

BarryThePenguin:
Success! 🎉 

image

https://www.npmjs.com/package/@hono/hello

PR #1555, opened 2025-11-13
feat: introduce Cap'n Web Adapter

yusukebe:
This PR introduces an adapter/middleware for using Cap'n Web in the Hono application.

https://github.com/cloudflare/capnweb

You can use Cap'n Web with WebSocket helper, and it runs on Cloudflare Workers/Deno/Bun/Node.js etc.

```ts
import { Hono } from 'hono'
import { upgradeWebSocket } from 'hono/cloudflare-workers'
import { newRpcResponse } from '@hono/capnweb'
import { MyApiServer } from './my-api-server'

const app = new Hono()

app.all('/api', (c) => {
  return newRpcResponse(c, new MyApiServer(), {
    upgradeWebSocket,
  })
})

export default app
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/chilly-cups-roll.md
+++ b/.changeset/chilly-cups-roll.md
@@ -1,0 +1,5 @@
+---
+'@hono/capnweb': minor
+---
+
+initial release


BarryThePenguin:
Let me know if you need any help 👀

yusukebe:
@BarryThePenguin

Thanks! The coverage can't be reflected in Codecov since I think the test uses an actual server(?). But it's okay to merge right now.

yusukebe:
It's not bad. If you find any issues, let's handle them another time.

yusukebe:
@BarryThePenguin 

Sorry, one thing. The CI fails:

https://github.com/honojs/middleware/actions/runs/19885668585/job/56992321373#step:7:44

I think `@hono/node-ws` is related to it. It does not need to use workspaces' `@hono/node-ws`, so specify the version as `1.2.0`, and can we avoid the error?

PR #1571, opened 2025-11-22
docs(arktype-validator): add JSDoc for enhancing JSR Score

ysknsid25:

- [ ] ~~Add tests~~
- [x] Run tests

### DIFF
--- a/packages/arktype-validator/src/index.ts
+++ b/packages/arktype-validator/src/index.ts
@@ -1,8 +1,27 @@
+/**
+ * This module provides an ArkType validator middleware for Hono.
+ * It allows you to validate incoming requests (query, JSON, form data, headers, etc.)
+ * using ArkType schemas and provides a flexible hook for custom error handling.
+ *
+ * @module
+ */
 import { type } from 'arktype'
 import type { Type, ArkErrors } from 'arktype'
 import type { Context, MiddlewareHandler, Env, ValidationTargets, TypedResponse } from 'hono'
 import { validator } from 'hono/validator'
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
ysknsid25, line 1: Like Hono, it would be better to encourage middleware to fill in TS/JS Docs, right?🙋

yusukebe: I don't want to force the middleware author to avoid maintenance costs. If a middleware author does not add JS/TS Docs, I/we may have to notice it. It's trouble. We should make it depend on the middleware author.

ysknsid25: I see. Reverted it 🙏

packages/arktype-validator/src/index.ts
yusukebe, line 1: This is also a problematic issue. Adding JSDocs seems appropriate at this time. But if the APIs change, the maintainer or the PR author has to update the JSDoc.

This means "double management". We have to change both code and JSDoc. It may be suitable for the `hono` package, which many people watch, but for middleware, I think it's sometimes not good. Updating the code side, but JSDoc may not be updated. I'm not sure about whether it will occur, actually, but double management will cause some trouble for the maintainer.

ysknsid25: I see,  Yes sir.  I'll close this PR 🫡

yusukebe: @ysknsid25 Thank you for understanding!

PR #1589, opened 2025-12-01
fix(valibot-validator, standard-validator): Fix undefined response

miyaji255:
### Summary

I would like to apologize for mistakenly treating undefined as a Response in the previous PR.

Typically, it's `({ success: true }) => undefined`, `({ success: false }) => Response`, and undefined should be ignored. This PR will simply remove the undefined.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/witty-dolls-move.md
+++ b/.changeset/witty-dolls-move.md
@@ -1,0 +1,6 @@
+---
+'@hono/standard-validator': patch
+'@hono/valibot-validator': patch
+---
+
+Fix undefined Response type
--- a/packages/standard-validator/src/index.test.ts
+++ b/packages/standard-validator/src/index.test.ts
@@ -394,6 +394,56 @@ describe('Standard Schema Validation', () => {
             >
           >
         })
+
+        it('Should return hook result types with Env', () => {
+          const app = new Hono<{
+            Bindings: {
+              DATABASE_URL: string
+            }
+            Variables: {
+              userId: string
+            }
+          }>()
+
+          const schema = schemas.postJSONSchema
+          const route = app.post(
+            '/post',
+            sValidator('json', schema, (result, c) => {
+              if (!result.success) {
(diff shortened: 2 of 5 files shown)


yusukebe:
Hi @miyaji255 

Thank you for the PR!

I think that the code I pointed to with the following code is not typed as expected. What do you think of it?

```ts
import { number, object } from 'valibot'
import { Hono } from 'hono'
import { vValidator } from '@hono/valibot-validator'
import { hc } from 'hono/client'

const app = new Hono()
const schema = object({
  id: number()
})
const route = app.post(
  '/post',
  vValidator('json', schema, (result, c) => {
    if (!result.success) {
      // Comment out below
      // return c.json(`Invalid Post ID: ${result.output}` as const, 400)
    }
  }),
  (c) => {
    const data = c.req.valid('json')
    return c.json(data, 200)
  }
)

const client = hc<typeof route>('/')
const res = await client.post.$post({ json: { id: 123 } })

if (res.status === 200) {
  const data = await res.json()
} else {
  const data = await res.json() // <== { id: number } is correct???
}

export default app
```

miyaji255:
At the type level, there's no distinction between *success undefined* and *failure undefined*. That's why I gave up.

Therefore, I'm exporting FailedResponse to allow explicit type annotations like this:
```ts
const route = app.post(
  '/post',
  vValidator('json', schema, (result, c) => {
    if (!result.success) {
      // Uncomment below
      // return c.json(`Invalid Post ID: ${result.output}` as const, 400)
      return undefined as unknown as FailedResponse<typeof schema>;
    }
  }),
  (c) => {
    const data = c.req.valid('json')
    return c.json(data, 200)
  }
)
```

yusukebe (approved):
LGTM!

yusukebe:
@miyaji255 

Thank you for the explanation. Let's go with this. Thank you!

PR #1596, opened 2025-12-02
fix(mcp): updated the package versions in hono/mcp

### LINKED ISSUE #1588: Use Zod 4 (using backwards compatibility) for MCP middleware

@hono/mcp

The Model Context Protocol SDK for TypeScript supports Zod 4 using backwards compatibility:

https://github.com/modelcontextprotocol/typescript-sdk/releases/tag/1.23.0

that means middlwaree Zod dependency could be upgraded too.

MathurAditya724:
Closes #1588

### DIFF
--- a/packages/mcp/deno.json
+++ b/packages/mcp/deno.json
@@ -8,8 +8,7 @@
   },
   "imports": {
     "hono": "jsr:@hono/hono@^4.10.6",
-    "hono-rate-limiter": "jsr:@hono-rate-limiter/hono-rate-limiter@^0.4.2",
-    "zod": "npm:zod@^4.0.0"
+    "hono-rate-limiter": "jsr:@hono-rate-limiter/hono-rate-limiter@^0.4.2"
   },
   "publish": {
     "include": ["deno.json", "README.md", "src/**/*.ts"],


packages/mcp/deno.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hono/mcp",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
yusukebe: You don't need to change this since CI can fix it.

packages/mcp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hono/mcp",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
yusukebe: You don't need change this since Changeset will change this.

packages/mcp/deno.json
MathurAditya724: my bad, fixed it

yusukebe:
@MathurAditya724 

Do you know why the CI https://github.com/honojs/middleware/actions/runs/19852104490/job/56881037701?pr=1596 fails?

MathurAditya724:
Looking at it still, but my understanding is that deno changes the imports like `import z from "zod/v4"` to `import z from "npm:zod@${version in package.json}"/v4` and as I have changed this to `"^3.25.0 || ^4.0.0"` for having the backward compatibility, it's throwing an error.

Not sure how to denote this in deno, maybe we need to add this in deno.json file

yusukebe:
@MathurAditya724 

I see!

Hey @BarryThePenguin ! Do you have any idea?

packages/mcp/package.json
```diff
@@ -52,19 +52,19 @@
...
-    "@modelcontextprotocol/sdk": "^1.19.1",
+    "@modelcontextprotocol/sdk": "^1.23.0",
     "@types/node": "^24.2.1",
     "hono": "^4.10.6",
     "hono-rate-limiter": "^0.4.2",
     "tsdown": "^0.15.9",
     "typescript": "^5.8.2",
     "vitest": "^3.2.4",
-    "zod": "^3.25.34"
+    "zod": "^3.25.0 || ^4.0.0"
```
BarryThePenguin: There's no need to use a range here. It's a dev dependency, and only one version will be used, so just pick one

MathurAditya724:
Based upon my search, deno doesn't support `||` or `>=` in package versions. One possible solution is we add `npm:zod@^4.0.0` in `deno.json`. I have committed these changes if you wish to check it out.

If there is any better solution, happy to pivot

packages/mcp/package.json
MathurAditya724: My bad, we should include zod in `peerDependencies`

yusukebe: @MathurAditya724 Can you work on this? Pick up the one version for `devDependencies`.

MathurAditya724: Done

yusukebe:
@MathurAditya724 

Thank you! Merging.

PR #1600, opened 2025-12-03
fix(capnweb): build node-ws reference

BarryThePenguin:
Add `tsconfig` reference to build `@hono/node-ws` when building `@hono/capnweb`


yusukebe:
@BarryThePenguin Thank you!

Unfortunately, the CI fails with another reason:
https://github.com/honojs/middleware/actions/runs/19887203521/job/56997030129?pr=1600#step:6:17

Is it related to `@hono/mcp`? Can you take a look?

PR #1611, opened 2025-12-04
fix(publish): set access public with cli args

BarryThePenguin:
Similar to #1610, but this time with the [`--access public`](https://yarnpkg.com/cli/npm/publish#options) cli argument


yusukebe:
@BarryThePenguin 

Thanks! Let's try this.

BarryThePenguin:
🤞🏻

PR #1612, opened 2025-12-04
chore(yarn): update to v4.12.0

BarryThePenguin:
Lets try updating to the latest version of yarn. Unsure if this will help at all.. but it's worth trying 🙈

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version-file: .tool-versions
+          registry-url: https://registry.npmjs.org
 
       - uses: denoland/setup-deno@v2
         with:


.github/workflows/release.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version-file: .tool-versions
+          registry-url: https://registry.npmjs.org
```
BarryThePenguin, line 26: [`actions/setup-node`](https://github.com/actions/setup-node/tree/v6/?tab=readme-ov-file#usage) creates an `.npmrc` file when `registry-url` is set.

Setting it to the default registry will prevent `actions/changesets` from writing it's own `.npmrc` file with an invalid (`undefined`) authentication token 😩 

https://github.com/changesets/action/blob/04d574e831923498156e0b2b93152878063203a3/src/index.ts#L92-L96

I believe that's why we're seeing the "YN0033: No authentication configured for request" error message. There's more discussion about it in https://github.com/changesets/changesets/issues/1152

yusukebe: I see. Nice catch!

yusukebe (approved):

yusukebe:
@BarryThePenguin 

Thanks! I'll merge.

PR #1623, opened 2025-12-09
fix(node-ws): fix WebSocket not working with app.route()

poteboy:
## Summary

Fix WebSocket connections failing when the endpoint is registered under `app.route()`.

## Problem

When using `app.route()`, WebSocket connections fail with `400 Bad Request`.
```typescript
// ❌ This doesn't work
const subApp = new Hono()
subApp.get('/ws', upgradeWebSocket(() => ({ ... })))
app.route('/api', subApp)

// ✅ This works
app.get('/ws', upgradeWebSocket(() => ({ ... })))
```

## Cause

The original implementation stored a connection symbol in `c.env`:
```typescript
c.env[CONNECTION_SYMBOL_KEY] = connectionSymbol
```

However, this doesn't persist when the WebSocket endpoint is registered via `app.route()`. The exact reason is unclear, but it seems `c.env` reference changes in that case.

## Solution

Instead of writing to `c.env`, use a `WeakSet` to track the `request` object directly. Since `c.env.incoming` (the request object) remains the same reference even through `app.route()`, this approach works reliably.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/node-ws/src/index.ts
+++ b/packages/node-ws/src/index.ts
@@ -36,22 +36,20 @@ const CONNECTION_SYMBOL_KEY: unique symbol = Symbol('CONNECTION_SYMBOL_KEY')
  */
 export const createNodeWebSocket = (init: NodeWebSocketInit): NodeWebSocket => {
   const wss = new WebSocketServer({ noServer: true })
-  const waiterMap = new Map<
-    IncomingMessage,
-    { resolve: (ws: WebSocket) => void; connectionSymbol: symbol }
-  >()
+  const upgradeAllowed = new WeakSet<IncomingMessage>()
+  const waiterMap = new Map<IncomingMessage, (ws: WebSocket) => void>()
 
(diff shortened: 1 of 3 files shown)


yusukebe:
@poteboy Thank you for the PR!

Hey @nakasyou , can you review this? No rush.

nakasyou:
Hi @poteboy thank you for fixing a problem!

Hey @yusukebe, I have no time to review now, but I can review it after Saturday.

packages/node-ws/src/index.ts
nakasyou, line 30: This should be deleted.

nakasyou (commented):
Sorry for late reply. This is great idea to solve it, thank you.

packages/node-ws/src/index.ts
yusukebe: @poteboy Can you handle this?

yusukebe:
`@hono/node-ws` is deprecated. Closing.

PR #1624, opened 2025-12-13
docs(hono-mcp): fix variable name mismatch in Streamable HTTP example

sauryagur:
### What changed

Fixes a variable name mismatch in the Hono MCP documentation where
`mcp` was referenced instead of the declared `mcpServer`.

### Why

The example does not work when copied as-is. Updating the variable
name makes the snippet runnable out of the box.


yusukebe:
@sauryagur 

Thank you!!

PR #1651, opened 2025-12-23
chore: upgrade vitest to v4

BarryThePenguin:
Looks like there is finally some progress on making `@cloudflare/vitest-pool-workers` with with Vitest 4 https://github.com/cloudflare/workers-sdk/pull/11632 👀

### DIFF
--- a/packages/session/src/index.test.ts
+++ b/packages/session/src/index.test.ts
@@ -1,4 +1,5 @@
 import { Hono } from 'hono'
+import { testClient } from 'hono/testing'
 import { createTestSession } from './helper/testing'
 import type { Refresh } from './session'
 import type { SessionData } from '.'
@@ -952,6 +953,8 @@ describe('session.id', () => {
       return c.json({ id: c.var.session.id })
     })
 
+  const client = testClient(app)
+
   it('throws an error when session is not initialised', async () => {
     const app = new Hono().get('/session-id', useSession<TestData>({ secret }), (c) => {
       return c.json({ id: c.var.session.id })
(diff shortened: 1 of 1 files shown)


packages/firebase-auth/vitest.config.ts
```diff
@@ -1,4 +1,5 @@
-import { defineWorkersProject } from '@cloudflare/vitest-pool-workers/config'
+import { cloudflareTest } from '@cloudflare/vitest-pool-workers'
+import { defineProject } from 'vitest/config'
```
BarryThePenguin, line 2: `@cloudflare/vitest-pool-workers` is now a plugin, so `defineWorkersProject` no longer exists and can just use `defineProject`

packages/firebase-auth/vitest.config.ts
```diff
@@ -20,20 +21,19 @@ const firebasePlugin = {
...
-
-  plugins: [firebasePlugin],
+  plugins: [
+    cloudflareTest({
+      miniflare: {
+        compatibilityDate: '2025-03-10',
+        compatibilityFlags: ['nodejs_compat'],
+        kvNamespaces: ['PUBLIC_JWK_CACHE_KV'],
+      },
+    }),
```
BarryThePenguin, line 36: The plugin accepts the same options as the old `poolOptions.workers` configuration

packages/session/src/index.test.ts
```diff
@@ -447,31 +447,27 @@ describe('session.get(refresh) with storage', () => {
...
-        'Max-Age': '7200',
-        Path: '/',
-        SameSite: 'Lax',
-      })
-      expect(sessionCookie?.payload.exp).toBeGreaterThan(expiredSession.exp)
-      expect(sessionCookie?.payload.iat).toStrictEqual(expiredSession.iat)
-      expect(sessionCookie?.payload.sid).toStrictEqual(expiredSession.sid)
-      expect(storage.size).toBe(1)
-    },
-    { retry: 3 }
```
BarryThePenguin, line 473: Using the `retry` here never really helped with the flakiness of this test 🫠

yusukebe:
@BarryThePenguin 

Wow great!! We are super early.

The PR has not been merged, and this branch is using pkg-pr-new, but can I merge it? I believe it's okay to use pkg-pr-new.

BarryThePenguin:
I think it should be fine to merge. It should be safe to use pkg-pr-new as the dependencies won't change. If there is an issue, we can always revert this change.

I can keep an eye on the Wrangler PR and make updates if/when anything changes 👍🏻

yusukebe:
@BarryThePenguin 

Okay! Let's go!

PR #1653, opened 2025-12-24
chore(dependabot): group react dependencies

BarryThePenguin:
This should fix CI for #1647


yusukebe:
@BarryThePenguin Thanks!

PR #1662, opened 2025-12-25
fix(swagger-ui): set `<!doctype html>` when serving swagger-ui to avoid quirks mode

stefanprobst:
this sets `<!doctype html>` when serving swagger-ui.


yusukebe:
@stefanprobst 

Thanks! Looks good.

PR #1666, opened 2025-12-27
chore(eslint): suppress existing violations

BarryThePenguin:
Use [Bulk Suppressions](https://eslint.org/docs/latest/use/suppressions) to suppress all existing eslint rule violations. This allows rules to remain enabled and for any new violations to be reported by CI


.github/workflows/autofix.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           node-version-file: .tool-versions
       - run: yarn install
       - run: yarn format:fix
-      - run: yarn lint:fix
+      - run: yarn lint:fix --suppress-all --pass-on-unpruned-suppressions
```
BarryThePenguin: Allows autofix to pass when there are unused suppressions

yusukebe:
@BarryThePenguin 

I didn't know Bulk Suppressions! Seems to be good. Thanks.

PR #1668, opened 2025-12-28
chore: upgrade hono to 4.11.3

BarryThePenguin:


packages/ajv-validator/src/index.ts
```diff
@@ -82,7 +82,7 @@ export function ajvValidator<
 > {
   const ajv = new Ajv()
   const validate = ajv.compile(schema)
-
+  // @ts-expect-error not typed well
```
BarryThePenguin, line 85: It looks like there was a change in either v4.11.2 or v4.11.3 that caused the type of the `validator` function to change 🤔

MathurAditya724:
Can we integrate catalogs? As described here - https://yarnpkg.com/features/catalogs

Happy to create a seperate PR for this

BarryThePenguin:
> Can we integrate catalogs? As described here - https://yarnpkg.com/features/catalogs
> 
> Happy to create a seperate PR for this

Is there a specific problem you want to solve? Does dependabot support yarn catalogs?

Adopting catalogs makes sense, just wondering if there's anything we should be aware of. For example, switching to bun if it becomes feasible

packages/ajv-validator/src/index.ts
yusukebe: Yes. The `validator` type was changed to make the types correct. The type definition became complex. So using `@ts-expect-error` makes sense.

yusukebe:
Hey @BarryThePenguin Happy New Year!

Looks good to me. Thank you!

BarryThePenguin:
🎆 Happy New Year! 🥳

PR #1669, opened 2025-12-28
Feature/event emitter/env

BarryThePenguin:
Introduces `EmitterEnv` to simplify `Context` types

### DIFF
--- a/packages/event-emitter/src/index.ts
+++ b/packages/event-emitter/src/index.ts
@@ -10,15 +10,20 @@ export type EventHandler<T, E extends Env = Env> = (
   c: Context<E>,
   payload: T
 ) => void | Promise<void>
-export type EventHandlers<T> = { [K in keyof T]?: EventHandler<T[K]>[] }
+export type EventHandlers<T> = { [K in keyof T]?: EventHandler<T[K], EmitterEnv<T>>[] }
 export type EventPayloadMap = { [key: string]: unknown }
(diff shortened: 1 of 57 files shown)


packages/event-emitter/src/index.ts
```diff
@@ -33,9 +38,9 @@ export interface Emitter<EPMap extends EventPayloadMap> {
  * @returns The event handler.
  */
 export const defineHandler = <
-  EPMap extends EventPayloadMap,
```
BarryThePenguin, line 36: Having the `EPMap extends EventPayloadMap` constraint makes it difficult to type `Emitter` and `EmitterEnv` correctly due to the circular reference

yusukebe:
@BarryThePenguin Thank you for the PR.

Hey @DavidHavl, can you review this?

DavidHavl:
@BarryThePenguin By introducing the `EmitterEnv`, you have indeed created a circular reference.
Removing the `EventPayloadMap` constraint (which dictates that key needs to be a string) isn’t a solution because it disables autocompletion in some cases and introduces potential errors (the key can be anything).
The constraint ensures type safety without creating coupling.

DavidHavl (changes requested):
- By introducing the EmitterEnv, you have created a circular reference, so that is not desired.
- Removing the EventPayloadMap constraint (which dictates that key needs to be a string) isn’t a solution because it disables autocompletion in some cases and introduces potential errors (the key can be anything). The constraint ensures type safety without creating coupling.

PR #1675, opened 2026-01-01
fix(eslint-config): use correct export type

BarryThePenguin:
This should fix the build error in #1670


yusukebe:
@BarryThePenguin I see. Thanks!

PR #1676, opened 2026-01-02
chore(capnweb): build workspace reference

BarryThePenguin:
Fixes workspace references for `@hono/capnweb`

### DIFF
--- a/packages/capnweb/package.json
+++ b/packages/capnweb/package.json
@@ -47,7 +47,7 @@
   },
   "devDependencies": {
     "@hono/node-server": "^1.19.2",
-    "@hono/node-ws": "^1.2.0",
+    "@hono/node-ws": "workspace:^",
     "@types/ws": "^8.5.0",
     "capnweb": "^0.3.0",
     "hono": "^4.11.3",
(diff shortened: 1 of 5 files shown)


packages/capnweb/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "devDependencies": {
     "@hono/node-server": "^1.19.2",
-    "@hono/node-ws": "^1.2.0",
+    "@hono/node-ws": "workspace:^",
```
BarryThePenguin, line 50: This is in `devDependencies` so there's no need to publish this change

packages/capnweb/tsconfig.build.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {},
-  "references": [{ "path": "../node-ws/tsconfig.build.json" }]
```
BarryThePenguin, line 4: As `@hono/node-ws` is only used in tests, this references should be in `tsconfig.spec.json` and not in `tsconfig.build.json`

yusukebe: Ah, I see!

packages/capnweb/package.json
yusukebe: Why I did not use `workspace` was intended, but I forgot the reason 😇 , and I think this is good!

PR #1679, opened 2026-01-04
chore(typescript): add `@tsconfig/strictest`

BarryThePenguin:
- Adds [`@tsconfig/strictest`](https://github.com/tsconfig/bases/blob/a1854fcbfbca380870f20218bfbf94984a654560/bases/strictest.json) to `tsconfig.base.json`
- Disables `compilerOptions` for each package that isn't compatible with `@tsconfig/strictest`

### DIFF
--- a/packages/ajv-validator/tsconfig.spec.json
+++ b/packages/ajv-validator/tsconfig.spec.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "noUnusedLocals": false,
     "outDir": "../../dist/packages/ajv-validator",
     "types": ["vitest/globals"]
   },
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,4 +1,5 @@
 {
+  "extends": "@tsconfig/strictest",
   "compilerOptions": {
     "target": "ES2022",
     "module": "esnext",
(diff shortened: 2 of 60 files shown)


packages/ajv-validator/tsconfig.spec.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "noUnusedLocals": false,
```
BarryThePenguin, line 4: Tests often use type assertions like..

```ts
type verify = Expect<Equal<Expected, Actual>>
```

..which leaves a lot of unused declarations laying around.

One day I'd like to replace these with [`expectTypeOf`](https://vitest.dev/api/expect-typeof.html#expecttypeof) assertions

tsconfig.base.json
```diff
@@ -1,21 +1,17 @@
...
-    "esModuleInterop": true,
     "emitDecoratorMetadata": true,
     "emitDeclarationOnly": true,
     "experimentalDecorators": true,
     "forceConsistentCasingInFileNames": true,
     "noEmitOnError": true,
-    "strict": true,
-    "skipLibCheck": true,
-    "noUnusedLocals": false,
-    "noUnusedParameters": true,
```
BarryThePenguin, line 18: Safe to remove as these are the default values provided by `@tsconfig/strictest`

tsconfig.json
```diff
@@ -36,6 +36,7 @@
     { "path": "packages/session" },
     { "path": "packages/ssg-plugins-essential" },
     { "path": "packages/standard-validator" },
+    { "path": "packages/stytch-auth"},
```
BarryThePenguin, line 39: Looks like this was missing 👀

packages/ajv-validator/tsconfig.spec.json
yusukebe: Ah, I see. I also tried to replace these with Vitest's, but gave up 🙃

yusukebe:
@BarryThePenguin 

Thank you for editing a bunch of files. Thanks!

PR #1680, opened 2026-01-05
fix(eslint-config): enable curly rule explicitly

3w36zj6:
cf. https://github.com/honojs/hono/pull/4599#discussion_r2660225002

`@hono/eslint-config` specifies `'curly': ['error', 'all']`[^1], but this rule is actually disabled by `eslint-config-prettier`[^2]. According to the README, we need to manually override this setting if we want to enforce it[^3]. 

[^1]: https://github.com/honojs/middleware/blob/main/packages/eslint-config/index.js#L33  
[^2]: https://github.com/prettier/eslint-config-prettier/blob/main/index.js#L12  
[^3]: https://github.com/prettier/eslint-config-prettier#curly

### DIFF
--- a/.changeset/four-olives-cross.md
+++ b/.changeset/four-olives-cross.md
@@ -1,5 +1,5 @@
 ---
-'@hono/eslint-config': minor
+'@hono/eslint-config': patch
 ---
 
 enable curly rule explicitly


.changeset/four-olives-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/eslint-config': minor
```
yusukebe: I wonder whether we should mark this as `patch` or `minor,` but I think this should be `patch` because the change is small and not a new feature. Can you change this to `patch`?

yusukebe (approved):

yusukebe:
@3w36zj6 This is nice! What I want. Thank you!

PR #1681, opened 2026-01-06
fix: less strict template expressions

BarryThePenguin:
Use recommended options when restricting template expressions

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -92,6 +92,18 @@ const config = [
       '@typescript-eslint/no-require-imports': 'off',
       '@typescript-eslint/no-unused-vars': 'warn',
       '@typescript-eslint/no-var-requires': 'off',
+      
+      '@typescript-eslint/restrict-template-expressions': [
+        'error',
+        {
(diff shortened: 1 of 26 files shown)


packages/eslint-config/index.js
```diff
@@ -92,6 +92,18 @@ const config = [
...
+      '@typescript-eslint/restrict-template-expressions': [
+        'error',
+        {
+          allow: [{ name: ['Error', 'URL', 'URLSearchParams'], from: 'lib' }],
+          allowAny: true,
+          allowBoolean: true,
+          allowNullish: true,
+          allowNumber: true,
+          allowRegExp: true,
+        },
```
BarryThePenguin, line 105: These are just the default options for [`@typescript-eslint/restrict-template-expressions`](https://typescript-eslint.io/rules/restrict-template-expressions/)

```js
{
  allow: [{ name: ['Error', 'URL', 'URLSearchParams'], from: 'lib' }],
  allowAny: true,
  allowBoolean: true,
  allowNullish: true,
  allowNumber: true,
  allowRegExp: true,
}
```

The strict options are.. too strict 😅

yusukebe:
@BarryThePenguin 

Nice! The code bases became clean. LGTM! Thanks.

PR #1688, opened 2026-01-07
refactor(types): move `@types/*` dependencies into individual workspaces

BarryThePenguin:
Moving `@types/*` dependencies around and cleaning up some type assertions

### DIFF
--- a/.github/actions/what-changed/main.ts
+++ b/.github/actions/what-changed/main.ts
@@ -3,7 +3,7 @@
 import * as core from '@actions/core'
 import * as exec from '@actions/exec'
 
-const exclude = new Set(['@hono/bun-transpiler'])
+const exclude = new Set(['@hono/bun-compress', '@hono/bun-transpiler'])
 const since = core.getInput('since')
 
 const workspaces = await exec.getExecOutput('yarn workspaces list', [
--- a/.github/workflows/ci-bun-compress.yml
+++ b/.github/workflows/ci-bun-compress.yml
@@ -1,0 +1,30 @@
+name: ci-bun-compress
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/bun-compress/**'
+  pull_request:
+    branches: ['*']
+    paths:
+      - 'packages/bun-compress/**'
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        bun-version: [1.3.2, 1.3.3]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: ${{ matrix.bun-version }}
+      - run: yarn workspaces focus hono-middleware @hono/bun-compress
+      - run: yarn turbo --filter @hono/bun-compress build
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -42,9 +42,6 @@
     "@hono/eslint-config": "workspace:*",
     "@ryoppippi/unplugin-typia": "^2.6.5",
     "@tsconfig/strictest": "^2.0.8",
-    "@types/bun": "^1.0.0",
-    "@types/node": "^25.0.3",
-    "@types/ws": "^8.18.0",
```
BarryThePenguin, line 47: I can't remember why exactly.. but in #1077 and #1098 I moved these packages to the monorepo root.

I think it may have been related to `tsup` config needing to resolve types relative to the root? But we don't have that problem anymore..

packages/bun-compress/src/index.test.ts
```diff
@@ -224,8 +224,6 @@ async function decompressResponse(res: Response): Promise<string> {
   const readableStream = Readable.fromWeb(resBody)
   const decompressedStream = readableStream.pipe(createGunzip())
   const decompressedReadableStream = Readable.toWeb(decompressedStream)
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const decompressedResponse = new Response(decompressedReadableStream as any)
+  const decompressedResponse = new Response(decompressedReadableStream)
```
BarryThePenguin, line 227: No more type assertion ✨

.github/actions/what-changed/main.ts
```diff
@@ -3,7 +3,7 @@
 import * as core from '@actions/core'
 import * as exec from '@actions/exec'
 
-const exclude = new Set(['@hono/bun-transpiler'])
+const exclude = new Set(['@hono/bun-compiler', '@hono/bun-transpiler'])
```
yusukebe: Adding the `@hono/bun-compiler` is a typo?

yusukebe: Hey @BarryThePenguin !

Can you look at this?

BarryThePenguin: Ah, yes! Forgot about this.. There are a couple of things here..

- `@hono/bun-compiler` is a typo, yes. This should be `@hono/bun-compress`
- Bun v1.3.3 introduced [CompressionStream and DecompressionStream](https://bun.com/blog/bun-v1.3.3#compressionstream-and-decompressionstream), so the tests for `@hono/bun-compress` aren't actually testing the code.
- It looks like [`@hono/bun-compress`](https://www.npmjs.com/package/@hono/bun-compress) gets quite a few downloads, so it's probably worth keeping around

So I think it's probably worth the effort of testing the package on the older version of bun

yusukebe: @BarryThePenguin 

Got them! So, can you update this branch?

BarryThePenguin: Looks like that's sorted now 👀

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Thank you! Let's go!

PR #1698, opened 2026-01-13
fix(standard-validator): use unpinned `@standard-schema/spec`

ryuapp:
This is a peer dependency optimisation.
Additionally, as it was included in devDependencies, it had been updated to v1.1.0 in #1685, so I have removed it from the deps.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/packages/standard-validator/package.json
+++ b/packages/standard-validator/package.json
@@ -46,6 +46,7 @@
     "hono": ">=3.9.0"
   },
   "devDependencies": {
+    "@standard-schema/spec": "1.1.0",
     "arktype": "^2.1.29",
     "hono": "^4.11.3",
     "tsdown": "^0.15.9",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2321,6 +2321,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@hono/standard-validator@workspace:packages/standard-validator"
   dependencies:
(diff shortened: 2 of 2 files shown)


packages/standard-validator/package.json
```diff
@@ -42,11 +42,10 @@
   },
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
-    "@standard-schema/spec": "1.1.0",
+    "@standard-schema/spec": "^1.0.0",
     "hono": ">=3.9.0"
   },
   "devDependencies": {
-    "@standard-schema/spec": "1.1.0",
```
yusukebe, line 49: Is it okay to remove `@standard-schema/spec` in `devDependencies`? I wonder if it's better to specify `"@standard-schema/spec": "^1.0.0"` explicitly. But is it okay?

BarryThePenguin: It's best to be explicit and keep `@standard-schema/spec` in `devDependencies`. Otherwise it is depending on other packages in the monorepo to install it

ryuapp: Thank you for your comments. Fix 6e25cc327d0fc4f36d6dad4053c49340186dd86a

yusukebe:
@ryuapp @BarryThePenguin Thanks!

PR #1708, opened 2026-01-19
feat(arktype-validator): Fix/arktype match zod

WolfieLeader:

- [x] Add tests
- [x] Run tests

All tests pass locally. Changeset included for a major version bump due to behavior and typing changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ settings.local.json
 
 # Code editor
 .zed
+
+# MacOS
+.DS_Store


WolfieLeader:
Hey Yusuke Wada!

First of all, thank you for the amazing work on this project, it’s been a pleasure using it.

I wanted to ask about the current recommendation around validators like `zod-validator` / `arktype-validator` / `valibot-validator`.

Given that the platform now provides a standard validator with strong typing and good DX, is there still a practical reason to prefer or keep these external validators for most use cases?

If the `standard-validator` is now the recommended default, would it make sense to explicitly mention this in the docs - for example, clarifying when external validators are still beneficial (advanced schema reuse, cross-runtime sharing, etc.) versus when the `standard-validator` is sufficient?

I’d be happy to help update or improve the documentation around this if you think it makes sense.

Appreciate your time and feedback!

yusukebe:
Hey @WolfieLeader Thank you for the PR!

This is a bit of a trouble problem. We may have to switch to encourage the use of Standard Schema; on the other hand, Zod is very "standard" compared to others, and actually, we are now forced to maintain `@hono/zod-validator` 

I'll consider it again.

PR #1715, opened 2026-01-23
fix(eslint-config): less strict no base to string

BarryThePenguin:
Make [`@typescript-eslint/no-base-to-string`](https://typescript-eslint.io/rules/no-base-to-string/) less strict, eg; allowing `new URL().toString()`, `new Error().toString()`, etc

Similar to #1681


yusukebe:
@BarryThePenguin 

`no-base-to-string`! Ah, I see. Thanks!

PR #1726, opened 2026-02-01
fix(zod-openapi): Incomplete context internal error with x-www-form-urlencoded submissions

### LINKED ISSUE #1719: [hono/zod-openapi] Incomplete context internal error with x-www-form-urlencoded submissions

hono/zod-openapi

1.2.0

4.11.5

nodejs v22.x 

1- Create a route with a schema expecting a body
2- Submit a request to the server with `application/x-www-form-urlencoded` content-type and incorrect body

the api should return a 400 validation error and not crash

### What do you see instead?

500 internal error and error trace shows `Error: Context is not finalized. Did you forget to return a Response object or `await next()`?`

### Additional information

Upon investigating the code I found that the `await validate(...)` in [here]( https://github.com/honojs/middleware/blob/1c27ef687ad3cdfe6f709070dbc73e26b7c10f6a/packages/zod-openapi/src/index.ts#L565-L566) does not throw any exceptions it just returns the error and the following `return` statement is just returning nothing, causing the Hono server to throw the error above.

I tampered with the code and I found that returning the validator response seems to achieve the expected result 

```ts
return await validator(c, next);
```

with this change it returns 400 when the validation doesn't pass and it returns the parsed data from the validator to route handler correctly.

happy to open a PR if needed 

Thanks

mouhannad-sh:
…-form-urlencoded submissions

Fixes #1719

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/moody-moose-draw.md
+++ b/.changeset/moody-moose-draw.md
@@ -1,5 +1,5 @@
 ---
-'@hono/zod-openapi': major
+'@hono/zod-openapi': patch
 ---
 
 fix Incomplete context internal error with x-www-form-urlencoded submissions when the payload is invalid


.changeset/moody-moose-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/zod-openapi': major
```
yusukebe: This is a `fix` change. So, can you replace `major` with `patch`?

yusukebe:
@mouhannad-sh 

Thank you so much! I'll merge and release a new version.

PR #1743, opened 2026-02-12
fix(auth-js): handle absolute URL in basePath

### LINKED ISSUE #1104: [@hono/auth-js] basePath can not be set to an absolute path

@hono/auth-js

1.0.5

4.7.5

bun 1.2.9

The backend is running on a different port / url. I attempt to set a new base path via the authConfigManager

````typescript
import { authConfigManager, SessionProvider } from '@hono/auth-js/react';

authConfigManager.setConfig({
  basePath: 'http://localhost:8000/api/auth', // if auth route is diff from /api/auth
});
````

And use a simple sign up method

````typescript
"use client"

import { cn } from '@/lib/utils';
import { Button } from '@/components/ui/button';
import { Input } from '@/components/ui/input';
import { Label } from '@/components/ui/label';
import { signIn } from '@hono/auth-js/react';

export function LoginForm({
  className,
  ...props
}: React.ComponentProps<'form'>) {
  // const providers = await getProviders();

  // console.log('Providers', providers);

  return (
    <form
      action={() => {
        // 'use server';
        signIn('github');
      }}
    >
      <button type="submit">Sign In</button>
    </form>
  );
}


A request is made against `http://localhost:8000/api/auth/session`

### What do you see instead?

A request is attempted against `http://localhost:3000http://localhost:8000/api/auth/session`

### Additional information

Additional question is if CORS would work correctly in this case and if the cookies would be set for the correct domains.

### PR DESCRIPTION
**Pull request by te2wow**
Fixes #1104

When `basePath` is set to an absolute URL like `http://localhost:8000/api/auth`, the URL is built by simple string concatenation (`${baseUrl}${basePath}/${path}`), resulting in a broken URL like `http://localhost:3000http://localhost:8000/api/auth/session`. This PR adds `normalizeBasePath()` in `client.ts` that splits an absolute URL in `basePath` into `baseUrl` and `basePath`, so that the URL concatenation produces the correct result.

### Changes
- `client.ts`: Add `normalizeBasePath()` function
- `react.tsx`: Use `normalizeBasePath()` in `AuthConfigManager.setConfig()`
- `client.test.ts`: Add tests for `normalizeBasePath()`

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/auth-js/src/client.ts
+++ b/packages/auth-js/src/client.ts
@@ -173,7 +173,7 @@ interface ParsedUrl {
   toString: () => string
 }
 
-export function normalizeBasePath<T extends { basePath?: string }>(config: T): T {
+export function normalizeBasePath(config: Partial<AuthClientConfig>): Partial<AuthClientConfig> {
   if (config.basePath && /^https?:\/\//.test(config.basePath)) {
     const url = new URL(config.basePath)
     return {

### CODE REVIEW

File: packages/auth-js/src/client.ts [hunk 1]
```diff
@@ -173,6 +173,18 @@ interface ParsedUrl {
   toString: () => string
 }
 
+export function normalizeBasePath<T extends { basePath?: string }>(config: T): T {
```
yusukebe: `Partial<AuthClientConfig>` is better for the `config` and the return value.

```ts
export function normalizeBasePath(config: Partial<AuthClientConfig>): Partial<AuthClientConfig> {
```

The current `T extends { basePath?: string }` is wrong if the return type includes `baseUrl`.

yusukebe:
Hey @te2wow 

Thank you for your first PR. I've left a comment. Please check it.

packages/auth-js/src/client.ts
te2wow: Thank you for the review! Updated to use `Partial<AuthClientConfig>`.

te2wow:
Thank you for the review! I've pushed the fix.

yusukebe:
@te2wow 

Thanks! Looks good. I'll merge and release a new version.

PR #1760, opened 2026-02-18
fix(mcp): remove trailing slash from issuer URL in OAuth metadata

### LINKED ISSUE #1753: createOAuthMetadata forces trailing slash in issuer (cannot   represent exact external issuer in authorization_servers)

@hono/mcp

0.2.3

4.11.9

node.js

## Summary

In `packages/mcp/src/auth/router.ts`, `createOAuthMetadata` currently takes `issuerUrl: URL` and emits `issuer: issuer.href`.

Because `URL#href` normalizes values, an issuer like `https://example.com` becomes `https://example.com/` (trailing slash added).
This is fine for built-in AS usage, but it prevents exact issuer representation when integrating with an external authorization server whose issuer must be matched as `https://example.com` (without trailing slash).

## Where it happens

- [packages/mcp/src/auth/router.ts:94](https://github.com/honojs/middleware/blob/66a9ece298593a443d1a50627bea2423a057d27a/packages/mcp/src/auth/router.ts#L74C1-L120C2)

## Current impact / Why this matters

In many current clients, this is not a major issue because they normalize or strip a trailing slash before comparing.

However, there are still strong reasons to fix this:

1. **Spec compliance and interoperability**
   - RFC 8414 requires strict identity/equality behavior for issuer metadata validation.
   - OIDC Discovery also relies on exact issuer matching semantics.
   - Relying on client-side normalization is implementation-dependent and not guaranteed across clients.

2. **Future compatibility risk**
   - A stricter client implementation (or future version) may reject `https://example.com/` when `https://example.com` is expected.

3. **Library API expressiveness**
   - Users cannot represent an externally defined issuer exactly if it does not include a trailing slash.

## Suggested fix

Change `issuerUrl` from `URL` to `string` in `createOAuthMetadata` (and related option types), so the exact issuer value can be preserved without URL normalization.

## Breaking change note

This is a breaking API change (`URL` -> `string`).

Before implementing, I want to confirm maintainers are okay with this direction.

Would you accept a PR with this breaking change?

## References

- RFC 8414: <https://www.rfc-editor.org/rfc/rfc8414>
- OpenID Connect Discovery 1.0: <https://openid.net/specs/openid-connect-discovery-1_0.html>

- If a user configures issuer as `https://example.com` (no trailing slash), metadata should return exactly `https://example.com`.
- `@hono/mcp` should not rewrite/normalize issuer formatting in a way that changes its string value.
- `authorization_servers[].issuer` should be representable exactly as defined by an external authorization server, including slash/no-slash differences.

### What do you see instead?

- `createOAuthMetadata` accepts `issuerUrl: URL` and emits `issuer: issuer.href`.
- JavaScript `URL` normalization adds a trailing slash for origin-only values.
- So input intended as `https://example.com` is output as `https://example.com/`.
- As a result, exact issuer strings without trailing slash cannot be represented in metadata.

### Additional information

_No response_

MathurAditya724:
Closes #1753


yusukebe:
@MathurAditya724 If it's ready to review, please ping me!

MathurAditya724:
it's ready for review @yusukebe

yusukebe:
@MathurAditya724 Thanks!

PR #1764, opened 2026-02-24
feat(universal-cache): add universal cache middleware package

lord007tn:
## Summary
This PR introduces `@hono/universal-cache` as a new third-party middleware package for Hono.

It follows the direction discussed in `honojs/hono#3857`:
https://github.com/honojs/hono/issues/3857

This middleware has been running internally in our production systems for a while. After validating it under real traffic and edge cases, we felt it was in a good place to open-source and get broader feedback on it.

## Dependency Model
`@hono/universal-cache` depends on [`unstorage`](https://unstorage.unjs.io/) as the storage abstraction.

That gives us:
- in-memory cache by default
- adapter-based backends through unstorage drivers (Redis, KV, filesystem, and others)
- one consistent cache API across runtimes

## Major Features
- response caching for Hono handlers via `cacheMiddleware()`
- function-level caching via `cacheFunction()`
- request-scoped defaults via `cacheDefaults()`
- stale-while-revalidate (SWR)
- in-flight deduplication for refresh paths
- configurable cache keys (`getKey`) and integrity (`integrity` / `hash`)
- custom `serialize` / `deserialize` hooks for response and function entries
- cache control hooks: `shouldBypassCache`, `shouldInvalidateCache`, `shouldRevalidate`, `validate`
- configurable HTTP methods and `varies` support for cache key variation
- opt-in manual revalidation via `revalidateHeader`
- `keepPreviousOn5xx` behavior for safer refresh failures
- global helpers for cache defaults and storage instance management

## Runtime Notes
- default `maxAge` is `60`
- manual revalidation is disabled by default
- on `workerd`, stale middleware entries refresh synchronously instead of relying on background self-fetch

## Test Coverage
Added coverage for:
- middleware caching behavior across methods
- bypass / invalidation / manual revalidation flows
- `shouldRevalidate` gating
- SWR behavior in standard runtimes and `workerd`
- vary-based keying
- custom serialization / deserialization
- function cache deduplication and validation
- global defaults / storage helpers

## Validation
- `corepack yarn eslint packages/universal-cache/src packages/universal-cache/vitest.config.ts packages/universal-cache/vitest.workerd.config.ts --cache --cache-location .cache/.eslintcache`
- `corepack yarn workspace @hono/universal-cache typecheck`
- `corepack yarn workspace @hono/universal-cache test --run`
- `corepack yarn workspace @hono/universal-cache test:workerd --run`
- `cd packages/universal-cache && bunx vitest run src/index.test.ts src/utils.test.ts`

## Related
- Docs PR: https://github.com/honojs/website/pull/825

## Merge Note
Please merge this together with the docs PR above.

- [x] Add tests
- [x] Run tests


lord007tn:
Addressed the Codecov coverage gaps with additional targeted tests.

### What I added
- New utility test suite:
  - `packages/universal-cache/src/utils.test.ts`
  - Covers all `utils.ts` branches and lines (`normalizePathToName`, `stableStringify`, TTL logic, expiry helpers).
- Expanded `index.test.ts` with cache edge-case scenarios:
  - malformed path decoding fallback key path
  - cache entry rejection on integrity mismatch
  - cache entry rejection via `validate()`
  - rejection of cached headers containing `etag: "undefined"` / `last-modified: "undefined"`
  - invalidation behavior for non-cacheable `no-store` responses
  - function cache default key hashing path (without `getKey`)
  - removal of malformed function cache entries

### Validation run
- `corepack yarn eslint packages/universal-cache/src --cache --cache-location .cache/.eslintcache`
- `corepack yarn workspace @hono/universal-cache typecheck`
- `corepack yarn workspace @hono/universal-cache test --run`

All passed locally.

The new commit is: `d79c864`

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,802 @@
...
+    return globalThis.btoa(binary)
+  }
+  const bufferCtor = getBufferCtor()
+  if (bufferCtor) {
+    return bufferCtor.from(bytes).toString('base64')
+  }
+  throw new Error('Base64 encoding is not available in this runtime')
+}
+
+const base64ToUint8Array = (base64: string) => {
```
yusukebe: How about using `decodeBase64` in `hono/utils/encode`?

https://github.com/honojs/hono/blob/b6f4bcdc2e991652cda78cb52626cb5c39406a82/src/utils/encode.ts#L25

Btw, you can use `atob` always. It's standard: https://min-common-api.proposal.wintertc.org/#api-interfaces

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,802 @@
...
+
+const getBufferCtor = (): BufferLike | null => {
+  const bufferCtor = (globalThis as unknown as { Buffer?: unknown }).Buffer
+  if (typeof bufferCtor === 'function' && 'from' in (bufferCtor as object)) {
+    return bufferCtor as unknown as BufferLike
+  }
+  return null
+}
+
+const arrayBufferToBase64 = (buffer: ArrayBuffer) => {
```
yusukebe: How about using `encodeBase64` in `hono/utils/encode`?

https://github.com/honojs/hono/blob/b6f4bcdc2e991652cda78cb52626cb5c39406a82/src/utils/encode.ts#L15

packages/universal-cache/src/utils.ts
```diff
@@ -0,0 +1,64 @@
+/** Default storage base prefix. */
+export const DEFAULT_CACHE_BASE = 'cache'
+/** Default storage group for cached handlers. */
+export const DEFAULT_HANDLER_GROUP = 'hono/handlers'
+/** Default storage group for cached functions. */
+export const DEFAULT_FUNCTION_GROUP = 'hono/functions'
+/** Default cache max age in seconds. */
+export const DEFAULT_MAX_AGE = 1
```
yusukebe: `1` is really proper? Isn't it too short for real users?

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,802 @@
...
+    if (maxAge <= 0) {
+      return next()
+    }
+
+    const bypass = await shouldBypassMiddlewareCache(ctx, merged)
+    if (bypass) {
+      return next()
+    }
+
+    const isRevalidateRequest = ctx.req.header(revalidateHeader) === '1'
```
yusukebe: Is it good design that any user agent can revalidate if it sends the request with the revalidate header like `x-cache-revalidate: 1`?

packages/universal-cache/src/cache.ts
yusukebe, line 1: I don't think SWR works on Workers. The behaviors of the following code on Bun and Workers are different:

```ts
import { Hono } from 'hono'
import { cacheMiddleware } from '@hono/universal-cache'

const app = new Hono()

let handlerCallCount = 0

app.get(
  '/api/data',
  cacheMiddleware({
    maxAge: 2,
    swr: true,
    staleMaxAge: 10,
  }),
  (c) => {
    handlerCallCount++
    console.log(`[handler] called #${handlerCallCount} at ${new Date().toISOString()}`)
    return c.json({
      count: handlerCallCount,
      time: Date.now(),
    })
  }
)

app.get('/api/stats', (c) => {
  return c.json({ handlerCallCount })
})

export default app
```

packages/universal-cache/src/utils.ts
lord007tn: By default, for our needs, we kind of bypass it by default. Can I change it to 60, maybe?

packages/universal-cache/src/cache.ts
lord007tn: currently we use a custom header ourselves, for that case 

Maybe I need to mention that in the docs, that its recommanded to change the header name 

or make the shouldRevalidate required so they can implement whatever gates they want here

packages/universal-cache/src/cache.ts
lord007tn: i will check that

packages/universal-cache/src/utils.ts
yusukebe: Yes. `60` and writing about it in the docs seem good.

packages/universal-cache/src/cache.ts
yusukebe: Adding `shouldRevalidate` and allowing users to define the logic is good for DX. And, I think "revalidate" should be optional. It makes safety-side.

lord007tn:
could you check it please
@yusukebe @usualoma

drag0n-app:
could you check it please
@yusukebe @usualoma

drag0n-app:
@yusukebe @usualoma I really need this feature. Could you please review it? Thanks!

yusukebe:
@lord007tn 

Sorry to be late. Can you merge the `main` branch into this branch to use the current stack (like `pnpm` instead of `yarn`) for `honojs/middleware`?

lord007tn:
@yusukebe Done, 
I also merged the new tweaks that we have internally.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+          const body =
+            method === 'GET' || method === 'HEAD'
+              ? undefined
+              : await ctx.req.raw.clone().arrayBuffer()
+          const request = new Request(ctx.req.url, {
+            ...(body ? { body } : {}),
+            method,
+            headers: refreshHeaders,
+          })
+          const response = await fetch(request)
```
yusukebe: The SWR self-fetch design looks fragile in real deployments:

1. It re-sends the user's authorization/cookie headers to the public URL, so credentials pass through proxies/CDNs again.
2. The internal token is per-process. With multiple instances (the main use case for Redis storage), the self-fetch hits another instance, the token doesn't match, and revalidation never completes.
3. If the server can't reach its own public URL, the fetch fails silently.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+    now,
+    serialize
+  )
+
+  if (getRuntimeKey() === 'workerd') {
+    ctx.executionCtx?.waitUntil?.(cachePromise)
+    return
+  }
+
+  await cachePromise
```
yusukebe: This blocks before returning the response on non-workerd runtimes. serialize calls `response.clone().arrayBuffer()`, which only resolves when the body stream closes. For an SSE / streaming response, the stream never closes until the client disconnects — but the client hasn't received anything yet, so the request hangs forever.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+let internalRevalidateToken: string | null | undefined
+
+const getInternalRevalidateToken = () => {
+  if (internalRevalidateToken === undefined) {
+    internalRevalidateToken = createInternalRevalidateToken()
+  }
+  return internalRevalidateToken
+}
+
+let defaultStorage: Storage = createStorage({
```
yusukebe, line 97: The default setup can grow memory without bound:
1. unstorage's memory driver ignores the ttl option, so `setItem(..., { ttl })` never expires anything here.
2. Expired entries are only removed when the same key is read again — never-read keys stay forever.
3. The default key includes the full cookie header, so with analytics cookies almost every visitor creates a new key that will never be read again.

So cacheMiddleware with defaults leaks memory in proportion to traffic, with a near-zero hit rate.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+/**
+ * Wrap a function with cache behavior.
+ * Provide `hash` in options to use WebCrypto or node:crypto for key hashing.
+ */
+export const cacheFunction = <TArgs extends unknown[], TResult>(
+  fn: (...args: TArgs) => Promise<TResult> | TResult,
+  options: CacheFunctionOptions<TArgs> | number = {}
+): ((...args: TArgs) => Promise<TResult>) => {
+  const normalized = typeof options === 'number' ? { maxAge: options } : options
+  const merged = { ...defaultCacheOptions, ...normalized }
```
yusukebe: cacheMiddleware reads `defaultCacheOptions` on every request, but `cacheFunction` copies it once here, at wrap time. Only storage is resolved at call time. So this ignores the defaults silently:

```ts
const getStats = cacheFunction(fetchStats) // module scope
setCacheDefaults({ maxAge: 300 })          // too late — ignored
```
Can you resolve all options at call time, like the middleware does? Also, setCacheDefaults merges and there is no way to reset.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+  }
+  const etag = response.headers.get('etag')
+  if (etag === 'undefined') {
+    return false
+  }
+  const lastModified = response.headers.get('last-modified')
+  if (lastModified === 'undefined') {
+    return false
+  }
+  if (response.headers.get('vary')?.trim() === '*') {
```
yusukebe: Only `Vary: *` is handled here. A response with `Vary: Accept-Language` is still cached under one key, unless the user also sets `varies: ['accept-language']` by hand. If they forget, users get responses cached for other users' headers.

An HTTP cache should read the response's Vary header and use it for the key. Or, to be safe: don't cache responses that have Vary when varies is not configured.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+    }
+
+    const maxAge = merged.maxAge ?? DEFAULT_MAX_AGE
+    const staleMaxAge = merged.staleMaxAge ?? DEFAULT_STALE_MAX_AGE
+    const swr = merged.swr ?? true
+    const keepPreviousOn5xx = merged.keepPreviousOn5xx ?? true
+    const base = merged.base ?? DEFAULT_CACHE_BASE
+    const group = merged.group ?? DEFAULT_HANDLER_GROUP
+    const methods = merged.methods?.map((method) => method.toUpperCase()) ?? ['GET', 'HEAD']
+    const hashFn = merged.hash ?? ((value: string) => ohash(value))
```
yusukebe, line 803: The default key separates users by `hash(authorization)` / `hash(cookie)`. With ohash, this is a short truncated hash — a collision here means one user gets another user's cached (possibly private) response. That makes hash collision a security issue, not just a cache miss.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+}
+
+const shouldInvalidateMiddlewareCache = async (ctx: Context, options: CacheMiddlewareOptions) => {
+  if (!options.shouldInvalidateCache) {
+    return false
+  }
+  return await options.shouldInvalidateCache(ctx)
+}
+
+const shouldManualRevalidateMiddlewareCache = async (
```
yusukebe, line 673: When `revalidateHeader` is set but `shouldRevalidate` is not, this allows everyone. Any client that knows the header name can skip the cache and hit the origin on every request — a cache-busting DoS.

The README warns about this, but I think the default should be safe: deny when s`houldRevalidate` is missing, or throw at setup time if revalidateHeader is set without it.

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+
+const isCacheableResponse = (response: Response) => {
+  if (response.status < 200 || response.status >= 300 || response.status === 206) {
+    return false
+  }
+  if (response.headers.has('set-cookie')) {
+    return false
+  }
+  const etag = response.headers.get('etag')
+  if (etag === 'undefined') {
```
yusukebe: Why these checks? A literal `"undefined"` header value only happens when upstream code stringifies undefined by mistake — that's an app bug, not something a generic cache middleware should special-case. And if we do, why only etag / last-modified, and why not `"null"` too?

packages/universal-cache/src/cache.ts
```diff
@@ -0,0 +1,855 @@
...
+ * Hono middleware that caches responses based on request data.
+ * Provide `hash` in options to use WebCrypto or node:crypto for key hashing.
+ */
+export const cacheMiddleware = (
+  options: CacheMiddlewareOptions | number = {}
+): MiddlewareHandler => {
+  const normalized: CacheMiddlewareOptions =
+    typeof options === 'number' ? { maxAge: options } : options
+  const { config: middlewareConfig, ...routeOptions } = normalized
+  const isConfigOnly = middlewareConfig !== undefined && Object.keys(routeOptions).length === 0
```
yusukebe: `cacheMiddleware({ config })` does the same thing as `cacheDefaults()` — two APIs for one feature. And the `isConfigOnly` switch is subtle: with only config this middleware is a pass-through, but adding one more option silently turns it into a caching middleware.

Also, config is written to the request-scoped defaults, so it affects other cache middleware later in the chain, not just this route.

packages/universal-cache/src/cache.ts
lord007tn: done, using Hono decodeBase64 now

packages/universal-cache/src/cache.ts
lord007tn: done, using Hono encodeBase64 now

packages/universal-cache/src/utils.ts
lord007tn: done, default is 60 and documented

packages/universal-cache/src/cache.ts
lord007tn: done, revalidation is optional and denied without shouldRevalidate

packages/universal-cache/src/cache.ts
lord007tn: removed response SWR, stale refresh is synchronous on all runtimes now

packages/universal-cache/src/cache.ts
lord007tn: removed the self-fetch

packages/universal-cache/src/cache.ts
lord007tn: done, SSE and common streaming types bypass caching

packages/universal-cache/src/cache.ts
lord007tn: done, default storage is bounded and TTL-aware; credentialed requests bypass by default

packages/universal-cache/src/cache.ts
lord007tn: done, function defaults resolve on every call and setCacheDefaults({}) resets

packages/universal-cache/src/cache.ts
lord007tn: done, responses with unkeyed Vary headers are not cached

packages/universal-cache/src/cache.ts
lord007tn: credentialed requests bypass by default now; ohash v2 also uses SHA-256

packages/universal-cache/src/cache.ts
lord007tn: done, denied when shouldRevalidate is missing

packages/universal-cache/src/cache.ts
lord007tn: removed

packages/universal-cache/src/cache.ts
lord007tn: removed config, cacheDefaults() is the scoped defaults API

lord007tn:
@yusukebe done, fixed the remaining edge cases and pushed. all checks pass locally.

packages/universal-cache/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "directory": "packages/universal-cache"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": ">=4.8.3"
+  },
+  "dependencies": {
+    "lru-cache": "^10.4.3",
+    "ohash": "^2.0.11",
+    "unstorage": "1.17.3"
```
yusukebe: What is this not  `^1.17.3` with a caret?

packages/universal-cache/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/universal-cache"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "peerDependencies": {
+    "hono": ">=4.8.3"
+  },
+  "dependencies": {
+    "lru-cache": "^10.4.3",
+    "ohash": "^2.0.11",
```
yusukebe: I think `ohash` is bundled in `dist`, but `ohash` is also in `dependencies`. So users install a copy that is never used. Should it move to devDependencies? (Or drop the inlining.)

packages/universal-cache/package.json
lord007tn: done, changed it to ^1.17.3

packages/universal-cache/package.json
lord007tn: done, moved ohash to devDependencies since it is bundled

lord007tn:
@yusukebe can you please check

yusukebe:
@lord007tn 

Thank you for the fixes!

I honestly am wondering whether we should merge this or not. The idea of "Universal Cache" is very good. But the codebase is too large for maintenance. More code means it's easier to have bugs. If so, the maintenance cost will increase. You may say "I'll handle them". BUT you may go away or throw this. Then who will resolve the issue? Who fixes the PR? Who reviews? And if you're still available, I still have to handle the security reports for this. So, it's not easy to merge.

drag0n-app:
This is a common feature in production environments, please merge it.

lord007tn:
> @lord007tn
> 
> Thank you for the fixes!
> 
> I honestly am wondering whether we should merge this or not. The idea of "Universal Cache" is very good. But the codebase is too large for maintenance. More code means it's easier to have bugs. If so, the maintenance cost will increase. You may say "I'll handle them". BUT you may go away or throw this. Then who will resolve the issue? Who fixes the PR? Who reviews? And if you're still available, I still have to handle the security reports for this. So, it's not easy to merge.

I understand that, 
Tbh, I said first, this is the same copy of a package that we use in production env currently that power multi million users on a daily basis, and we are maintaining it based on our needs, and scope, it may have other bugs, or a properly tested env, beyond bun, nodejs, it may also have missing features. But I found that this is useful for everyone who works in production.

Moreover, i dont know what I can do or say that will assure you, or everyone here, that I let you merge it. and make it sustainable, beyond "I will make this upstream to what we have now in production", so any bugs we face, issues we fix will be pushed here also, and anything that is flagged by other users will be pulled to our codebase and tested in our systems.

If you see/have another formula that might benefit everyone more here, I am open.

PR #1769, opened 2026-02-26
chore: autofix eslint prune suppressions

BarryThePenguin:
This should make autofix CI manage [`--prune-suppressions`](https://eslint.org/docs/latest/use/suppressions) 👀

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -34,7 +34,7 @@ jobs:
           node-version-file: .tool-versions
       - run: yarn install
       - run: yarn lint:fix --cache-strategy content --pass-on-unpruned-suppressions
-      - run: yarn lint:fix --cache-strategy content --prune-suppressions
+      - run: yarn lint --cache-strategy content --prune-suppressions
       - run: yarn format:fix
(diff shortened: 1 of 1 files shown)


.github/workflows/autofix.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           node-version-file: .tool-versions
       - run: yarn install
       - run: yarn lint:fix --cache-strategy content --pass-on-unpruned-suppressions
+      - run: yarn lint:fix --cache-strategy content --prune-suppressions
```
yusukebe: `yarn lint:fix` is written twice?

BarryThePenguin: The issue is that `--prune-suppressions` doesn't remove suppression if they have been fixed using the same command. So we can't just use `eslint --fix --prune-suppressions` as that potentially leaves unpruned suppressions, which is why the first command needs `--pass-on-unpruned-suppressions`.

The second line can probably just be

```suggestion
      - run: yarn lint --cache-strategy content --prune-suppressions
```

yusukebe:
@BarryThePenguin Thanks!

PR #1774, opened 2026-03-02
feat(mcp): relax Accept header validation with opt-in strict mode

### LINKED ISSUE #1773: `@hono/mcp`: Relax Accept header validation to accommodate common MCP clients

@hono/mcp

### Summary

The Streamable HTTP transport in `@hono/mcp` strictly requires POST requests to include both `application/json` and `text/event-stream` in the Accept header. While this matches the spec language, it breaks compatibility with several widely-used MCP clients that don't send both values.

### Current behavior

In `streamable-http.ts`, `handlePostRequest`:

```typescript
const acceptHeader = ctx.req.header('Accept')
if (
  !acceptHeader?.includes('application/json') ||
  !acceptHeader.includes('text/event-stream')
) {
  throw new HTTPException(406, { ... })
}
```

This rejects requests from:
- **Gemini CLI** (native HTTP transport) — sends only `Accept: application/json`
- **Java MCP SDK** (`io.modelcontextprotocol`) — sends only `Accept: text/event-stream`
- **Open WebUI** — missing `text/event-stream` in health check/validation requests
- **Standard HTTP clients** (curl, httpx, fetch) — send `*/*` or no Accept header
- **Dify MCP plugin** — same issue

### Impact

This is becoming a common pain point across the MCP ecosystem. The same strict check in the official TypeScript SDK and other implementations has generated issues in multiple projects:

- [open-webui/open-webui#19568](https://github.com/open-webui/open-webui/discussions/19568)
- [anomalyco/opencode#6972](https://github.com/anomalyco/opencode/issues/6972)
- [junjiem/dify-plugin-tools-mcp_sse#70](https://github.com/junjiem/dify-plugin-tools-mcp_sse/issues/70)
- [agentgateway/agentgateway#166](https://github.com/agentgateway/agentgateway/issues/166)
- [tadata-org/fastapi_mcp#233](https://github.com/tadata-org/fastapi_mcp/issues/233)

We ran into this ourselves building an MCP Code Mode surface on Cloudflare Workers using `@hono/mcp`. Our server works perfectly with spec-compliant clients (Claude Desktop, Cursor, etc.) but rejects everything else — including simple `curl` calls during development and testing.

### Proposed fix

Apply [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle): be liberal in what you accept. The server already knows whether it will respond with JSON or SSE, so it can make the right choice regardless of what the client advertises.

Something like:

```typescript
const acceptHeader = ctx.req.header('Accept') || '*/*'
const acceptsJson = acceptHeader.includes('application/json') || acceptHeader.includes('*/*')
const acceptsSse = acceptHeader.includes('text/event-stream') || acceptHeader.includes('*/*')

if (!acceptsJson && !acceptsSse) {
  // Reject only if the client explicitly accepts neither
  throw new HTTPException(406, { ... })
}
```

This would:
- ✅ Continue working with fully spec-compliant clients
- ✅ Accept `*/*` (standard default for curl, fetch, httpx)
- ✅ Accept `application/json` alone (Gemini CLI, standard JSON-RPC clients)
- ✅ Accept `text/event-stream` alone (Java MCP SDK)
- ✅ Accept missing Accept header (treat as `*/*` per HTTP semantics)
- ❌ Still reject explicitly incompatible Accept headers (e.g., `Accept: text/html`)

### Alternative: make it configurable

If maintaining strict spec compliance by default is important, a `strictAcceptHeader` option on `StreamableHTTPTransport` (defaulting to `false`) would let users opt into strict mode while keeping the default permissive.

### Environment

- `@hono/mcp` latest
- Deployed on Cloudflare Workers
- Tested against: curl, Gemini CLI, `@modelcontextprotocol/sdk` client, custom HTTP clients

bookernath:
## Summary

- Default to permissive `Accept` header validation in the Streamable HTTP transport — accept `application/json`, `text/event-stream`, or `*/*` individually (and treat missing `Accept` as `*/*`)
- Add `strictAcceptHeader` option to opt in to the original strict behavior (requiring both `application/json` and `text/event-stream`)
- Also accept `*/*` in the GET handler for consistency

## Motivation

The current strict requirement for both `application/json` **and** `text/event-stream` in the `Accept` header breaks compatibility with Gemini CLI, Java MCP SDK, Open WebUI, curl, and any standard HTTP client that doesn't send both. The server already knows whether it's responding with JSON or SSE — pre-rejecting based on `Accept` is unnecessarily strict.

This matches the approach taken in the upstream MCP TypeScript SDK ([modelcontextprotocol/typescript-sdk#551](https://github.com/modelcontextprotocol/typescript-sdk/pull/551)).

Closes #1773

## Test plan

- POST with `Accept: application/json` only → 200
- POST with `Accept: text/event-stream` only → 200
- POST with `Accept: */*` → 200
- POST with no `Accept` header → 200 (defaults to `*/*`)
- POST with `Accept: text/html` only → 406
- POST with `Accept: application/json, text/event-stream` → 200 (fully compliant)
- POST with `strictAcceptHeader: true` and `Accept: application/json` only → 406
- GET with `Accept: */*` → 200 (SSE stream)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/relax-accept-header.md
+++ b/.changeset/relax-accept-header.md
@@ -1,0 +1,5 @@
+---
+'@hono/mcp': patch
+---
+
+relax Accept header validation to accept `application/json`, `text/event-stream`, or `*/*` individually, improving compatibility with Gemini CLI, Java MCP SDK, Open WebUI, and curl. Add `strictAcceptHeader` option for strict MCP spec compliance.
--- a/.changeset/relax-accept-header.md
+++ b/.changeset/relax-accept-header.md
@@ -1,5 +1,5 @@
 ---
-'@hono/mcp': patch
+'@hono/mcp': minor
 ---
 
 relax Accept header validation to accept `application/json`, `text/event-stream`, or `*/*` individually, improving compatibility with Gemini CLI, Java MCP SDK, Open WebUI, and curl. Add `strictAcceptHeader` option for strict MCP spec compliance.
--- a/packages/mcp/src/streamable-http.ts
+++ b/packages/mcp/src/streamable-http.ts
@@ -54,13 +54,24 @@ export class StreamableHTTPTransport implements Transport {
   #allowedHosts?: string[]
   #allowedOrigins?: string[]
   #enableDnsRebindingProtection: boolean
+  #strictAcceptHeader: boolean
 
   sessionId?: string
   onclose?: () => void
   onerror?: (error: Error) => void
   onmessage?: (message: JSONRPCMessage, extra?: MessageExtraInfo) => void
 
-  constructor(options?: StreamableHTTPServerTransportOptions) {
+  constructor(
+    options?: StreamableHTTPServerTransportOptions & {
+      /**
+       * When `true`, strictly requires both `application/json` and `text/event-stream`
+       * in the `Accept` header per the MCP spec. When `false` (the default), accepts
+       * any of `application/json`, `text/event-stream`, or `*\/\*`, improving
+       * compatibility with clients like Gemini CLI, Java MCP SDK, curl, and Open WebUI.
+       */
+      strictAcceptHeader?: boolean
+    }
+  ) {
     this.#sessionIdGenerator = options?.sessionIdGenerator
     this.#enableJsonResponse = options?.enableJsonResponse ?? false
     this.#eventStore = options?.eventStore
@@ -69,6 +80,7 @@ export class StreamableHTTPTransport implements Transport {
     this.#allowedHosts = options?.allowedHosts
     this.#allowedOrigins = options?.allowedOrigins
     this.#enableDnsRebindingProtection = options?.enableDnsRebindingProtection ?? false
+    this.#strictAcceptHeader = options?.strictAcceptHeader ?? false
   }
 
   /**
@@ -148,8 +160,8 @@ export class StreamableHTTPTransport implements Transport {
   private async handleGetRequest(ctx: Context) {
     try {
       // The client MUST include an Accept header, listing text/event-stream as a supported content type.
-      const acceptHeader = ctx.req.header('Accept')
-      if (!acceptHeader?.includes('text/event-stream')) {
+      const getAcceptHeader = ctx.req.header('Accept') ?? '*/*'
+      if (!getAcceptHeader.includes('text/event-stream') && !getAcceptHeader.includes('*/*')) {
         throw new HTTPException(406, {
           res: Response.json({
             jsonrpc: '2.0',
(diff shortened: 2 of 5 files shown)


bookernath:
Hey @yusukebe @BarryThePenguin @MathurAditya724 👋

Would appreciate your eyes on this. We hit this while building an MCP server on Cloudflare Workers with @hono/mcp — the strict Accept header check breaks curl, Gemini CLI, the Java MCP SDK, and basically any client that isn't using the official TypeScript SDK.

Companion issue at #1773 with the full context. This is the same problem showing up across the MCP ecosystem (Open WebUI, Dify, Agent Gateway, fastapi_mcp have all filed similar issues against their servers).

The fix makes the default permissive (accept */*, application/json alone, or missing header) while adding a strictAcceptHeader option for anyone who needs spec compliance.

Happy to adjust anything, thanks!

yusukebe:
@bookernath Thank you for the PR.

Hey @MathurAditya724 Can you review this?

packages/mcp/src/streamable-http.ts
```diff
@@ -54,13 +54,16 @@ export class StreamableHTTPTransport implements Transport {
...
+  #strictAcceptHeader: boolean
 
   sessionId?: string
   onclose?: () => void
   onerror?: (error: Error) => void
   onmessage?: (message: JSONRPCMessage, extra?: MessageExtraInfo) => void
 
-  constructor(options?: StreamableHTTPServerTransportOptions) {
+  constructor(
+    options?: StreamableHTTPServerTransportOptions & { strictAcceptHeader?: boolean }
```
MathurAditya724: Can we add some JSDocs around this new prop

packages/mcp/src/streamable-http.ts
```diff
@@ -275,23 +279,46 @@ export class StreamableHTTPTransport implements Transport {
...
-              code: ErrorCode.ConnectionClosed,
-              message:
-                'Not Acceptable: Client must accept both application/json and text/event-stream',
-            },
-            id: null,
-          }),
-        })
+      const acceptHeader = ctx.req.header('Accept') || '*/*'
+
+      if (this.#strictAcceptHeader) {
```
MathurAditya724: This whole block can be like -
```ts
const isAcceptable = this.#strictAcceptHeader
  ? acceptHeader.includes('application/json') && acceptHeader.includes('text/event-stream')
  : acceptHeader.includes('application/json') ||
    acceptHeader.includes('text/event-stream') ||
    acceptHeader.includes('*/*')

if (!isAcceptable) {}
```
Just to keep things clean

MathurAditya724 (commented):
Sorry for the delay on this. 2 small changes but the rest looks good.

Thank you @bookernath for the PR

MathurAditya724:
Also, can you resolve the CI issues?

bookernath:
@MathurAditya724 thanks for your review! I've addressed the things you mentioned, can you take a fresh look?

bookernath:
@MathurAditya724 just following up 🫡

MathurAditya724 (approved):
changes look good, both items addressed — thanks for the quick turnaround

.changeset/relax-accept-header.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/mcp': patch
```
yusukebe: Isn't this `minor` instead of `patch`?

yusukebe (approved):
LGTM!

yusukebe:
@bookernath @MathurAditya724 Thank you both! Let's go.

PR #1782, opened 2026-03-03
feat(structured-logger): add @hono/structured-logger middleware

gabry-ts:
## Summary

Adds a new structured logging middleware for Hono that provides a request scoped logger on c.var.logger.

This middleware is library agnostic and works with pino, winston, bunyan, console, or any logger implementing the BaseLogger interface. It has zero dependencies and is compatible with all runtimes supported by Hono.

Core features:
  * Request scoped logger instance via c.var.logger (configurable key)
  * Automatic response time measurement using performance.now()
  * Native integration with hono/request-id
  * Customizable onRequest, onResponse, and onError hooks
  * Full type safety with generics
  * Works on Node.js, Deno, Bun, Cloudflare Workers, AWS Lambda, Vercel Edge, Fastly Compute

Related: honojs/hono#3963

## Test plan

  * 20 unit tests covering all options, hooks, error handling, custom context keys, and default behaviors
  * TypeScript compiles with no errors


packages/structured-logger/README.md
```diff
@@ -0,0 +1,201 @@
...
+Returns a Hono `MiddlewareHandler`.
+
+#### Options
+
+| Option | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `createLogger` | `(c: Context) => L` | Yes | | Factory that creates a request scoped logger instance |
+| `contextKey` | `string` | No | `'logger'` | Key used to store the logger on `c.var` |
+| `onRequest` | `(logger: L, c: Context) => void \| Promise<void>` | No | Logs method + path at info level | Called before handler execution |
+| `onResponse` | `(logger: L, c: Context, elapsedMs: number) => void \| Promise<void>` | No | Logs status + elapsed at info level | Called after handler execution |
```
jansepke: `elapsed time` word missing

packages/structured-logger/README.md
```diff
@@ -0,0 +1,201 @@
...
+### `structuredLogger(options)`
+
+Returns a Hono `MiddlewareHandler`.
+
+#### Options
+
+| Option | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `createLogger` | `(c: Context) => L` | Yes | | Factory that creates a request scoped logger instance |
+| `contextKey` | `string` | No | `'logger'` | Key used to store the logger on `c.var` |
```
jansepke: would the API maybe be simpler if you merge `contextKey` option into `createLogger`? like:
`createLogger: (c) => ({ logger: rootLogger.child({ requestId: c.var.requestId }) })`

could maybe be more flexible, but also more verbose. But I think I prefer your approach...

packages/structured-logger/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+function defaultOnRequest<L extends BaseLogger>(logger: L, c: Context): void {
+  logger.info({ method: c.req.method, path: c.req.path }, 'request start')
+}
+
+function defaultOnResponse<L extends BaseLogger>(logger: L, c: Context, elapsedMs: number): void {
+  logger.info({ status: c.res.status, elapsedMs }, 'request end')
+}
+
+function defaultOnError<L extends BaseLogger>(logger: L, err: Error, c: Context): void {
+  logger.error({ err, method: c.req.method, path: c.req.path }, 'request error')
```
jansepke: maybe also log the status code here as it could be modified by hono's onError handler

packages/structured-logger/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const now =
+  typeof performance !== 'undefined' ? () => performance.now() : () => Date.now()
+
+function defaultOnRequest<L extends BaseLogger>(logger: L, c: Context): void {
+  logger.info({ method: c.req.method, path: c.req.path }, 'request start')
+}
+
+function defaultOnResponse<L extends BaseLogger>(logger: L, c: Context, elapsedMs: number): void {
+  logger.info({ status: c.res.status, elapsedMs }, 'request end')
```
jansepke: I would prefer to include method+path also here, so one does not need to always combine this log line with the 'request start' log line to do analytics. This would also more align `defaultOnResponse` and `defaultOnError`. But it could also be a personal preference and likely everyone will modify those settings for a production setup.

gabry-ts:
Thanks for the review.

* Fixed "elapsed time" wording in the README
* Keeping contextKey as a separate option since you also prefer that approach
* Added status to defaultOnError
* Added method and path to defaultOnResponse

Let me know if you have any other feedback.

jansepke (approved):

packages/structured-logger/deno.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@hono/structured-logger",
+  "version": "0.1.0",
```
yusukebe: Set `0.0.0`. GitHub Actions will automatically set the correct version.

packages/structured-logger/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@hono/structured-logger",
+  "version": "0.1.0",
```
yusukebe: Set `0.0.0`. GitHub Actions will automatically set the correct version.

packages/structured-logger/README.md
yusukebe, line 1: How about adding the introduction for adding type support for `c.var.logger`?

Like this:

```ts
import { Hono } from 'hono'
import { structuredLogger } from '@hono/structured-logger'

import pino from 'pino'

const rootLogger = pino()

const app = new Hono<{
  Variables: {
    logger: ReturnType<typeof rootLogger.child>
  }
}>()

app.use(
  structuredLogger({
    createLogger: (c) => rootLogger.child({ foo: 'bar' })
  })
)

app.get('/', (c) => {
  c.var.logger.info('hello') // typed!
  return c.json({ foo: 'bar' })
})
```

yusukebe: @gabry-ts Ahh, sorry! I missed the `Type safe context` section you wrote. Passing `pino.Logger` is better than `ReturnType<typeof rootLogger.child>`. My request is not necessary. Can you revert the change?

gabry-ts:
@yusukebe Sorry for the late reply, I missed the notification.
Reverted the README change as requested in b7e3161.

yusukebe (approved):
LGTM!

yusukebe:
@gabry-ts 

Great! Let's go with this. I'll merge and release an initial version. Thank you!

Issue #1759, opened 2026-02-17
OpenTelemetry request duration reporting milliseconds instead of seconds


ajkachnic:
### Which middleware has the bug?

@hono/otel

### What version of the middleware?

1.0.1

### What version of Hono are you using?

4.11.4

### What runtime/platform is your app running on? (with version if possible)

Node.js, Vercel

### What steps can reproduce the bug?

When using the `@hono/otel` middleware, request duration is reported as milliseconds, rather than seconds, contrary to the declared unit of the metric (seconds).

https://github.com/honojs/middleware/blob/66a9ece298593a443d1a50627bea2423a057d27a/packages/otel/src/trackers.ts#L14-L27

In the code, `performance.now()` is used to calculate the request duration:
https://github.com/honojs/middleware/blob/66a9ece298593a443d1a50627bea2423a057d27a/packages/otel/src/index.ts#L123-L126

The delta between the start time and the current time (both measured using `performance.now()`) are ingested into OpenTelemetry's metrics SDK, with no scaling. The [`performance.now()` API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) measures time in milliseconds

> The performance.now() method returns a high resolution timestamp in milliseconds

This causes issues in downstream metrics systems (e.g. Prometheus) as the ingested data is 1000x its expected values.

I am happy to open up a PR to fix, but wanted to report the issue and confirm its legitimacy first.

I looked through the commit history, and it appears to be a regression from the refactor in https://github.com/honojs/middleware/pull/1513. Specifically, the prior code correctly divided by 1000, but the division was removed in the refactor towards semantic conventions:
https://github.com/honojs/middleware/blob/5e9f153e3b510978dd3266dee398559e594cfa35/packages/otel/src/metrics.ts#L40-L45
https://github.com/honojs/middleware/blob/fc4a1c5c44956ec81a9b67bd731a92dfaaf6d9b1/packages/otel/src/index.ts#L120-L125

yusukebe:
@ajkachnic Thank you for the issue. Sorry to be late.

Hey, @dahlia Can you take a look at this issue?

mrlorentx:
Hey @ajkachnic, thanks a bunch for raising this! I've had a look and you are completely correct - I created this regression with my refactor..

Dropping a fix and PR for this in just a little while!

yusukebe:
@mrlorentx Thanks!!

dahlia:
[...]

PR #1784, opened 2026-03-04
fix(otel): convert duration to seconds before recording

mrlorentx:
### Fixes duration recording unit (#1759)

- [ ] Add tests
- [X] Run tests
- [X] `yarn changeset` at the top of this repo and push the changeset
- [X] Follow [the contribution guide](https://github.com/honojs/middleware?tab=readme-ov-file#how-to-contribute)


yusukebe:
Hey @mrlorentx 

Thank you for the quick fix. If possible, can you add a proper test?

mrlorentx:
I've added a test which verifies the recording - but unless we want to hide this in a pure function that does the recording and sacrifice some readability I don't see how we can make it more precise than this. Any thoughts?

yusukebe:
@dahlia Any idea?

mrlorentx:
Maybe I am just tired today - I updated the test to be stricter and simply mock `performance.now()`. Should be good now.

yusukebe:
@mrlorentx Nice! Thanks!

PR #1799, opened 2026-03-17
fix(mcp): relax Accept header validation for broader client compatibility

### LINKED ISSUE #1773: `@hono/mcp`: Relax Accept header validation to accommodate common MCP clients

ghost:
## Summary

Relaxes the strict Accept header validation in the Streamable HTTP transport's `handlePostRequest` to accommodate MCP clients that don't send both `application/json` and `text/event-stream`.

**Before:** Requests must include both `application/json` AND `text/event-stream` in the Accept header, or get a 406.

**After:** Requests are accepted if they include `application/json` OR `text/event-stream` OR `*/*`. Missing Accept header defaults to `*/*`. Only rejects when the client explicitly accepts neither.

## Problem

The strict check breaks compatibility with several widely-used MCP clients:
- **Gemini CLI** — sends only `Accept: application/json`
- **Java MCP SDK** — sends only `Accept: text/event-stream`
- **Open WebUI** — missing `text/event-stream` in validation requests
- **curl, httpx, fetch** — send `*/*` or no Accept header
- **Dify MCP plugin** — same issue

This is a known pain point across the MCP ecosystem (see linked issues in #1773).

## Changes

- `packages/mcp/src/streamable-http.ts`: Updated Accept header validation to use Postel's Law (be liberal in what you accept)
- `packages/mcp/src/streamable-http.test.ts`: Updated existing test + added 4 new tests for `application/json` only, `text/event-stream` only, `*/*`, and missing Accept header

## Test plan

- [x] Existing test updated to match new behavior
- [x] New test: POST with only `application/json` Accept — accepted
- [x] New test: POST with only `text/event-stream` Accept — accepted
- [x] New test: POST with `*/*` Accept — accepted
- [x] New test: POST with no Accept header — accepted
- [x] New test: POST with `text/plain` Accept — rejected with 406

Fixes #1773

### DIFF
--- a/.changeset/relax-accept-header.md
+++ b/.changeset/relax-accept-header.md
@@ -1,0 +1,5 @@
+---
+"@hono/mcp": patch
+---
+
+Relax Accept header validation in MCP Streamable HTTP transport for broader client compatibility


yusukebe:
@MathurAditya724 Can you also take a look at this?

PR #1804, opened 2026-03-17
fix(mcp): call onclose when standalone SSE stream aborts on client disconnect

### LINKED ISSUE #1762: StreamableHTTPTransport: SSE stream abort should notify via onclose (or new callback)
Human note: Cursor generated this report for me, since it helped me give perfect specifics, but it's all from me and human-edited.

The impact is that we can't actually tell if an agent is still there when using http-transport, since the SSE stream closing is hidden from our app-code, and agents do not send DELETE requests when they poof.

---

When a client opens a GET SSE stream and then disconnects (e.g. the agent process exits, network drops, etc.), the transport's internal `stream.onAbort` handler fires and cleans up `#streamMapping`, but the disconnection is never propagated to the `onclose` callback or any other external hook.

**Current behavior**
```js
// Inside handleGetRequest — stream abort only cleans up internal state
stream.onAbort(() => {
this.#streamMapping.get(resolvedStreamId)?.cleanup();
});
```

`onclose` is only called inside `close()`, which is only triggered by an explicit DELETE request. In practice, MCP clients (Cursor, Claude Code, etc.) do not send DELETE when they disconnect — the SSE stream dropping is the only signal available.

**Expected behavior:**

When the standalone SSE stream (`_GET_stream`) aborts, the transport should notify the application. This could be:

1. Calling `onclose` when the standalone SSE stream aborts (since it represents the client's only persistent connection), or
2. A new callback like `onsessiondisconnected` that fires specifically for SSE stream drops, keeping `onclose` reserved for explicit termination.

Option 1 seems simplest and matches the semantic intent of `onclose` — the client is gone.

ghost:
## Summary

Calls `onclose` when the standalone GET SSE stream aborts (client disconnect), so applications can detect when an MCP client is gone.

## Problem

When a client opens a GET SSE stream and then disconnects (process exit, network drop), the transport's `stream.onAbort` handler fires and cleans up `#streamMapping`, but the disconnection is never propagated to the `onclose` callback.

`onclose` is only called inside `close()`, which is triggered by an explicit DELETE request. In practice, MCP clients (Cursor, Claude Code, etc.) do not send DELETE when they disconnect. The SSE stream dropping is the only signal available.

This means applications have no way to know when a client has gone away.

## Fix

When the standalone GET SSE stream (identified by `_GET_stream` ID) aborts, call `this.onclose?.()` to notify the application. This only fires for the standalone stream, not for per-request POST SSE streams.

## Changes

- `packages/mcp/src/streamable-http.ts`: Added `onclose` call in the standalone SSE stream's `onAbort` handler

Fixes #1762

### DIFF
--- a/.changeset/sse-onclose-abort.md
+++ b/.changeset/sse-onclose-abort.md
@@ -1,0 +1,5 @@
+---
+"@hono/mcp": patch
+---
+
+Call onclose when standalone SSE stream aborts on client disconnect


yusukebe:
@ctonneslan Thank you for the PR.

Hey @MathurAditya724 , can you handle this?

PR #1824, opened 2026-03-25
fix(mcp): allow loopback redirect_uri with any port per RFC 8252

### LINKED ISSUE #1823: fix(mcp): authorize handler rejects localhost redirect_uri with different port (RFC 8252)
**Which middleware has the bug?**
`@hono/mcp`

**What version of the middleware?**
0.2.4

**What version of Hono are you using?**
4.7.x

**What runtime/platform is your app running on?**
Cloudflare Workers

**What steps can reproduce the bug?**

1. Register an OAuth client with `redirect_uris: ["http://localhost/callback"]`
2. Send an authorize request with `redirect_uri=http://localhost:58621/callback` (any port)
3. Get back `{"error":"invalid_request","error_description":"Unregistered redirect_uri"}`

This happens because `authorizeHandler` uses `client.redirect_uris.includes(redirect_uri)` which is a strict string comparison — `localhost:58621` ≠ `localhost`.

**What is the expected behavior?**

Per [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3), loopback redirect URIs should match on scheme + host + path, ignoring port. Native OAuth clients (Claude Code, VS Code, etc.) bind an ephemeral port each time they start an auth flow, so the port will never match a pre-registered URI.

**What do you see instead?**

400 with `Unregistered redirect_uri`. This blocks any native MCP client from completing OAuth through a server that uses `@hono/mcp`'s authorize handler.

**Additional information**

We hit this in production at [Smithery](https://smithery.ai) — Claude Code couldn't connect to any OAuth-requiring MCP server (Linear, GitHub, etc.). We worked around it by monkey-patching `getClient()` to expand `redirect_uris` for loopback addresses before passing to `authorizeHandler`.

PR with a fix + tests: #1822
Related: https://github.com/smithery-ai/cli/issues/708

arjunkmrm:
We ran into this while running [Smithery](https://smithery.ai) as an MCP OAuth proxy using `@hono/mcp`. Claude Code registers `http://localhost/callback` in its client metadata but connects with an ephemeral port at runtime (e.g. `http://localhost:58621/callback`). The authorize handler rejects it because `redirect_uris.includes()` does a strict string comparison.

[RFC 8252 Section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) says loopback redirect URIs should match on scheme + host + path only, ignoring the port — native OAuth clients pick a random available port each time. Here's the excerpt:
```
The authorization server MUST allow any port to be specified at the
  time of the request for loopback IP redirect URIs, to accommodate
  clients that obtain an available ephemeral port from the operating
  system at the time of the request.
```

This adds a small `isLoopbackRedirectAllowed()` helper that kicks in only for `localhost`, `127.0.0.1`, and `[::1]`. Everything else still uses strict matching. Four new tests cover the happy path, path mismatch rejection, and non-loopback strictness.

Fixes #1823

- [x] Add tests
- [ ] Run tests

> Note: haven't run `yarn changeset` yet — happy to add one if this approach looks right.

### DIFF
--- a/packages/mcp/src/auth/helpers/authorize.ts
+++ b/packages/mcp/src/auth/helpers/authorize.ts
@@ -53,7 +53,10 @@ function isLoopbackRedirectAllowed(redirectUri: string, registeredUris: string[]
     try {
       const reg = new URL(registered)
       return (
-        reg.hostname === req.hostname && reg.protocol === req.protocol && reg.pathname === req.pathname && reg.search === req.search
+        reg.hostname === req.hostname &&
+        reg.protocol === req.protocol &&
+        reg.pathname === req.pathname &&
+        reg.search === req.search
       )
     } catch {
       return false


yusukebe:
Hey @MathurAditya724 , Can you reivew this?

packages/mcp/src/auth/helpers/authorize.ts
```diff
@@ -30,6 +30,37 @@ const RequestAuthorizationParamsSchema = z.object({
...
+
+  if (!LOOPBACK_HOSTS.includes(req.hostname)) {
+    return false
+  }
+
+  return registeredUris.some((registered) => {
+    try {
+      const reg = new URL(registered)
+      return (
+        reg.hostname === req.hostname && reg.protocol === req.protocol && reg.pathname === req.pathname
```
MathurAditya724: can you also compare `req.search` here

arjunkmrm: done!

yusukebe:
@arjunkmrm @MathurAditya724 Thank you!

PR #1855, opened 2026-04-15
fix(oidc-auth): await revokeSession in getAuth to avoid unhandled rejection

### LINKED ISSUE #1851: Bug: `revokeSession()` called without `await` in `getAuth()` causes unhandled promise rejection

@hono/oidc-auth

1.8.1

4.12.8

AWS Lambda

1. Configure `@hono/oidc-auth` with an IdP that has a revocation endpoint. Let a session expire (`ssnexp` passes)
3. Make a request — `getAuth()` calls `revokeSession(c)` without await, returns null, request completes normally
4. The background `revokeSession` promise rejects (e.g. IdP returns `invalid_grant`)
5. Node.js crashes with unhandled promise rejection

When a session expires (ssnexp < now), `getAuth()` should cleanly return null so the middleware can redirect to login. Token revocation at the IdP should be best-effort and never crash the application

### What do you see instead?

`getAuth()` fires off `revokeSession(c)` without `await` and returns `null`. The request completes fine, but the background revocation promise rejects (e.g. IdP returns `invalid_grant` for an already-revoked token), causing an unhandled promise rejection that crashes the Node.js process. On AWS Lambda, this kills the execution environment and causes the *next* request to fail with `Runtime.ExitError`.

### Additional information

In `getAuth()`, when `ssnexp < now`, `revokeSession(c)` is called without `await`:

```javascript
if (auth.ssnexp < now) {
    revokeSession(c);  // ← not awaited
    return null;
}
```

`revokeSession` is async — it calls the IdP's revocation endpoint and throws an `HTTPException` if the revocation fails:

```javascript
if (oauth2.isOAuth2Error(result)) {
    throw new HTTPException(500, {
        message: `OAuth2Error: [${result.error}] ${result.error_description}`,
    });
}
```

Even if the `await` is added, `revokeSession` throws on revocation errors. Token revocation is best-effort. The server returning an error (e.g. token already revoked) shouldn't crash the application.

## Suggested fix

```diff
 if (auth.ssnexp < now) {
-    revokeSession(c);
+    revokeSession(c).catch(() => {});
     return null;
 }
```

Or better, don't throw on revocation errors at all:

```diff
 const result = await oauth2.processRevocationResponse(response)
-if (oauth2.isOAuth2Error(result)) {
-    throw new HTTPException(500, {
-        message: `OAuth2Error: [${result.error}] ${result.error_description}`,
-    })
-}
```

## Workaround

```typescript
process.on("unhandledRejection", (reason) => {
  const msg = reason instanceof Error ? reason.message : String(reason);
  if (msg.includes("OAuth2Error") || msg.includes("invalid_grant")) return;
  throw reason;
});
```

SAY-5:
Per #1851, when a session expires (`ssnexp < now`), `getAuth()` fired `revokeSession(c)` without awaiting it. The background promise could reject (e.g. the IdP returns `invalid_grant` for an already-revoked token), causing an unhandled promise rejection that crashes Node.js. On AWS Lambda this kills the execution environment and breaks the *next* request with `Runtime.ExitError`.

Await the revocation inside a try/catch so the IdP outcome is best-effort and a rejection can never escape `getAuth()`. Behaviour is unchanged on success — `getAuth()` still returns `null` after revocation so the middleware can redirect to login.

Closes #1851

### DIFF
--- a/.changeset/oidc-auth-await-revoke-session.md
+++ b/.changeset/oidc-auth-await-revoke-session.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': patch
+---
+
+Fix `getAuth()` swallowing `revokeSession()` failures: the call was invoked without `await` inside the refresh branch, so a rejecting `revokeSession` turned into an unhandled rejection and `getAuth()` returned `null` without surfacing the error. Await the call inside a try/catch so errors are observable and the session is cleaned up consistently.
--- a/packages/oidc-auth/eslint-suppressions.json
+++ b/packages/oidc-auth/eslint-suppressions.json
@@ -8,9 +8,6 @@
     }
   },
   "src/index.ts": {
-    "@typescript-eslint/no-floating-promises": {
-      "count": 1
-    },
     "@typescript-eslint/no-non-null-assertion": {
       "count": 1
     },
--- a/packages/oidc-auth/src/index.test.ts
+++ b/packages/oidc-auth/src/index.test.ts
@@ -249,6 +249,41 @@ describe('oidcAuthMiddleware()', () => {
     expect(res.headers.get('set-cookie')).toMatch('code_verifier=')
     expect(res.headers.get('set-cookie')).toMatch('continue=http%3A%2F%2Flocalhost%2F')
   })
+  test('Should swallow revocationRequest failure on expired session (no unhandled rejection)', async () => {
+    // Regression for #1851: when getAuth() detects an expired session it
+    // calls revokeSession() to delete the cookie. Before the fix, the
+    // call was fire-and-forget; if the IdP returned an error
+    // (e.g. invalid_grant on an already-revoked refresh token) the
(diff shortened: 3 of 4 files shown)


yusukebe:
Thanks @SAY-5 

Can you add a test?

SAY-5:
Hey @yusukebe, thanks — pushed a regression test in commit `37acd8a` (`test(oidc-auth): pin getAuth() expired-session revocation against unhandled rejection`).

It mocks `oauth4webapi.revocationRequest` to reject (mirroring the original `invalid_grant` scenario from #1851), sends a request with an expired-session cookie, and asserts:
- the redirect response still completes with status 302
- `revocationRequest` was actually invoked (proving the await is wired)
- no `unhandledRejection` fires on `process` during the request

`vitest run src/index.test.ts` → 37/37 passing.

yusukebe (approved):
LGTM!

yusukebe:
@SAY-5 Thanks.

PR #1869, opened 2026-04-28
feat(inertia): return props for JSON requests

adwd:
Adds a JSON response path to `@hono/inertia` so `c.render(component, props)` returns the rendered props directly when the request accepts `application/json`.

`X-Inertia` requests still return the Inertia page object, and normal initial page loads still return HTML. Responses now vary on both `Accept` and `X-Inertia` so caches can distinguish those response shapes.

Docs and tests are updated for the new response mode.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/giant-candies-work.md
+++ b/.changeset/giant-candies-work.md
@@ -1,0 +1,5 @@
+---
+'@hono/inertia': minor
+---
+
+feat(inertia): return props for JSON requests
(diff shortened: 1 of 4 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@adwd 

Thank you for the quick work! Looks good!

PR #1862, opened 2026-04-21
fix(node-ws): deprecated

yusukebe:
### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [x] `yarn changeset` at the top of this repo and push the changeset
- [x] Follow [the contribution guide](https://github.com/honojs/middleware?tab=readme-ov-file#how-to-contribute)

PR #1873, opened 2026-04-29
fix(node-ws): forward response headers during upgrade

gentamura:
## Summary

Forward Hono response headers to the WebSocket upgrade response so headers set by middleware (e.g. `Set-Cookie`, custom auth headers, `WWW-Authenticate` on reject) are no longer dropped during the handshake.

Previously, `injectWebSocket` discarded `response.headers` after `init.app.request(...)`, so any header attached by Hono middleware never reached the client — both on successful upgrades and on rejected upgrades (4xx/5xx).

## Changes

- On successful upgrade, response headers are appended via `wss.on('headers', ...)` (the official `ws` API for injecting handshake headers). The listener is removed in `finally`; `headers` is emitted synchronously inside `handleUpgrade`, so it cannot leak across concurrent upgrades.
- On rejected upgrade (no waiter / connectionSymbol mismatch), response headers are written into the manual `socket.end(...)` HTTP response.
- Skip hop-by-hop headers per [RFC 9110 §7.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-connection) (`connection`, `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailer`, `transfer-encoding`, `upgrade`), plus framing (`content-length`) and WebSocket handshake headers managed by `ws` (`sec-websocket-accept`, `sec-websocket-extensions`, `sec-websocket-protocol`) to avoid corrupting the handshake.

## Verification

- `yarn workspace @hono/node-ws typecheck`
- `yarn workspace @hono/node-ws test run` (15/15 passing, including 2 new tests)
- `yarn prettier --check packages/node-ws/src/index.ts packages/node-ws/src/index.test.ts .changeset/polite-cooks-report.md`

New tests cover:
- Custom response header set by middleware reaches the client on a successful upgrade.
- Custom response header is preserved on a rejected (401) upgrade response.

### DIFF
--- a/.changeset/polite-cooks-report.md
+++ b/.changeset/polite-cooks-report.md
@@ -2,4 +2,7 @@
 '@hono/node-ws': patch
 ---
 
-Forward Hono response headers to WebSocket upgrade responses.
+Fix: forward Hono response headers (e.g., `Set-Cookie`, custom auth headers)
+to WebSocket upgrade responses so headers set by middleware are not dropped
+during the handshake. Hop-by-hop headers (per [RFC 9110 Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-connection))
+and headers managed by `ws` are skipped to avoid corrupting the handshake.
--- a/.changeset/polite-cooks-report.md
+++ b/.changeset/polite-cooks-report.md
@@ -1,0 +1,5 @@
+---
+'@hono/node-ws': patch
+---
+
+Forward Hono response headers to WebSocket upgrade responses.
--- a/packages/node-ws/src/index.test.ts
+++ b/packages/node-ws/src/index.test.ts
@@ -270,6 +270,57 @@ describe('WebSocket helper', () => {
     expect(await mainPromise).toBe(true)
   })
 
+  it('Should include response headers in the WebSocket upgrade response', async () => {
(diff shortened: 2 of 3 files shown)


gentamura:
Hi @yusukebe — first time contributing here, thanks for creating Hono and maintaining this ecosystem!

Quick context check on this PR: I noticed `@hono/node-ws` is being deprecated in #1862 in favor of `@hono/node-server`'s built-in `upgradeWebSocket`. The same response-header drop bug exists in `honojs/node-server/src/websocket.ts` too — `rejectUpgradeRequest` writes only `Connection: close` / `Content-Length: 0`, and the success path doesn't hook `wss.on('headers', ...)`, so headers set by Hono middleware (auth, `Set-Cookie`, etc.) never reach the client during the handshake.

I've opened a companion PR with the same fix against `honojs/node-server`: honojs/node-server#346. Happy to land both, or close this one if you'd prefer to only fix forward in `node-server`. Let me know which direction works best for the deprecation timeline.

gentamura:
Thanks for the quick turnaround and for merging the node-server fix, @yusukebe! Closing this PR as agreed — the fix lives forward in `@hono/node-server` (honojs/node-server#346), which is the right home given the deprecation. Looking forward to contributing again.

PR #1881, opened 2026-05-05
feat(zod-validator): surface the default 400 on the no-hook overload and keep the return assignable to MiddlewareHandler

T4ko0522:
## Summary

- Declare the default `400` failure response (`c.json(result, 400)`) on the no-hook `zValidator(target, schema)` overload so it propagates to the RPC schema via Hono's `MergeMiddlewareResponse<M_k>`. Refs honojs/hono#3746.
- Intersect inferred middleware responses with `Response` so a `zValidator(...)` return remains assignable to a plain `MiddlewareHandler` (avoids `TS2322`).
- Collapse the previous two no-hook signatures into a single overload that also accepts `options.validationFunction`.

## Why

The Hono core type machinery introduced by honojs/hono#4393 (and extended by honojs/hono#4906 for `app.on`) already pulls a middleware's typed response into the RPC schema. But the no-hook overload returned `MiddlewareHandler<E, P, V>` (R defaults to `Response`), and `Response` collapses to `never` inside `ExtractTypedResponseOnly<...>`. As a result the `c.json(result, 400)` that the implementation runs on validation failure never reached `hc<typeof app>`, and users had to fall back to calling `c.status(400)` before `c.json(...)` — exactly what's reported in honojs/hono#3746.

## What changed

**`packages/zod-validator/src/index.ts`**

- Add `ZodValidatorFailureBody<T>` — the `success: false` variant of `safeParseAsync`'s return value (covers both Zod v3 `SafeParseError<input>` and Zod v4 `ZodSafeParseError<output>`).
- Widen the no-hook overload return type to `MiddlewareHandler<E, P, V, Response & TypedResponse<ZodValidatorFailureBody<T>, 400, 'json'>>`. The `Response &` intersection prevents a `TS2322` where the bare `TypedResponse<...>` could not be assigned to a generic `MiddlewareHandler` (R must extend `Response`).
- Apply the same `Response &` intersection inside `ExtractValidationResponse<VF>` so the with-hook path is symmetrical.
- Collapse the previous two no-hook signatures into one that takes `hook?: undefined, options?: ValidationFunctionOption<T, Target>`, so `zValidator(target, schema, undefined, { validationFunction })` matches the typed-failure path. `ValidationFunctionOption<T, Target>` is extracted as a named type to deduplicate the options shape across overloads.

**`packages/zod-validator/src/{v3,v4}.test.ts`**

- Update each `Equal<Expected, Actual>` assertion to include the new `400` failure variant alongside the existing success variant.
- Add a `Should be assignable to a generic MiddlewareHandler` it-block to the `Basic` and `With Hook` describes (both versions).
- Add a schema assertion to `With options + validationFunction` so the new failure variant is also covered there.

**`.changeset/zod-validator-typed-400-default.md`**

- New `patch` changeset summarising all of the above.

## Test plan

- [x] `yarn workspace @hono/zod-validator test --run` — 28 passed.
- [x] `yarn workspace @hono/zod-validator typecheck` — clean.
- [x] `yarn run -T prettier --write` on the affected files; `yarn run -T eslint src` reports 0 errors.

## Related history

- honojs/hono#4393 — feat(types): passing middleware types — introduced `MergeMiddlewareResponse<M_k>` for `HandlerInterface`.
- honojs/hono#4906 — fix(types): propagate middleware response types to app.on overloads — extended the same pattern to `OnHandlerInterface`.
- honojs/hono#3746 — the user-facing report. This PR is the missing piece on the validator side.

---

- [x] Add tests
- [x] Run tests


T4ko0522:
If you need any additional PRs for https://github.com/honojs/hono to test whether this resolves the issue at https://github.com/honojs/hono/issues/3746, I would be happy to create one!

T4ko0522:
This fix requires Hono v4.10.0+, since the middleware response type parameter and RPC response extraction used here were introduced in that release.

I bumped `@hono/zod-validator`'s `hono` peer dependency to `>=4.10.0`, and also bumped `@hono/zod-openapi` for consistency because it depends on `@hono/zod-validator` directly. Without that cascade bump, `@hono/zod-openapi` can still be installed with `hono@4.9.x` and then fail due to the bundled validator types.

If you'd prefer not to take these floor bumps in this PR, please feel free to close it.

T4ko0522:
Sorry, I forgot to update the root lockfile. Pushed the regenerated yarn.lock

.changeset/zod-openapi-hono-peer-bump.md
yusukebe, line 1: This should be considered, but I wonder if it is a `minor` bump instead of a `patch` because it updates the `hono` version in `peerDependencies`.  What do you think of this?

T4ko0522: I think so too. Initially, I thought this was a change to address a bug fix for `zod-validator`, so I considered a `patch` to be appropriate, but I believe a `minor` version update would be a more appropriate choice.

.changeset/zod-validator-typed-400-default.md
```diff
@@ -0,0 +1,10 @@
+---
+'@hono/zod-validator': patch
```
yusukebe: This also should be a `minor`.

T4ko0522: I'm so sorry...
I was a bit spaced out and forgot about it.
I think that is correct, so I will fix it.

yusukebe: No problem! Thanks.

yusukebe (approved):
LGTM!

yusukebe:
@T4ko0522 

Looks good! This PR solves a long-time problem. Awesome. Thanks!

T4ko0522:
@yusukebe 

Thanks a lot!

I am glad that I could resolve this problem!

PR #1885, opened 2026-05-08
fix(otel): set requestduration buckets to follow semconv

### LINKED ISSUE #1861: @hono/otel: histogram buckets default to ms-scale, quantile queries don't work

@hono/otel

1.1.1

4.12.14

Node.js 24.15.0 (Railway)

@hono/otel creates its `http.server.request.duration` histogram without
passing `advice.explicitBucketBoundaries`:

```ts
// packages/otel/src/index.ts
const histogram = getMeter(config).createHistogram(
  METRIC_HTTP_SERVER_REQUEST_DURATION,
  {
    unit: 's',
    description: '...',
  },
);
```

With no advice, the OTel SDK falls back to its static default bucket
boundaries, which are legacy ms-scale values:
`[0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000]`
(see `sdk-metrics/src/view/Aggregation.ts`).

Since PR #1784 correctly converts the recorded value to seconds, every
sub-second request (the common case) lands in the first bucket `[0, 5s]`.

Minimal repro:

1. Use `@hono/otel` >= 1.1.1 with any OTLP metric exporter.
2. Serve any route that returns in <5 seconds (i.e. ~all routes).
3. Query Prometheus / VictoriaMetrics / any backend:

   ```promql
   histogram_quantile(0.99,
     sum by(le) (rate(http_server_request_duration_seconds_bucket[5m]))
   )
   ```

4. Observe the result is always ~4.95 regardless of actual latency.

`histogram_quantile(0.99, ...)` should approximate the real p99 latency.

For a service where requests typically take a few ms, p99 should be on the
order of tens-to-hundreds of milliseconds — not ~4.95 seconds.

### What do you see instead?

`histogram_quantile(0.99)` reports ~4.95s for every route, because the SDK
interpolates linearly within the `[0, 5s]` bucket: 99% of the way through
that bucket is `0 + 5 * 0.99 = 4.95s`.

Direct query on a sample histogram confirms the bucket layout:

```
le=0       count=0
le=5       count=29
le=10      count=29
le=25      count=29
... (all identical through le=10000)
le=+Inf    count=29
```

All 29 requests (mean latency 0.68ms per `_sum / _count`) sit in the first
bucket. `histogram_quantile` has no signal to differentiate fast from slow
within that bucket.

### Additional information

The OTel JS SDK's static histogram default was inherited from the pre-stable-semconv era, when HTTP durations were recorded in milliseconds. Stable HTTP semconv later switched units to seconds, but the SDK default boundaries were never updated — instead, well-behaved instrumentations pass their own advice.explicitBucketBoundaries to override.

@opentelemetry/instrumentation-http does this correctly:

```ts
this._stableHttpServerDurationHistogram = this.meter.createHistogram(
  METRIC_HTTP_SERVER_REQUEST_DURATION,
  {
    description: '...',
    unit: 's',
    advice: {
      explicitBucketBoundaries: [
        0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1,
        2.5, 5, 7.5, 10,
      ],
    },
  },
);
```

@hono/otel needs the same advice block. PR #1784 fixed the value unit (dividing by 1000 before recording) but did not add bucket advice — a one-line oversight that leaves the histogram effectively unreadable via histogram_quantile for the common sub-second request case.

Proposed fix: add advice.explicitBucketBoundaries with the stable HTTP semconv defaults (same values instrumentation-http uses), so Hono apps and apps instrumented via instrumentation-http land on identical bucket grids and share PromQL queries / dashboards.

Backward compatibility: similar character to #1784 — which was itself a behavior change shipped as a patch. Any user relying on the current (broken) bucket layout for alerts would see le labels change. Users who already override via an SDK-level View are unaffected (View precedence wins over advice).

Workaround for users hitting this today — override at the SDK level via a View in your telemetry bootstrap:

```ts
import { AggregationType } from '@opentelemetry/sdk-metrics';

new NodeSDK({
  views: [{
    instrumentName: 'http.server.request.duration',
    aggregation: {
      type: AggregationType.EXPLICIT_BUCKET_HISTOGRAM,
      options: {
        boundaries: [
          0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1,
          2.5, 5, 7.5, 10,
        ],
      },
    },
  }],
  // ...
});
```

References:
- Related fix converting value to seconds: #1784 (closes #1759)
- OTel HTTP semconv bucket recommendation: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration
- OTel Instrument Advisory Parameters spec: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advisory-parameters

mrlorentx:
Fixes #1861 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/six-onions-fry.md
+++ b/.changeset/six-onions-fry.md
@@ -1,5 +1,5 @@
 ---
-'@hono/otel': minor
+'@hono/otel': patch
 ---
 
 Fixes request duration histogram bucketing
--- a/.changeset/six-onions-fry.md
+++ b/.changeset/six-onions-fry.md
@@ -1,0 +1,5 @@
+---
+'@hono/otel': minor
+---
+
+Fixes request duration histogram bucketing
--- a/packages/otel/src/index.test.ts
+++ b/packages/otel/src/index.test.ts
@@ -345,7 +345,7 @@ describe('OpenTelemetry middleware - Metrics (combined)', () => {
     const durationMetric = metrics.find((m) => m.descriptor.name === 'http.server.request.duration')
     assert.ok(durationMetric)
     assert.strictEqual(durationMetric.descriptor.unit, 's')
-    const dps = durationMetric.dataPoints
+    const dps = durationMetric.dataPoints as DataPoint<Histogram>[]
     assert.strictEqual(dps.length, 1)
     const dp = dps[0]
     assert.strictEqual(dp.attributes['http.request.method'], 'GET')
--- a/packages/otel/src/index.test.ts
+++ b/packages/otel/src/index.test.ts
@@ -353,6 +353,10 @@ describe('OpenTelemetry middleware - Metrics (combined)', () => {
     if (dp.attributes['http.response.status_code']) {
       assert.strictEqual(dp.attributes['http.response.status_code'], 200)
     }
+    assert.deepStrictEqual(
+      dp.value.buckets.boundaries,
+      [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
+    )
   })
 
   it('Should include serviceName and serviceVersion in metric attributes', async () => {
--- a/packages/otel/src/trackers.ts
+++ b/packages/otel/src/trackers.ts
@@ -18,6 +18,11 @@ export const createRequestDurationTracker = (
     description: 'Duration of HTTP requests in seconds',
     unit: 's',
     valueType: ValueType.DOUBLE,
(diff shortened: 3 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @mrlorentx 

I've updated some myself. Looks good. Thank you for your contribution!

Issue #1900, opened 2026-05-20
Proposal: support Inertia's partial reloads, deferred props, merge & infinite scroll in @hono/inertia


ashunar0:
## Summary

`@hono/inertia` currently implements the core of the Inertia protocol (the page
object, asset versioning with `409` reloads, and the XHR/JSON/HTML response
split). The next layer of the protocol — **partial reloads, deferred props,
merge props, and infinite scroll** — isn't covered yet.

I'd like to contribute these upstream, but since they're a connected set of
features (and partly depend on each other), I wanted to open an issue first to
align on direction before sending PRs.

## Background

While building a small SaaS-style app (a Todo app on Hono + Inertia + hono/jsx,
SSR), I needed these features and implemented them in a protocol-compatible fork,
`@ashunar0/hono-inertia-plus`, which I've been dogfooding. The API stays close to
the current `c.render`:

```ts
import { inertiaPlus, defer, scroll } from '@ashunar0/hono-inertia-plus'

app.use(inertiaPlus())

app.get('/', (c) =>
  c.render('Home', {
    // deferred: resolver is skipped on the initial response, advertised via
    // page.deferredProps, then fetched by the client's <Deferred> as a partial reload
    tasks: defer(() => db.tasks.list(userId)),
  })
)
```

What I've verified end-to-end in that app:

- **Partial reloads** (`only` / `except`) — `router.reload({ only: ['tasks'] })`
- **Deferred props** — `defer(() => ...)` + `<Deferred>` (shown above)
- **Merge props** & **Infinite scroll** — `scroll({ ... })` emits
  `page.scrollProps` / `mergeProps` for Inertia v3's `<InfiniteScroll>`

All four run through a single renderer, mirroring how the Laravel/Rails adapters
layer these on top of the same partial-reload machinery. Source:
https://github.com/ashunar0/hono-playground/blob/main/packages/hono-inertia-plus/src/index.ts

## Proposed plan

These have a natural dependency order, so I'd send them as separate, reviewable
PRs rather than one big one:

```
① partial reloads (only / except)   ← foundation
   ├─ ② deferred props               ← builds on ①
   └─ ③ merge props                  ← builds on ①
         └─ ④ infinite scroll        ← builds on ① + ③
```

`①` is the foundation: it reads `X-Inertia-Partial-Component` /
`X-Inertia-Partial-Data` / `X-Inertia-Partial-Except` and filters props
accordingly. The other three advertise extra fields on the page object that the
client adapter reads.

[...]

PR #1904, opened 2026-05-22
feat(inertia): add partial reloads support with lazy function props

ashunar0:
Resolves the first item ("partial reloads") of #1900.

## Summary

Adds server-side support for [Inertia.js partial reloads](https://inertiajs.com/partial-reloads):

- Reads `X-Inertia-Partial-Component`, `X-Inertia-Partial-Data` (only), and `X-Inertia-Partial-Except` headers.
- Accepts function-valued props that are evaluated lazily, so heavy data fetching is skipped for props that the client did not request.

```ts
app.get('/', (c) =>
  c.render('Home', {
    fast: 'inline value',
    slow: () => db.expensiveQuery(), // not invoked when excluded by a partial
  })
)
```

## Scope

This PR sticks to the foundation that the rest of the breakdown in #1900 builds on:

- ✅ `only` / `except` props filtering
- ✅ Lazy function props (`() => T | Promise<T>`)
- ✅ Component-mismatch guard: partial headers are ignored when `X-Inertia-Partial-Component` does not match the rendered component

Deferred to follow-up PRs:
- ② Deferred props (`<Deferred>`)
- ③ Merge / prepend props
- ④ Infinite scroll
- ⑤ `always(fn)` marker for one-shot props such as flashed messages

## Behavior notes

- A request is treated as a partial reload only when `X-Inertia-Partial-Component` matches the component being rendered **and** at least one of `X-Inertia-Partial-Data` / `X-Inertia-Partial-Except` is present. Otherwise the request is processed as a normal Inertia visit.
- The renderer stays fully synchronous when no function-valued prop is present, preserving the existing `Response` (non-`Promise<Response>`) return type for the common case.
- Function props excluded from the partial render are never invoked.

## Backward compatibility

Callers that pass only plain values keep the exact same behavior. The new type `ResolvedProps<P>` awaits the return type of any function-valued prop — plain props are pass-through.

## Tests

9 new tests in `packages/inertia/src/index.test.ts` covering the partial filter, lazy invocation skip, async resolution, component mismatch, whitespace handling, only+except intersection, and the "partial-only header without component" guard.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/inertia/src/index.ts
+++ b/packages/inertia/src/index.ts
@@ -137,7 +137,8 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
       const partialComponent = c.req.header('X-Inertia-Partial-Component')
       const partialData = c.req.header('X-Inertia-Partial-Data')
       const partialExcept = c.req.header('X-Inertia-Partial-Except')
-      const isPartial = partialComponent === component && Boolean(partialData || partialExcept)
+      const isPartial =
+        partialComponent === component && (partialData !== undefined || partialExcept !== undefined)
 
       const parseKeys = (header: string | undefined): string[] | null =>
         header
@@ -161,8 +162,12 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
       const kept: [string, unknown][] = []
       let hasFunction = false
       for (const [key, value] of Object.entries(propsInput)) {
-        if (isExcluded(key)) continue
-        if (typeof value === 'function') hasFunction = true
+        if (isExcluded(key)) {
+          continue
+        }
+        if (typeof value === 'function') {
+          hasFunction = true
+        }
         kept.push([key, value])
       }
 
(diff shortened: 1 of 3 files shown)


packages/inertia/src/index.ts
```diff
@@ -109,31 +128,87 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
...
+      // Partial reload negotiation: the client (Inertia core) signals
+      // "only re-evaluate these props" via X-Inertia-Partial-Component +
+      // X-Inertia-Partial-Data / X-Inertia-Partial-Except.
+      const partialComponent = c.req.header('X-Inertia-Partial-Component')
+      const partialData = c.req.header('X-Inertia-Partial-Data')
+      const partialExcept = c.req.header('X-Inertia-Partial-Except')
+      const isPartial =
+        partialComponent === component && (partialData !== undefined || partialExcept !== undefined)
+
+      const parseKeys = (header: string | undefined): string[] | null =>
```
yusukebe: You can define `parseKeys` outside of `inertia`, at the top level.

ashunar0: Good catch, thanks! Hoisted in 0c96f56.

yusukebe (approved):
LGTM!

yusukebe:
@ashunar0 Thanks! I'll merge.

PR #1911, opened 2026-05-28
feat(inertia): add deferred props with defer()

ashunar0:
## Summary

Second step of the [`@hono/inertia` v3 protocol breakdown](https://github.com/honojs/middleware/issues/1900): **② Deferred props**, built on top of the partial reload foundation from #1904.

Adds a `defer(resolver, group?)` helper. On the initial response the resolver is skipped and the prop key is advertised under `page.deferredProps[group]`. The Inertia client then issues one partial reload per group, at which point the resolver runs and the value is sent down — letting heavy data fetching happen after the page becomes interactive.

```ts
import { defer, inertia } from '@hono/inertia'

app.use(inertia())

app.get('/', (c) =>
  c.render('Dashboard', {
    user: { id: 1 },                       // sent on initial response
    posts: defer(() => fetchPosts()),      // fetched after mount
    stats: defer(() => fetchStats(), 'secondary'),
  }),
)
```

Multiple deferred props that share a `group` are fetched together in a single partial reload. The default group is `"default"`.

## Protocol behaviour

| Visit kind | `posts: defer(fn)` behaviour |
|---|---|
| Initial (full page or `X-Inertia` GET) | `fn` **not invoked**, `page.deferredProps.default = ['posts']` emitted |
| Partial reload that includes `posts` in `X-Inertia-Partial-Data` | `fn` invoked, value returned under `props.posts` |
| Partial reload that does **not** include `posts` | `fn` **not invoked**, key omitted from response |
| Non-Inertia `Accept: application/json` request | `fn` **not invoked**, key omitted |

`deferredProps` is only emitted on the initial response; subsequent partial reloads strip it, so the client does not re-trigger the same group.

## Implementation

- `defer<T>(resolver, group = 'default'): T` returns an opaque marker (`Symbol.for('@hono/inertia/defer')`) that the renderer unwraps. The return type is `T`, so call sites stay transparent and `TypedResponse` infers the awaited resolver type.
- The renderer loop now distinguishes three prop kinds — plain values, function-valued props (from #1904), and deferred markers — and decides on a single sync/async path based on whether any resolver needs awaiting. The non-deferred sync path is preserved when no async work is needed.
- `PageObject` gains an optional `deferredProps?: Record<string, string[]>` field, only present on the initial response when at least one prop was deferred.

Scope intentionally kept small (single PR per protocol feature, like #1904):

- ✅ Top-level `defer()` markers
- ✅ Multiple groups + default group
- ✅ Skip on initial, resolve on requested partial
- ❌ Nested / dot-path deferred markers — separate PR
- ❌ `merge` / `prepend` / `deepMerge` / `always(fn)` / infinite scroll markers — separate PRs (③④⑤ in the breakdown)

## Test plan

- [x] `yarn workspace @hono/inertia test` (`28 passed`, 9 new tests covering initial skip / group emission / partial resolve / `only` filter / JSON request / mixed deferred + function props / no-defer no-emit / non-XHR HTML visit)
- [x] `eslint packages/inertia` clean
- [x] `prettier --check` clean
- [x] `tsc -b` clean
- [x] `tsdown` build success

### DIFF
--- a/packages/inertia/src/index.test.ts
+++ b/packages/inertia/src/index.test.ts
@@ -539,7 +539,7 @@ describe('inertia', () => {
 
       const html = await res.text()
       expect(html).toContain('"deferredProps":{"default":["posts"]}')
-      expect(html).not.toContain('"posts"\\:[{')
+      expect(html).not.toContain('"posts":[{')
     })
 
     it('mixes deferred props with eager function props on partial reloads', async () => {
--- a/packages/inertia/src/index.test.ts
+++ b/packages/inertia/src/index.test.ts
@@ -1,7 +1,7 @@
 import { Hono } from 'hono'
 import { describe, expect, it } from 'vitest'
 import type { PageObject } from './index'
-import { inertia } from './index'
+import { defer, inertia } from './index'
 
 describe('inertia', () => {
   describe('full page (non Inertia request)', () => {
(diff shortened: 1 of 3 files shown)


packages/inertia/src/index.test.ts
```diff
@@ -386,4 +386,219 @@ describe('inertia', () => {
...
+          user: { id: 1 },
+          posts: defer(() => [{ id: 1 }]),
+        })
+      )
+
+      const res = await app.request('/')
+
+      const html = await res.text()
+      expect(html).toContain('"deferredProps":{"default":["posts"]}')
+      expect(html).not.toContain('"posts"\\:[{')
```
yusukebe: I think this should be below because the colon is not escaped:

```ts
expect(html).not.toContain('"posts":[{')
```

yusukebe:
I forgot to say that in #1904. Can you update the README for these new features? You can put both #1904 and this PR now.

packages/inertia/src/index.test.ts
ashunar0: Good catch — confirmed it was a no-op assertion. `serializePage` only escapes `/` → `\/`, so the `\:` form never appeared in the output and the check passed trivially. Fixed in 84f47fe to assert against the real serialized form (`'"posts":[{'`). Test still passes, but now it actually verifies the deferred value is omitted from the initial HTML.

ashunar0:
@yusukebe Thanks for the review!

- Inline assertion fix: addressed in 84f47fe (root cause and details in the [thread](https://github.com/honojs/middleware/pull/1911#discussion_r3339685776)).
- README update: opened as a separate PR — #1920 — covering both partial reloads (#1904) and deferred props (this PR).

yusukebe (approved):
LGTM!

yusukebe:
@ashunar0 Thank you so much!

PR #1912, opened 2026-05-28
docs(sentry): Add note about official `@sentry/hono` SDK

s1gr1d:
Adding a note with the link to the official docs.

### DIFF
--- a/packages/sentry/README.md
+++ b/packages/sentry/README.md
@@ -2,6 +2,9 @@
 
 [![codecov](https://codecov.io/github/honojs/middleware/graph/badge.svg?flag=sentry)](https://codecov.io/github/honojs/middleware)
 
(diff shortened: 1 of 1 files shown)


s1gr1d:
Hi, @yusukebe what do you think of adding this? :) 
Greetings from Sentry!

yusukebe:
Thanks! @s1gr1d 

This makes sense. I'll merge it immediately.

And shall we mark `@hono/sentry` as deprecated? What do you think about? @sam-lippert

sam-lippert:
@yusukebe Yes, sounds good to me. I would still prefer to have a solution that doesn't require swapping out the sentry implementation for cross-engine compatibility. The viability of that has been degraded for now, but has workarounds.

PR #1915, opened 2026-05-29
feat(otel): add getRoute callback so adapters can resolve http.route

### LINKED ISSUE #1914: @hono/otel overwrites http.route set by downstream middleware

@hono/otel

1.1.2

4.12.23

Node.js 24

A Hono app uses `@hono/otel` around an adapter route such as `/rpc/*`. A downstream RPC adapter resolves the actual operation and sets `http.route` on the active span.

Example:

- Hono route: `/rpc/*`
- Resolved operation: `/rpc/user/getProfile`
- Downstream middleware sets `http.route = "/rpc/user/getProfile"`

After the request finishes, `@hono/otel` runs its finalize step and sets `http.route` back to `routePath(c)`, which is `/rpc/*`.

If downstream middleware changes `http.route`, `@hono/otel` should not overwrite it during finalize, or it should provide a config hook for choosing the final route value.

The same final route value should be used for `http.server.request.duration`, since today that metric also uses `routePath(c)` directly.

### What do you see instead?

The span is finalized with `http.route = "/rpc/*"`, and request duration metrics are also recorded under `/rpc/*`. All RPC operations behind the adapter route are grouped together even though the actual operation is known during request handling.

### Additional information

`spanNameFactory` does not fully solve this because it only changes the span name. It does not control the final `http.route` attribute or the request duration metric attributes.

spokodev:
Closes #1914.

`@hono/otel` overwrote the `http.route` attribute on the active span at finalize time with the matched Hono pattern (e.g. `/rpc/*`), even when a downstream adapter resolved a more specific operation name (e.g. `/rpc/user/getProfile`) during request handling. The same Hono pattern was reported as the `http.route` attribute of the `http.server.request.duration` metric, so per-operation latency collapsed into the pattern bucket.

This PR adds a `getRoute(c)` config callback. When it returns a non-empty string the value is used at finalize for both the span attribute and the metric attribute. When it returns `undefined`, an empty string, or throws, the middleware falls back to `routePath(c)`, so the existing behavior is preserved when the callback is not provided.

Typical adapter usage:

```ts
app.use(
  httpInstrumentationMiddleware({
    tracerProvider,
    getRoute: (c) => c.get('rpc.route'),
  })
)
```

Verification:

- 5 new tests in `packages/otel/src/index.test.ts`:
  - span `http.route` reflects the `getRoute` return value
  - metric `http.route` reflects the `getRoute` return value
  - `getRoute` returning `undefined` falls back to `routePath(c)`
  - `getRoute` returning an empty string falls back to `routePath(c)`
  - `getRoute` throwing falls back to `routePath(c)`
- RED baseline confirmed: reverting `index.ts` + `types.ts` to `main` makes the two positive tests fail; applying the fix turns them green
- `yarn test --run` for `@hono/otel`: 27 / 27 (22 existing + 5 new)
- `yarn typecheck` clean
- `yarn turbo run lint --filter=@hono/otel` clean
- `yarn format` reports `All matched files use Prettier code style!`
- Changeset added at `.changeset/otel-get-route-callback.md` (`@hono/otel`: minor)

Out of scope:

- The initial attributes passed to `tracer.startActiveSpan` still use `routePath(c)`, because `getRoute` may rely on context state that downstream middleware writes after the span is started. The finalize step updates the attribute to the resolved value so the exported span and the metric agree.

### DIFF
--- a/.changeset/otel-get-route-callback.md
+++ b/.changeset/otel-get-route-callback.md
@@ -1,0 +1,9 @@
+---
+'@hono/otel': minor
+---
+
+feat(otel): add `getRoute` config callback so downstream adapters can resolve `http.route`
+
+The middleware previously overwrote the `http.route` attribute on the active span with the matched Hono route pattern at finalize time, even when a downstream adapter (such as an RPC layer) had already set a more specific value on the span. The same pattern was also reported as the `http.route` attribute of the `http.server.request.duration` metric, so per-operation latency collapsed into the pattern bucket.
+
+`getRoute(c)` is now an optional config callback. When it returns a non-empty string it is used for both the span attribute and the metric attribute; when it returns `undefined`, an empty string, or throws, the middleware falls back to the existing `routePath(c)` behavior. This is fully backward compatible and resolves [issue #1914](https://github.com/honojs/middleware/issues/1914).
--- a/packages/otel/src/index.test.ts
+++ b/packages/otel/src/index.test.ts
@@ -313,6 +313,71 @@ describe('OpenTelemetry middleware - Spans (combined)', () => {
     const spans = memoryExporter.getFinishedSpans()
     assert.strictEqual(spans.length, 0)
   })
+
+  // Regression tests for https://github.com/honojs/middleware/issues/1914 - a downstream
+  // adapter (such as an RPC layer) can now expose a resolved operation name via the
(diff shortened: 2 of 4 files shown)


yusukebe:
Oops, sorry for being super late.

Hey, one of @dahlia @mrlorentx, can you review this?

PR #1932, opened 2026-06-05
feat(inertia): add merge props with merge(), prepend(), and deepMerge()

ashunar0:
## Summary

Third step of the [`@hono/inertia` v3 protocol breakdown](https://github.com/honojs/middleware/issues/1900): **③ Merge props**, built on top of the partial reload foundation from #1904.

Adds three helpers — `merge()`, `prepend()`, and `deepMerge()` — that mark a prop for client-side combination on the **next partial reload**, instead of the default replace behavior. The wrapped value travels as-is; the renderer advertises which keys to combine via the new `page.mergeProps` / `page.prependProps` / `page.deepMergeProps` arrays and emits dot-paths on `page.matchPropsOn` for the client's dedupe logic.

```ts
import { deepMerge, inertia, merge, prepend } from '@hono/inertia'

app.use(inertia())

app.get('/feed', (c) =>
  c.render('Feed', {
    posts: merge(await db.posts.page(n), { matchOn: 'id' }),
    notifications: prepend(await fetchNotifications(), { matchOn: 'id' }),
    conversations: deepMerge(
      { data: await db.messages.page(n), meta: { nextCursor } },
      { matchOn: 'data.id' },
    ),
  }),
)
```

## Protocol behaviour

| Visit kind | Behaviour |
|---|---|
| Initial / partial / JSON request | Wrapped value is unwrapped and sent as a plain prop |
| Response with at least one merge marker | `page.mergeProps` / `prependProps` / `deepMergeProps` / `matchPropsOn` emitted on **every** response (initial + partial) so the client knows which keys to combine on the next partial reload |
| Merge-marked prop excluded by `only`/`except` | Stripped before any metadata is recorded (no leak in `mergeProps` etc.) |
| Function value inside `merge()` (e.g. `merge(() => fetchPosts(), ...)`) | Resolved on the same lazy path as plain function props from #1904 |

Merging itself only takes effect on subsequent partial reloads — full page visits always replace props entirely (matches the upstream protocol).

## Implementation

- `merge<T>(data, { matchOn? }): T`, `prepend<T>(...)`, `deepMerge<T>(...)` return opaque markers (`Symbol.for('@hono/inertia/merge')`) that the renderer unwraps. The return type is `T`, so call sites and `TypedResponse` stay transparent — mirrors the `defer()` pattern from #1911.
- `matchOn` accepts a string or array; entries are emitted as `"<propKey>.<field>"` on `page.matchPropsOn` (e.g. `merge(posts, { matchOn: 'id' })` ⇒ `matchPropsOn: ['posts.id']`).
- `PageObject` gains four optional fields — `mergeProps?: string[]`, `prependProps?: string[]`, `deepMergeProps?: string[]`, `matchPropsOn?: string[]` — emitted only when at least one prop in the response carries the corresponding marker.

Scope intentionally kept small (single PR per protocol feature, like #1904 / #1911):

- ✅ Top-level \`merge()\` / \`prepend()\` / \`deepMerge()\` markers
- ✅ Single string and array \`matchOn\`
- ✅ Function value inside \`merge()\` (resolved via the existing lazy path)
- ✅ Strategy mixing in a single response
- ❌ Nested / dot-path merge markers — separate PR (would need to land alongside dot-path partial reloads)
- ❌ \`defer()\` inside \`merge()\` — separate PR (semantics deserve their own design discussion)
- ❌ Infinite scroll markers — separate PR (④ in the breakdown)

## Test plan

- [x] \`yarn workspace @hono/inertia test\` (\`40 passed\`, 12 new tests covering each helper, single + array \`matchOn\`, strategy mixing, partial reload behaviour, excluded-prop guard, function-inside-merge, HTML embed, JSON request)
- [x] \`eslint packages/inertia\` clean
- [x] \`prettier --check\` clean
- [x] \`tsc -b\` clean
- [x] \`tsdown\` build success

### DIFF
--- a/.changeset/inertia-merge-props.md
+++ b/.changeset/inertia-merge-props.md
@@ -1,0 +1,38 @@
+---
+'@hono/inertia': minor
+---
+
+feat(inertia): add merge props with `merge()`, `prepend()`, and `deepMerge()`
+
+Adds three helpers that mark a prop for client-side combination on the
+**next partial reload**, instead of the default replace behavior:
+
+- `merge(data, { matchOn? })` — appends array items / shallow-spreads object keys.
+- `prepend(data, { matchOn? })` — same as `merge()` but inserts at the start.
+- `deepMerge(data, { matchOn? })` — recurses into nested arrays/objects (use for
+  wrapper-shaped paginated props like `{ data: [...], meta: {...} }`).
+
+The value travels as-is; the renderer advertises which keys to combine via the
+new `page.mergeProps` / `page.prependProps` / `page.deepMergeProps` arrays and
+emits dot-paths on `page.matchPropsOn` for the client's dedupe logic.
+
+```ts
+import { deepMerge, inertia, merge, prepend } from '@hono/inertia'
+
+app.use(inertia())
+
+app.get('/feed', (c) =>
+  c.render('Feed', {
+    posts: merge(await db.posts.page(n), { matchOn: 'id' }),
+    notifications: prepend(await fetchNotifications(), { matchOn: 'id' }),
+    conversations: deepMerge(
+      { data: await db.messages.page(n), meta: { nextCursor } },
+      { matchOn: 'data.id' },
+    ),
+  }),
+)
+```
+
+The merge metadata is emitted on every response (initial + partial) so the
+client knows which keys to combine on the next partial reload. Full page visits
+always replace props entirely.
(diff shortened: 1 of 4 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@ashunar0 Nice! Thank you!

PR #1934, opened 2026-06-06
feat(inertia): add infinite scroll with scroll()

ashunar0:
## Summary

Final step of the [`@hono/inertia` v3 protocol breakdown](https://github.com/honojs/middleware/issues/1900): **④ Infinite scroll**, built on top of the merge protocol from #1932.

Adds a `scroll()` helper that wraps a paginated page payload with the metadata Inertia's [`<InfiniteScroll>`](https://inertiajs.com/docs/v2/data-props/infinite-scroll) adapter needs to keep loading more items as the user scrolls. The wrapped value travels as-is; the renderer emits a new `page.scrollProps` map with the previous/next page cursor and opts the prop into the merge protocol so each incoming page is appended (or prepended) to the cached array instead of replacing it.

```ts
import { inertia, scroll } from '@hono/inertia'

app.use(inertia())

app.get('/users', async (c) => {
  const currentPage = Number(c.req.query('users_page') ?? 1)
  return c.render('Users/Index', {
    users: scroll({
      data: await db.users.page(currentPage),
      currentPage,
      lastPage: 10,
      pageName: 'users_page',
      matchOn: 'id',
    }),
  })
})
```

Mirrors [`Inertia::scroll(...)`](https://github.com/inertiajs/inertia-laravel/blob/3.x/src/ScrollProp.php) in inertia-laravel 3.x.

## Protocol behaviour

| Visit kind | Behaviour |
|---|---|
| Any response with a `scroll()` prop | `scrollProps[key] = { previousPage, nextPage, currentPage, pageName }` emitted and `mergeProps.push(key)` by default |
| Request with `X-Inertia-Infinite-Scroll-Merge-Intent: prepend` | `prependProps.push(key)` instead of `mergeProps` (the `<InfiniteScroll>` adapter sends this when scrolling backwards) |
| `matchOn` passed (string or array) | One `"<key>.<field>"` entry per field appended to `page.matchPropsOn` for client-side dedupe; omitted entirely when `matchOn` is not passed |
| Scroll-marked prop excluded by `only` / `except` | Stripped before any metadata is recorded (no leak in `scrollProps`, `mergeProps`, or `matchPropsOn`) |
| Initial / partial / JSON request | Wrapped value is unwrapped and sent as a plain prop |

`previousPage` is `currentPage - 1` (or `null` on the first page); `nextPage` is `currentPage + 1` (or `null` on the last page).

## Implementation

- `scroll<T>({ data, currentPage, lastPage, pageName, matchOn? }): T[]` returns an opaque marker (`Symbol.for('@hono/inertia/scroll')`) that the renderer unwraps — same pattern as `defer()` (#1911) and `merge()` (#1932). The return type is `T[]` so call sites and `TypedResponse` stay transparent.
- `PageObject` gains one optional field, `scrollProps?: Record<string, ScrollDescriptor>`, where `ScrollDescriptor = { previousPage, nextPage, currentPage, pageName }`. `matchOn` is intentionally **not** part of the descriptor — it travels only via `page.matchPropsOn`, matching `Inertia\ScrollProp::metadata()` in the Laravel adapter.
- Default merge direction is `append`; the client switches it by sending `X-Inertia-Infinite-Scroll-Merge-Intent: prepend`. Unknown header values fall back to `append`. Mirrors `ScrollProp::configureMergeIntent` in inertia-laravel.
- `matchOn` accepts a string or string array, optional with no default, matching `Inertia::merge(matchOn:)` and the existing `merge()` helper in this package.

Scope intentionally kept small (single PR per protocol feature, like #1904 / #1911 / #1932):

- ✅ Top-level `scroll()` marker
- ✅ Single string and array `matchOn`
- ✅ Default `append` + `prepend` via the merge-intent header
- ✅ Coexistence with `merge()` / `prepend()` / `deepMerge()` / `defer()` in the same response
- ❌ Laravel paginator auto-extraction — Hono has no equivalent paginator object, so `data` / `currentPage` / `lastPage` / `pageName` are passed explicitly
- ❌ Nested / dot-path scroll markers — separate PR (would need to land alongside dot-path partial reloads, same constraint as #1932)

## Test plan

- [x] \`yarn workspace @hono/inertia test\` (\`56 passed\`, 16 new tests covering paging metadata, \`previousPage\` / \`nextPage\` boundaries, default-append direction, \`prepend\` via header, single / array \`matchOn\`, no-\`matchOn\` omission, partial-reload preservation, \`only\` / \`except\` guard, multi-\`scroll()\` in one response, coexistence with \`merge()\` / \`prepend()\`, HTML embed, raw JSON request, unknown header value fallback)
- [x] \`eslint packages/inertia\` clean
- [x] \`prettier --check\` clean
- [x] \`tsc -b\` clean
- [x] \`tsdown\` build success

### DIFF
--- a/.changeset/inertia-infinite-scroll.md
+++ b/.changeset/inertia-infinite-scroll.md
@@ -1,0 +1,26 @@
+---
+'@hono/inertia': minor
+---
+
+feat(inertia): add infinite scroll with `scroll()`
+
+Adds a `scroll()` helper that wraps a paginated page payload with the metadata
+the Inertia client's `<InfiniteScroll>` adapter needs to keep loading more items
+as the user scrolls:
+
+```ts
+users: scroll({
+  data: await db.users.page(currentPage),
+  currentPage,
+  lastPage: 10,
+  pageName: 'users_page',
+  matchOn: 'id',
+})
+```
+
+The value travels as-is; the renderer emits `page.scrollProps[key] = {
+previousPage, nextPage, currentPage, pageName }` on every response and opts the
+prop into the merge protocol — defaulting to `append`, switched to `prepend`
+when the client sends `X-Inertia-Infinite-Scroll-Merge-Intent: prepend`. The
+optional `matchOn` is forwarded as `page.matchPropsOn` for client-side dedupe.
+Mirrors `Inertia::scroll(...)` in inertia-laravel 3.x.
--- a/packages/inertia/README.md
+++ b/packages/inertia/README.md
@@ -247,6 +247,37 @@ app.get('/feed', (c) =>
 
 The merge metadata is included on **every** response — initial and partial — so the client knows which keys to combine on the next partial reload. Full page visits always replace props entirely; merging only kicks in on subsequent partial reloads.
 
+## Infinite scroll
+
+[Infinite scroll](https://inertiajs.com/docs/v2/data-props/infinite-scroll) keeps loading more items as the user scrolls — feeds, search results, chat history. The Inertia client's `<InfiniteScroll>` adapter handles the IntersectionObserver and the partial-reload requests; the server just needs to advertise the paging cursor.
+
+Wrap the page payload with `scroll()`:
+
+```ts
+import { inertia, scroll } from '@hono/inertia'
+
+app.use(inertia())
+
+app.get('/users', async (c) => {
+  const currentPage = Number(c.req.query('users_page') ?? 1)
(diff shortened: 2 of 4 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@ashunar0 

I didn't know Inertia.js could do something like `scroll()`. Ineresting! Thanks!

PR #1986, opened 2026-06-18
fix(zod-openapi): infer route middleware env in defineOpenAPIRoute handler

### LINKED ISSUE #1879: [zod-openapi] defineOpenAPIRoute does not infer handler Context variables from route middleware

@hono/zod-openapi

1.3.0

4.12.16

@hono/node-server 4.12.16 on node runtime v22.22.0

  ```ts
  import { createMiddleware } from 'hono/factory'
  import { createRoute, defineOpenAPIRoute } from '@hono/zod-openapi'

  type Identity = { id: string }

  const requireIdentity = createMiddleware<{
    Variables: {
      identity: Identity
    }
  }>(async (c, next) => {
    c.set('identity', { id: '123' })
    await next()
  })

  const route = defineOpenAPIRoute({
    route: createRoute({
      method: 'get',
      path: '/me',
      middleware: [requireIdentity] as const,
      responses: {
        200: {
          description: 'ok',
        },
      },
    }),
    handler: (c) => {
      c.var.identity.id
      return c.text('ok')
    },
  })
```

type of `c.var.identity` to be correctly inferred from `middleware: [requireIdentity] as const` to be of type `Identity`.

### What do you see instead?

`c.var.identity` is of type `any`

### Additional information

There seems to be an easy fix in the code already we just need to use it. If we use [what has been done for `OpenAPIHono.openapi()`](https://github.com/honojs/middleware/blob/95c2039381a44fc817bd8205432d162e20912791/packages/zod-openapi/src/index.ts#L544) instead of ignoring the middleware in `HanlderFromRoute` and `HookFromRoute` we can fix this. If you think this is good I can open a pull request with the proposed fix.

Regards,
aros

### PR DESCRIPTION
**Pull request by VihaanAgarwal**
### What & why

`defineOpenAPIRoute` (added in #1752) types its `handler` with only the app `Env`, so variables set by route-level `middleware` are dropped. Inside the handler `c.var.*` falls back to `any`:

```ts
const requireIdentity = createMiddleware<{
  Variables: { identity: { id: string } }
}>(async (c, next) => {
  c.set('identity', { id: '123' })
  await next()
})

const route = defineOpenAPIRoute({
  route: createRoute({
    method: 'get',
    path: '/me',
    middleware: [requireIdentity] as const,
    responses: { 200: { description: 'ok' } },
  }),
  handler: (c) => {
    c.var.identity // typed as `any`, should be `{ id: string }`
    return c.text('ok')
  },
})
```

`OpenAPIHono.openapi()` already covers this case: when a route has middleware it widens the handler env to `RouteMiddlewareParams<R>['env'] & E`. `HandlerFromRoute`, which `defineOpenAPIRoute` relies on, only passed `E`. This change applies the same widening in `HandlerFromRoute`, so a route built through `defineOpenAPIRoute` infers the middleware variables.

Thanks to @aros for the report and for pointing at the existing `openapi()` handling.

### Test

Added a type test in `index.test-d.ts`, mirroring the existing `openapi()` middleware-env test, that asserts the handler's `c.var.foo` / `c.var.bar` resolve to the route middleware variable types. It fails on `main` (`Property 'foo' does not exist on type ...`) and passes with the fix.

Fixes #1879

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/tidy-mangos-grin.md
+++ b/.changeset/tidy-mangos-grin.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+fix(zod-openapi): infer route middleware env in `defineOpenAPIRoute` handler
--- a/packages/zod-openapi/src/index.test-d.ts
+++ b/packages/zod-openapi/src/index.test-d.ts
@@ -3,7 +3,7 @@ import { createMiddleware } from 'hono/factory'
 import type { ExtractSchema } from 'hono/types'
 import type { ContentfulStatusCode } from 'hono/utils/http-status'
 import type { Equal, Expect } from 'hono/utils/types'
-import { OpenAPIHono, createRoute, z, $ } from './index'
+import { OpenAPIHono, createRoute, defineOpenAPIRoute, z, $ } from './index'
 import type { HonoToOpenAPIHono, MiddlewareToHandlerType, OfHandlerType } from './index'
 
 describe('Types', () => {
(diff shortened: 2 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@VihaanAgarwal Thanks!

Issue #1989, opened 2026-06-21
[@hono/inertia] allow overriding page.url in c.render()


nkfr26:
### Which middleware is the feature for?

@hono/inertia

## Problem

`c.render(component, props)` always derives `page.url` from the current request's path:

```ts
const page: PageObject = {
  component,
  props: resolvedProps,
  url: url.pathname + url.search, // always the request URL
  version,
}
```

This works fine for GET requests, but breaks down when rendering a page directly in response to a non-GET request — e.g. returning validation errors from a POST handler instead of redirecting back:

```ts
app.post('/users', zValidator('json', userInput, (result, c) => {
  if (!result.success) {
    return c.render('Users/New', { errors: toInertiaErrors(result.error) })
  }
}), ...)
```

Here the response is for `POST /users`, so `page.url` becomes `/users`even though the component being shown is `Users/New`. The Inertia client uses `page.url` to update the browser's history state, so:

- the address bar shows `/users` while `Users/New` is rendered
- reloading the page issues a fresh `GET /users`, which resolves to a completely different route (`Users/Index`), losing the form state and errors entirely

This is a known class of issue in the wider Inertia ecosystem (see [inertiajs/inertia-laravel#520](https://github.com/inertiajs/inertia-laravel/issues/520) for a similar symptom), which is why Laravel's convention is to redirect back with flashed session data instead of rendering directly. That works, but it requires wiring up a session layer just to make the URL correct — somewhat heavy for cases where you don't otherwise need server-side flash data.

## Proposed solution

Add an optional third argument to `c.render()` that lets the caller override the `url` embedded in the page object:

```ts
export interface RenderOptions {
  /**
   * Overrides the `url` embedded in the page object instead of deriving it
   * from the current request's pathname + search. Useful when rendering a
   * page directly in response to a non-GET request.
   */
  url?: string
}
```

```ts
declare module 'hono' {
  interface ContextRenderer {
    <C extends PageName, P = Record<string, never>>(
      component: C,
      props?: P,
      options?: RenderOptions
    ): Response & TypedResponse<{ component: C; props: ResolvedProps<P> }, 200, 'html'>
  }
}
```

Usage:

```ts
app.post('/users', zValidator('json', userInput, (result, c) => {
  if (!result.success) {
    return c.render(
      'Users/New',
      { errors: toInertiaErrors(result.error) },
      { url: '/users/new' }
    )
  }
}), ...)
```

[...]

PR #1991, opened 2026-06-23
fix(inertia): resolve page.url for non-GET requests

nkfr26:
Issue: https://github.com/honojs/middleware/issues/1989

#### fix(inertia): use the Referer for non-GET requests to keep the original URL

For non-GET requests, `page.url` now uses the `Referer` header (falling back to `c.req.url`) instead of `c.req.url` to keep the original URL.

#### feat(inertia): add an option to override `page.url`

`c.render(component, props, options)` now accepts an optional third argument `options?: RenderOptions` with a `url` field. Overrides `page.url`. It is primarily used to prevent incorrect URLs when using no-referrer, origin, or strict-origin.

```ts
// You can set `/users/new` as the URL even without a Referer.
app.post('/users', (c) => c.render('Users/New', {}, { url: '/users/new' }))
```

> [!NOTE]
> This option does not exist in inertia-laravel. Whether to include it requires consideration.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/red-trees-tease.md
+++ b/.changeset/red-trees-tease.md
@@ -4,15 +4,15 @@
 
 fix(inertia): use the Referer for non-GET requests to keep the original URL
 
-For non-GET requests, `page.url` now uses the `Referer` header
-(falling back to `c.req.url`) instead of `c.req.url` to keep the original URL.
+For non-GET requests, `page.url` now uses the `Referer` header (falling
+back to `c.req.url`) instead of `c.req.url` to keep the original URL.
 
 feat(inertia): add an option to override `page.url`
 
-`c.render(component, props, options)` now accepts an optional third argument
-`options?: RenderOptions` with a `url` field. Overrides `page.url`.
+`c.render(component, props, options)` now accepts an optional third
+argument `options?: RenderOptions` with a `url` field. Overrides `page.url`.
 It is primarily used to prevent incorrect URLs when using no-referrer.
 
 ```ts
-app.post('/users', (c) => c.render('Users/New', { ok: true }, { url: '/users/new' }))
+app.post('/users', (c) => c.render('Users/New', {}, { url: '/users/new' }))
 ```
--- a/.changeset/red-trees-tease.md
+++ b/.changeset/red-trees-tease.md
@@ -1,0 +1,18 @@
+---
+'@hono/inertia': minor
+---
+
+fix(inertia): use the Referer for non-GET requests to keep the original URL
+
+For non-GET requests, `page.url` now uses the `Referer` header
+(falling back to `c.req.url`) instead of `c.req.url` to keep the original URL.
+
+feat(inertia): add an option to override `page.url`
+
+`c.render(component, props, options)` now accepts an optional third argument
+`options?: RenderOptions` with a `url` field. Overrides `page.url`.
+It is primarily used to prevent incorrect URLs when using no-referrer.
+
+```ts
+app.post('/users', (c) => c.render('Users/New', { ok: true }, { url: '/users/new' }))
+```
--- a/.changeset/red-trees-tease.md
+++ b/.changeset/red-trees-tease.md
@@ -4,15 +4,7 @@
 
 fix(inertia): use the Referer for non-GET requests to keep the original URL
 
-For non-GET requests, `page.url` now uses the `Referer` header (falling
-back to `c.req.url`) instead of `c.req.url` to keep the original URL.
-
-feat(inertia): add an option to override `page.url`
-
-`c.render(component, props, options)` now accepts an optional third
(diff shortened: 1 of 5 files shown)


packages/inertia/src/index.ts
```diff
@@ -473,8 +473,17 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
...
+      component: string,
+      propsInput: Record<string, unknown> = {},
+      options: RenderOptions = {}
+    ) => {
+      // Use the Referer for non-GET requests to keep the original URL.
+      // Override with options.url if provided.
+      const url =
+        options.url === undefined
+          ? new URL(c.req.method === 'GET' ? c.req.url : (c.req.header('Referer') ?? c.req.url))
+          : new URL(options.url, c.req.url)
```
ashunar0: @nkfr26 
When the `Referer` header is an empty string or a malformed value (e.g. a POST from a site with `Referrer-Policy: no-referrer`, or a broken value sent via curl), `??` only falls back on `undefined`, so `new URL()` will throw and the middleware will return 500.

I considered `URL.canParse()`, but this package's `engines` is `node: >=18.0.0` and `URL.canParse` was added in Node 18.17.0, so it may not always be available. Guarding with try/catch seems safer:

```ts
const referer = c.req.header('Referer')
let refererUrl = c.req.url
if (referer) {
  try {
    new URL(referer)
    refererUrl = referer
  } catch {
    // keep c.req.url
  }
}
```

.changeset/red-trees-tease.md
```diff
@@ -0,0 +1,18 @@
+---
+'@hono/inertia': minor
+---
+
+fix(inertia): use the Referer for non-GET requests to keep the original URL
```
ashunar0, line 5: @nkfr26 
Since there may be users relying on the previous `page.url` value (which was `c.req.url` for non-GET requests), it might help to mark this as a behavior change in the changeset body so readers of the CHANGELOG notice it.

Example:

```
**Behavior change**: For non-GET requests, `page.url` previously used
`c.req.url` (e.g. the POST target). It now uses the `Referer` header
(falling back to `c.req.url` when unavailable). If you were relying on
the old value, pass the explicit `options.url` instead.
```

Releasing this as `minor` is fine since the package is still `0.x` — this is just a suggestion about the wording.

packages/inertia/src/index.test.ts
```diff
@@ -194,6 +194,71 @@ describe('inertia', () => {
     })
   })
 
+  describe('page.url resolution', () => {
```
ashunar0, line 197: @nkfr26 
This block is well organized and easy to read 👍

nit: a couple more cases would help document the intended design:

1. When both `options.url` and `Referer` are present → `options.url` wins (documents the precedence).
2. When `options.url` is an absolute URL (e.g. `https://example.com/foo`) → only the pathname is kept, resulting in `/foo`.

Both should already work as intended from the implementation, but having tests would prevent future regressions and make the design intent clearer to readers.

ashunar0 (commented):
@nkfr26 
Thanks for the PR! Great investigation following discussion #477 — really appreciate the thoroughness 🙌

One thing I'd like to flag for @yusukebe:

This PR bundles two changes:
1. Changing the default `page.url` for non-GET requests to use `Referer` (a behavior change to existing behavior).
2. Adding the `c.render(component, props, { url })` third-argument override.

The original consensus on Issue #1989 was about (2) only — (1) was a great additional find by @nkfr26 based on discussion #477. @yusukebe, could you decide whether to land both together as `minor`, or split them? Personally I think they're closely related and shipping together is fine.

Other minor points are left as inline comments. Please take a look when you get a chance!

.changeset/red-trees-tease.md
yusukebe, line 1: You can include multiple changeset files in a single PR. I think this is about the same point (page.url), so you don't have to split this PR into two, and you can add two changesets.

packages/inertia/src/index.ts
yusukebe, line 1: This PR introduces a new feature. So, can you add a description of it to the README?

.changeset/red-trees-tease.md
yusukebe: I've written here https://github.com/honojs/middleware/pull/1991#discussion_r3466658659 . You can include another changeset!

nkfr26:
@ashunar0 @yusukebe 
Thank you for your review. I have made all the requested changes!

yusukebe (approved):
LGTM!

yusukebe:
@nkfr26 Nice!

Hey @ashunar0 , can you review this again?

ashunar0 (approved):
I think this looks good too! LGTM 🚀

yusukebe:
@nkfr26 @ashunar0 Thank you!! Let's go with this.

PR #2000, opened 2026-06-24
chore: add dependabot cooldown and disable install scripts

yusukebe:
To avoid supply chain attacks, add `cooldown` for the dependabot and disable installing scripts.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,12 +3,16 @@ version: 2
 updates:
   - package-ecosystem: github-actions
     directory: /
+    cooldown:
+      default-days: 7
     schedule:
       interval: weekly
 
   - package-ecosystem: npm
     allow:
       - dependency-type: development
+    cooldown:
+      default-days: 7
     directory: /
     groups:
       otel:
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,5 @@
+enableScripts: false
+
 nmHoistingLimits: workspaces
 
 nodeLinker: node-modules


.github/dependabot.yml
```diff
@@ -3,12 +3,16 @@ version: 2
 updates:
   - package-ecosystem: github-actions
     directory: /
+    cooldown:
+      default-days: 7
```
yusukebe, line 7: I don't know the best number but 7 days looks ok.

yusukebe:
Hey @BarryThePenguin ! Can you review this?

.github/dependabot.yml
BarryThePenguin: Sounds good 👍🏻

BarryThePenguin (approved):
This is a sensible idea 👍🏻

BlankParticle:
not entirely related to this pr, but is there any reason we are using yarn here, I feel like pnpm would be a pretty good choice in a monorepo and their sensible defaults for security

BarryThePenguin:
I don't think there are any reasons to stick with yarn. There is some release plumbing for changesets that is yarn specific, but that's about all.

There have been attempts to move to bun, mainly because https://github.com/honojs/hono uses it.

I'm not aware of any issues with using yarn. I don't have any preference, but I'm happy to discuss it, maybe in a new issue?

BlankParticle:
I myself don't have any preference either, but I think using what everyone else seem to be using is a good idea overall
will make an issue

yusukebe:
@BarryThePenguin Thanks!

@BlankParticle Let's discuss at https://github.com/honojs/middleware/issues/2001

PR #2007, opened 2026-06-27
test(inertia): add test for invalid url input

nkfr26:
The corresponding PR: https://github.com/honojs/middleware/pull/1991

I've added tests, as the tests for the implementation added later were missing.

**_Also, please let me know if I need to add myself to the Authors section in the README — I'll take care of it. I'm also planning to continue contributing (with at least two more feature proposals to come, btw)._**

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/inertia/src/index.test.ts
+++ b/packages/inertia/src/index.test.ts
@@ -265,8 +265,8 @@ describe('inertia', () => {
     it('falls back to c.req.url when options.url is unparseable on non-GET requests', async () => {
       const app = new Hono()
       app.use(inertia({ version: 'v1' }))
-      // '#' is unusable as a URL because it's just a URI fragment
-      app.post('/users', (c) => c.render('Users/New', {}, { url: '#' }))
+      // new URL('http://', 'http://localhost') throws an error because the host is empty
+      app.post('/users', (c) => c.render('Users/New', {}, { url: 'http://' }))
 
       const res = await app.request('/users', {
         method: 'POST',
--- a/packages/inertia/src/index.test.ts
+++ b/packages/inertia/src/index.test.ts
@@ -244,7 +244,40 @@ describe('inertia', () => {
       expect(body.url).toBe('/users')
     })
 
-    it('overrides page.url via options.url on a non-GET request', async () => {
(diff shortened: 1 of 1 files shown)


yusukebe:
Hey @ashunar0, can you review this?

@nkfr26 @ashunar0 

> Also, please let me know if I need to add myself to the Authors section in the README

Maybe my explanation is not enough. In this middleware repo, "Author" means the author has responsibility for the middleware and must be a reviewer for it. This means the author is stronger than just a contributor. For now, I think we should keep @nkfr26 as a contributor. If @nkfr26 really wants to be an author, though, we can add @nkfr26 to the README.

ashunar0:
@nkfr26 Thanks for backfilling the tests for the `catch` branch!

The first test (`Referer: 'invalid-url'`) looks good — `new URL('invalid-url')` throws, so it exercises the `catch` branch as intended 👍

However, the second test (`options.url: '#'`) doesn't actually hit the `catch` branch. `new URL('#', 'http://localhost/users')` does **not** throw — the WHATWG URL parser resolves it against the base to `http://localhost/users#` (pathname `/users`, empty search). So `body.url` ends up as `/users` via the `try` success path, not the fallback. The comment `// '#' is unusable as a URL because it's just a URI fragment` is therefore not quite accurate.

To actually exercise the `catch` branch, an absolute URL with malformed syntax would work better. For example:

```ts
app.post('/users', (c) => c.render('Users/New', {}, { url: 'http://[bad' }))
```

`new URL('http://[bad', c.req.url)` throws, so this would reliably route through the fallback.

nkfr26:
@ashunar0 
I apologize for the late reply. I thought relative paths would be better, but you are right that they do not throw errors. As the simplest example, I have modified it to pass http://, where the host is empty.

ashunar0:
Thanks for the update! Both fallback paths are properly covered now. LGTM from my side!

@yusukebe could you take a look when you have a chance? 🙏

yusukebe (approved):
LGTM!

yusukebe:
@nkfr26 @ashunar0 Thank you! Merging.

PR #2008, opened 2026-06-27
fix(standard-validator): preserve literal unions in RPC input type and fix hook data type

### LINKED ISSUE #1990: sValidator hook might have the wrong type

@hono/standard-validator

0.2.2

4.12.26

Node v26.3.1

Here's what the current implementation looks like
https://github.com/honojs/middleware/blob/a401fee0e0d3c20b4a0b84349043b82932d9571f/packages/standard-validator/src/index.ts#L138-L153

I suppose the question is about whether `data: value`(input) or `data: result`(output) is passed to the hook.

A couple comments on the original PR talk about the type of `data`.
@muningis made a couple arguments in favor of passing the original data to the hook.
https://github.com/honojs/middleware/pull/887#discussion_r1943184875

So if `value` (input) is being passed to the hook,
then shouldn't hook's parameter definition be `InferInput` instead of `InferOutput`?
```ts
hook?: Hook<StandardSchemaV1.Infer???<Schema>, ...
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

MathurAditya724:
## Summary

This PR makes two type-level fixes to `@hono/standard-validator`:

### 1. Preserve literal union types in RPC client input

The current `sValidator()` maps all input property values through `ValidationTargets[K][K2]`, which widens literal unions like `z.enum(["asc", "desc"])` to `string | string[]` for query targets. This loses autocomplete in the Hono RPC client (`hc`).

**Before:**
```ts
const schema = z.object({ sortBy: z.enum(["name", "type"]) })
sValidator("query", schema)
// RPC client sees: sortBy?: string | string[]
```

**After:**
```ts
// RPC client sees: sortBy?: "name" | "type"
```

This is achieved by porting the `InferInput` utility type from `@hono/zod-validator` (`packages/zod-validator/src/utils.ts`). The utility selectively preserves literal string unions and optional unions while falling back to wire-level types (`string | string[]` for query, `string` for param/header/cookie) for non-literal values. This ensures coercion schemas like `z.coerce.number()` still correctly show `string | string[]` for query targets.

Additionally, the `In extends ValidationTargets[K]` check is now wrapped in tuple brackets (`[In] extends [ValidationTargets[K]]`) to prevent distributive conditional type behavior, matching `@hono/zod-validator`'s approach.

### 2. Fix hook `data` type to use `InferInput` instead of `InferOutput`

Fixes #1990

The `hook` parameter was typed as `Hook<StandardSchemaV1.InferOutput<Schema>, ...>`, but at runtime the hook receives `value` (the raw input), not `result.value` (the validated output):

```ts
// Both success and failure branches pass `value` (input), not `result.value` (output)
result.issues
  ? { data: value, error: result.issues, success: false, target }
  : { data: value, success: true, target }
```

Changed to `Hook<StandardSchemaV1.InferInput<Schema>, ...>` to match the runtime behavior.

## Changes

- **New file**: `packages/standard-validator/src/utils.ts` — `InferInput`, `IsLiteralUnion`, and helper types (ported from `packages/zod-validator/src/utils.ts`)
- **Modified**: `packages/standard-validator/src/index.ts` — uses `InferInput` for the `I` type parameter and fixes hook type

## Testing

- All 33 existing tests pass
- `tsc --noEmit` passes with zero errors
- Existing type-level assertions (including the `querySortSchema` enum test expecting `{ order: 'asc' | 'desc' }`) continue to pass
- No runtime code changes — type-only fix

### DIFF
--- a/.changeset/standard-validator-infer-input.md
+++ b/.changeset/standard-validator-infer-input.md
@@ -1,0 +1,8 @@
+---
+'@hono/standard-validator': patch
+---
+
+fix(standard-validator): preserve literal union types in RPC input and fix hook data type
+
+- Port `InferInput` utility from `@hono/zod-validator` to preserve literal union types (e.g., `z.enum(["asc", "desc"])`) in the RPC client's input type, while still falling back to wire-level types (`string | string[]` for query, `string` for param/header/cookie) for non-literal values.
+- Fix hook `data` type to use `InferInput` instead of `InferOutput`, matching the runtime behavior where the raw input value is passed to the hook (fixes #1990).
--- a/packages/standard-validator/src/index.ts
+++ b/packages/standard-validator/src/index.ts
@@ -3,6 +3,7 @@ import type { Context, Env, Input, TypedResponse, ValidationTargets } from 'hono
 import type { Handler } from 'hono/types'
 import { validator } from 'hono/validator'
 import { sanitizeIssues } from './sanitize-issues'
+import type { InferInput } from './utils'
 
 type HasUndefined<T> = undefined extends T ? true : false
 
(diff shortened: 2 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@MathurAditya724 Thanks Aditya!

PR #2013, opened 2026-06-30
fix(standard-validator): use `InferInput` from `hono/validator` instead of a local copy

brilliantnz:
## Summary

Re-export the `InferInput` type from `@hono/standard-validator` so consumers can import it directly.

### Checklist

- [x] Changeset added (`patch`)
- [x] Run tests
- [x] `yarn changeset` and push


yusukebe:
Hi @brilliantnz 

Regarding this issue, I think it's better to do the following:

1. Bump the version of `hono` to `4.11.2` in the `peerDependencies` of standard-validator 
2. Encourage the user to use `InferInput` exported from `hono`'s `hono/validator`

packages/standard-validator/src/index.ts
```diff
@@ -160,5 +160,6 @@ const sValidator = <
     return (result as { value: StandardSchemaV1.InferOutput<Schema> }).value
   })
 
-export type { Hook }
+// Prefer import `InferInput` from `hono/validator` directly (hono >= 4.11.2)
+export type { Hook, InferInput }
```
yusukebe: If you set `>=4.11.2` for hono in the peerDependencies, I think you don't have to export `InferInput` here.

brilliantnz: sorry i forgot to remove that

yusukebe (approved):
LGTM!

yusukebe:
Hey @brilliantnz 

I updated the branch to use `InferInput` from `hono/vadalidator` completely. Looks good. Thanks!

PR #2014, opened 2026-07-05
feat(oidc-auth): opt-in client_secret_post via setClientAuth (fixes #1992 for consumers who enable it)

### LINKED ISSUE #1992: @hono/oidc-auth no longer supports Google Auth with version 1.8.3

@hono/oidc-auth

1.8.3

4.12.26

Cloudflare Workers

Use `oidcAuthMiddleware` with the Google Auth Platform

Normal login behaviour.

### What do you see instead?

`OAuth2Error: [invalid_client] The OAuth client was not found.`

### Additional information

This issue is caused by a confluence of problems:
1. Google's OAuth implementation uses certain characters (`.`, `-`, and `_`) within the Client ID and Client Secret which the OAuth standard requires to be %-encoded. However, HTTP does *not* require this, and indeed Google Auth will not accept %-encoded values, considering them to be different strings and therefore causing the `invalid_client` error.
2. The `oauth4webapi` library made a breaking change in version 3 which causes these strings to be %-encoded in order to conform with the spec, with no simple way to override this. Others have pointed out that the library no longer works with Google Auth, but the library maintainer is adamant about strictly following the spec: https://github.com/panva/oauth4webapi/issues/165
3. This issue is seemingly limited to the `client_secret_basic` mechanism, which the middleware uses exclusively, again with no simple way to change this to, for example, `client_secret_post` which is not affected.
4. The middleware moved to `oauth4webapi` major version 3 in its own patch version 1.8.3 expecting that any breaking changes would not be user-facing.

My suggestion would therefore be:
1. For now, revert to `oauth4webapi` version 2 in the 1.8.x branch. Then:
2. In a future version 1.9, implement a custom `ClientAuth` that uses the previous URL-encoding scheme, un-breaking the change in the dependency, Or:
3. In a future version 2.0, use the default `ClientAuth`, but allow the user to specify their own which is passed through to the library, providing an escape hatch for any future similar issues.

momomuchu:
Fixes #1992 for consumers who opt in.

`oauth4webapi` v3 percent-encodes the client secret for `client_secret_basic` per RFC 6749 Appendix B. Google does not decode that and returns `invalid_client` when the secret contains `-`, `_` or `.` (this includes Google's `GOCSPX-` secrets). It is the same failure mode the oauth4webapi author documents for this case (panva/oauth4webapi#165).

Adds `setClientAuth(c, clientAuth)` so consumers can opt into `client_secret_post` (or any other oauth4webapi client auth method) for the token endpoint requests (authorization-code exchange, refresh, revocation). Also adds `setClient(c, client)` to override the OAuth2 client metadata directly. The default stays `client_secret_basic`, so existing consumers are unaffected; only clients whose secret contains `-`, `_` or `.` need to call `setClientAuth(c, oauth2.ClientSecretPost(env.OIDC_CLIENT_SECRET))`.

Documented `setClientAuth` in the README, including the caveat that it must run on every request (not just the callback route) since refresh/revocation also read it. Added a regression test driving the `client_secret_post` path through `setClientAuth`, plus a test asserting the default stays `client_secret_basic`.

### DIFF
--- a/.changeset/oidc-auth-client-secret-post.md
+++ b/.changeset/oidc-auth-client-secret-post.md
@@ -1,0 +1,5 @@
+---
+'@hono/oidc-auth': patch
+---
+
+Fix Google login failing with `OAuth2Error: [invalid_client] The OAuth client was not found`. `oauth4webapi` v3 percent-encodes the client id/secret for `client_secret_basic` per RFC 6749 Appendix B, but Google does not decode that encoding and rejects the resulting credentials when they contain `-`, `_` or `.`. Switch the token endpoint requests (authorization code exchange, refresh, revocation) to `client_secret_post`, which sends the client secret unencoded in the request body.
--- a/.changeset/oidc-auth-client-secret-post.md
+++ b/.changeset/oidc-auth-client-secret-post.md
@@ -1,5 +1,5 @@
 ---
-'@hono/oidc-auth': patch
+'@hono/oidc-auth': minor
 ---
 
 Fix Google login failing with `OAuth2Error: [invalid_client] The OAuth client was not found`. `oauth4webapi` v3 percent-encodes the client id/secret for `client_secret_basic` per RFC 6749 Appendix B, but Google does not decode that encoding and rejects the resulting credentials when they contain `-`, `_` or `.`.
--- a/.changeset/oidc-auth-client-secret-post.md
+++ b/.changeset/oidc-auth-client-secret-post.md
@@ -2,4 +2,6 @@
 '@hono/oidc-auth': patch
 ---
 
-Fix Google login failing with `OAuth2Error: [invalid_client] The OAuth client was not found`. `oauth4webapi` v3 percent-encodes the client id/secret for `client_secret_basic` per RFC 6749 Appendix B, but Google does not decode that encoding and rejects the resulting credentials when they contain `-`, `_` or `.`. Switch the token endpoint requests (authorization code exchange, refresh, revocation) to `client_secret_post`, which sends the client secret unencoded in the request body.
+Fix Google login failing with `OAuth2Error: [invalid_client] The OAuth client was not found`. `oauth4webapi` v3 percent-encodes the client id/secret for `client_secret_basic` per RFC 6749 Appendix B, but Google does not decode that encoding and rejects the resulting credentials when they contain `-`, `_` or `.`.
+
+Add `setClientAuth()` to let consumers opt in to `client_secret_post` for the token endpoint requests (authorization code exchange, refresh, revocation), which sends the client secret unencoded in the request body. Also add `setClient()` to set the OAuth2 client metadata directly. The default `token_endpoint_auth_method` remains `client_secret_basic`, so existing consumers are unaffected; only clients whose secret contains `-`, `_` or `.` need to call `setClientAuth(c, oauth2.ClientSecretPost(env.OIDC_CLIENT_SECRET))`.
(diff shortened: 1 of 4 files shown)


packages/oidc-auth/src/index.ts
```diff
@@ -184,7 +184,7 @@ export const getClient = (c: Context): oauth2.Client => {
     client = {
       client_id: env.OIDC_CLIENT_ID,
       client_secret: env.OIDC_CLIENT_SECRET,
-      token_endpoint_auth_method: 'client_secret_basic',
+      token_endpoint_auth_method: 'client_secret_post',
```
BarryThePenguin: Instead of changing `token_endpoint_auth_method`, I'd recommend exporting a new `setClient` function

```ts
export const setClient = (c: Context, client: oauth2.Client) => {
  c.set('oidcClient', client)
}
```

This is something that consumers can already do today.

packages/oidc-auth/src/index.ts
```diff
@@ -235,7 +235,7 @@ export const getAuth = async (c: Context): Promise<OidcAuth | null> => {
       const response = await oauth2.refreshTokenGrantRequest(
         as,
         client,
-        oauth2.ClientSecretBasic(env.OIDC_CLIENT_SECRET),
+        oauth2.ClientSecretPost(env.OIDC_CLIENT_SECRET),
```
BarryThePenguin: Similar to a new `setClient` function, you could also introduce a new `setClientAuth` function

```ts
export const setClientAuth = (c: Context, clientAuth: oauth2.ClientAuth) => {
  c.set('clientAuth', clientAuth)
}
```

then this becomes

```suggestion
        c.get('clientAuth') ?? oauth2.ClientSecretBasic(env.OIDC_CLIENT_SECRET),
```

I think this last suggestion is the only real missing piece

momomuchu:
Done. Added setClientAuth(c, clientAuth) and setClient(c, client), reverted the default back to client_secret_basic, and switched the call sites to c.get('clientAuth') ?? oauth2.ClientSecretBasic(env.OIDC_CLIENT_SECRET), so client_secret_post is opt-in and existing consumers are unchanged. Documented setClientAuth in the README with a note that Google/GOCSPX- secrets need to opt in, and added explicit return types so the isolatedDeclarations typecheck passes. Updated the test to drive the client_secret_post path through setClientAuth.

BarryThePenguin (approved):
Thank you, it's a larger change, but I think it's the right direction to take. 👍🏻

momomuchu:
Added a test covering `setClient()`'s override path (the line codecov flagged), so patch coverage should be green now. Thanks for the review.

.changeset/oidc-auth-client-secret-post.md
```diff
@@ -0,0 +1,7 @@
+---
+'@hono/oidc-auth': patch
```
yusukebe: This introduces a new feature. Please replace `patch` with `minor`.

packages/oidc-auth/src/index.ts
```diff
@@ -31,6 +31,7 @@ declare module 'hono' {
     oidcAuth: OidcAuth | null
     oidcAuthJwt: string
     oidcClaimsHook?: OidcClaimsHook
+    clientAuth?: oauth2.ClientAuth
```
yusukebe: This should be added with an `oidc` prefix, like `oidcClientAuth`.

momomuchu:
Done. Changeset bumped to `minor`, and the context key renamed to `oidcClientAuth` to match the `oidc` prefix on the other keys (`oidcAuth`/`oidcAuthJwt`/`oidcClaimsHook`). Updated `setClientAuth` and the `c.get()` call sites.

yusukebe (approved):
LGTM!

yusukebe:
@momomuchu Thank you!

Issue #2021, opened 2026-07-09
[@hono/zod-openapi] .onError() set after .basePath() is ignored


tokiya-takai:
### Which middleware has the bug?

@hono/zod-openapi

### What version of the middleware?

1.4.0

### What version of Hono are you using?

4.12.28

### What runtime/platform is your app running on? (with version if possible)

Node.js (v20.19.6)

### What steps can reproduce the bug?

This surfaced while migrating an existing plain `Hono` app to `OpenAPIHono` to expose an OpenAPI spec: moving a `route()`-mounted sub-app from `Hono` to `OpenAPIHono` silently changed which　`onError` handled its thrown errors.

```js
import { Hono } from 'hono'
import { OpenAPIHono } from '@hono/zod-openapi'

const sub = new OpenAPIHono().basePath('/v2')
sub.onError(() => new Response('from sub'))
sub.get('/boom', () => { throw new Error('boom') })

const app = new Hono()
app.onError(() => new Response('from parent'))
app.route('/', sub)

const res = await app.request('/v2/boom')
console.log(await res.text()) // from parent ← Oops!
```

### What is the expected behavior?

The sub-app's own onError should handle the thrown error, so `from sub` should be output.

### What do you see instead?

The parent's `onError` handles it: output `from parent`. The `.onError()` registered on the `basePath()` derived instance is dropped — `sub.errorHandler` remains the built-in default.

aetosdios27:
Reproduced this, root cause confirmed. `OpenAPIHono.basePath()`wraps `super.basePath(path)`'s result in a second new instance via property spread. Core's `basePath()` already does this correctly via a private `#clone()`, but arrow-function class fields like `onError` (and similarly `notFound`, `fetch`, `request`) are bound to `this` at construction time spreading them into a new object copies stale bindings to the discarded intermediate instance. Route-registration methods happen to "work" only because the wrapper and intermediate instance share the same underlying `routes` array reference.

Planning to fix by having `basePath()` reuse the correctly-cloned instance directly (e.g. via `Object.setPrototypeOf`) rather than reconstructing a second `OpenAPIHono`, so existing arrow-field bindings stay valid. 

Happy to open a PR, let me know if there's a preferred approach given the type-narrowing implications of skipping the constructor path.

tokiya-takai:
[...]

PR #2031, opened 2026-07-12
fix(zod-openapi): preserve onError/notFound bindings set after basePath()

tokiya-takai:
Issue: #2021 

## Summary

`OpenAPIHono.basePath()` dropped any `.onError()` / `.notFound()` registered on the returned instance
after the `basePath()` call. The handler silently had no effect, and the parent app's handler
ran instead. This surfaced while migrating a plain Hono app to OpenAPIHono: a route()-mounted
sub-app quietly changed which onError handled its thrown errors.

## Reproduction

```ts
import { Hono } from 'hono'
import { OpenAPIHono } from '@hono/zod-openapi'

const sub = new OpenAPIHono().basePath('/v2')
sub.onError(() => new Response('from sub'))
sub.get('/boom', () => {
  throw new Error('boom')
})

const app = new Hono()
app.onError(() => new Response('from parent'))
app.route('/', sub)

const res = await app.request('/v2/boom')
console.log(await res.text()) // 'from parent' ← expected 'from sub'
```

## Root cause

`OpenAPIHono.basePath()` rebuilt the instance by spreading the plain Hono clone returned by `super.basePath()`:

```ts
return new OpenAPIHono({ ...(super.basePath(path) as any), defaultHook: this.defaultHook })
```

`onError`, `notFound`, `request`, and `fetch` are arrow-function instance fields, so they are own-enumerable properties permanently bound to the instance they were created on.
`super.basePath()` returns a throwaway clone, and spreading it copies those clone-bound methods
into the constructor options — where Object.assign(this, options) then overwrites the freshly constructed (and correctly bound) methods on the returned instance.

As a result, `sub.onError(h)` mutated errorHandler on the discarded clone, not on the returned
sub (sub.errorHandler stayed the built-in default). Route registration only appeared to work
because the returned instance and the clone share the same underlying router / routes. When
parent.route('/', sub) reads sub.errorHandler as a plain property, it still sees the default,
so it does not wrap the sub-app's handlers — and the parent's onError ends up handling the thrown error.

## Fix

Construct a properly initialized OpenAPIHono (so its own arrow-field methods stay bound to it) and transplant only the routing state — router, routes, getPath, errorHandler, _basePath — from the clone, instead of spreading the clone into the constructor. Carrying errorHandler over preserves the "onError() before basePath()" ordering that core Hono's clone already captured.

## Tests

- Should fire onError() registered after basePath() — regression guard for this issue (fails on main, passes with the fix).
- Should fire onError() registered before basePath() — guards the opposite ordering so the errorHandler transplant can't regress.
- notFound() set after basePath() targets the returned instance — a same-instance contract test. Note: route() propagates errorHandler but not notFoundHandler, so this documents the binding rather than acting as a regression guard (it passes with or without the fix; commented inline).

Verified locally: full @hono/zod-openapi suite passes (81 tests, type-check clean), ESLint clean, Prettier clean.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/zod-openapi-basepath-onerror.md
+++ b/.changeset/zod-openapi-basepath-onerror.md
@@ -1,0 +1,10 @@
+---
+'@hono/zod-openapi': patch
+---
+
+fix(zod-openapi): preserve `onError()`/`notFound()` set after `basePath()`
+
+`OpenAPIHono.basePath()` used to rebuild the instance by spreading the plain `Hono` clone returned by `super.basePath()`.
+That spread copied the clone's arrow-function instance fields (`onError`, `notFound`, `request`, `fetch`),
+which stay bound to the discarded clone — so calling `.onError()` (or `.notFound()`) *after* `.basePath()` mutated the throwaway clone instead of the returned app and silently had no effect (the parent's handler ran instead).
+`basePath()` now transplants only the routing state onto a properly constructed `OpenAPIHono`, keeping its own correctly-bound methods. Fixes #2021.
--- a/packages/zod-openapi/eslint-suppressions.json
+++ b/packages/zod-openapi/eslint-suppressions.json
@@ -46,7 +46,7 @@
       "count": 2
     },
     "@typescript-eslint/no-unsafe-argument": {
-      "count": 13
+      "count": 12
     },
     "@typescript-eslint/no-unsafe-assignment": {
       "count": 1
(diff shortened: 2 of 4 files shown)


yusukebe (approved):
LGTM!

CatLover01:
@tokiya-takai Seems like it's still happening for `notFound`, it's dropped completly. I'm on `1.5.1`.
To reproduce, you can simply setup `notFound`, then `basePath`, and it's dropped from there.

PR #2042, opened 2026-07-16
feat(zod-openapi): support schemas from any Standard JSON Schema library

### LINKED ISSUE #1763: Standard (json) schema support for @hono/(zod-)openapi

@hono/zod-openapi

Supporting Standard [Standard JSON schema](https://standardschema.dev/json-schema) instead of just Zod would be wonderful.

gusanthiago:
Hey everyone, that PR should be close #1763.

## Summary

`@hono/zod-openapi` takes Zod and nothing else today. This makes it take schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema) — ArkType, Valibot once wrapped with `toStandardJsonSchema()`, and Zod 4, which ships it natively. A route can mix them.

```ts

app.openapi(
  createRoute({
    method: 'post',
    path: '/users',
    request: {
      body: {
        required: true,
        content: { 'application/json': { schema: type({ name: 'string' }) } },
      },
    },
    responses: { 200: { description: 'Created' } },
  }),
  (c) => c.json({ name: c.req.valid('json').name }, 200)
)
```

## Solution

`@asteasolutions/zod-to-openapi` reads Zod internals, so a foreign schema cannot go to it directly. Non-Zod schemas are hidden inside a Zod carrier (`z.string().openapi(jsonSchema)`) — the generator merges `.openapi()` metadata over whatever it generated, so the output comes entirely from `~standard.jsonSchema`. Zod schemas are never converted and behave exactly as before.

## Test

`pnpm test` in `packages/zod-openapi` — 123 pass, the 111 existing ones untouched. The 12 new tests cover ArkType and mixed routes: validation, parameters, input vs output, response headers and sub-apps.

## Caveats

- **`doc()` and OpenAPI 3.0.** ArkType implements only the JSON Schema drafts, so those schemas fall back to `draft-07` and constructs 3.0 lacks pass through unconverted. `doc31()` needs no fallback.
- **Validation hooks.** `result.error` is an array of Standard Schema issues for non-Zod schemas, while `Hook` still calls it a `ZodError`. Widening it would break every hook reading `.issues`, so I left it — tell me if you would rather take the break.

Both are documented under Limitations in the README.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/tender-pots-attack.md
+++ b/.changeset/tender-pots-attack.md
@@ -2,4 +2,4 @@
 '@hono/zod-openapi': patch
 ---
 
-Accept schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema), such as ArkType, alongside Zod. A route can mix libraries: non-Zod schemas are validated through their Standard Schema interface and documented with the JSON Schema they generate, while Zod schemas keep going through `@asteasolutions/zod-to-openapi` untouched.
+Accept schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema), such as ArkType, alongside Zod. A route can mix libraries: non-Zod schemas are validated through their Standard Schema interface and documented with the JSON Schema they generate, while Zod schemas keep going through `@asteasolutions/zod-to-openapi` untouched. Configure which JSON Schema dialect to request via `jsonSchemaTargets` when a library only supports specific targets.
--- a/.changeset/tender-pots-attack.md
+++ b/.changeset/tender-pots-attack.md
@@ -1,5 +1,5 @@
 ---
-'@hono/zod-openapi': minor
+'@hono/zod-openapi': patch
 ---
 
 Accept schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema), such as ArkType, alongside Zod. A route can mix libraries: non-Zod schemas are validated through their Standard Schema interface and documented with the JSON Schema they generate, while Zod schemas keep going through `@asteasolutions/zod-to-openapi` untouched.
--- a/.changeset/tender-pots-attack.md
+++ b/.changeset/tender-pots-attack.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': minor
+---
+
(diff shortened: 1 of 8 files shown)


KTrain5169 (changes requested):
A solution to the fallback problem would be letting the user specify the version of JSON Schema they wish to use with their validators, that way if a user uses ArkType they can choose to use a JSON Schema version that ArkType supports. Similarly, if a Zod user wants to use an OpenAPI schema version, it will auto-convert too. This is the approach I found makes the most sense while making my own wrapper for openapi3-ts to work with Standard JSON Schema

KTrain5169:
Also FYI I think this package should be split from `@hono/zod-openapi` and be its own thing.

gusanthiago:
Hey @KTrain5169, thanks for the review.

About the points you touched:

* **Split from `@hono/zod-openapi`** — I agree, but that would make this PR much larger. I can open a new issue for that and also create a PR if necessary. WDYT?
* **Configurable JSON Schema target** — Adjusted so you can set it: https://github.com/honojs/middleware/pull/2042/changes#diff-c1564bb93a89a1adb74a4c1ec0b7668986816ea9eec13d8e9e9441c6db359d76R39

Can you check again?

KTrain5169:
I think at this point you may as well rename the PR since you're not really touching the code for the zod-openapi anymore (in fact you shouldn't be doing that in this PR at all now I think)

gusanthiago:
@KTrain5169 I edited my comment and PR, wdyt?

KTrain5169:
Nah, I wouldn't put this functionality into @hono/zod-openapi

yusukebe:
@KTrain5169 

Do you have ideas for good naming for the package?

KTrain5169:
> @KTrain5169
> 
> Do you have ideas for good naming for the package?

Either `@hono/standard-openapi` or `@hono/standard-json-openapi` works.
My only concern is that `standard-openapi` implies any Standard Schema validator as opposed to any Standard JSON Schema validator. They are two separate interfaces and should be made clear on the package README.

gusanthiago:
In the end, I think it's better to have a separate package. I would open another PR (and close this) so we can address that to have `@hono/standard-json-openapi` properly.

Is it okay for you @KTrain5169 @yusukebe ?

KTrain5169:
> In the end, I think it's better to have a separate package. I would open another PR (and close this) so we can address that to have `@hono/standard-json-openapi` properly.
> 
> Is it okay for you @KTrain5169 @yusukebe ?

works for me

gusanthiago:
I'm closing this PR due to the creation of another -> https://github.com/honojs/middleware/pull/2044

PR #2043, opened 2026-07-17
docs(inertial): remove `Experimental`

yusukebe:

- [ ] Add tests
- [ ] Run tests
- [ ] `pnpm changeset` at the top of this repo and push the changeset

### DIFF
--- a/packages/inertia/README.md
+++ b/packages/inertia/README.md
@@ -1,8 +1,5 @@
 # @hono/inertia
 
-> [!WARNING]
-> **Experimental.**
-
 **The SPA, minus the API.** [Inertia.js](https://inertiajs.com) on [Hono](https://hono.dev) — render React, Vue, or Svelte straight from your routes and skip the REST endpoints, the client side router, and the data fetching layer entirely.
 
 ```ts


yusukebe:
Hey @ashunar0 @nkfr26 

I'd like to make this not experimental. But do you think there are blockers?

ashunar0:
Hey @yusukebe

Nothing in particular from my side — I think we're good to go! Happy to see it graduating from experimental 🎉

nkfr26:
Hi, @yusukebe 
I don't have the time to work on these right now, but there are three features I would like to add in the future:
- [Shared Data](https://inertiajs.com/docs/v3/data-props/shared-data)
- A utility that accepts `StandardSchemaV1.Issue[]` and converts it into the error format expected by Inertia
I have defined my own version [here](https://github.com/nkfr26/hono-inertia-app/blob/main/app/lib/utils.ts). I might also need to consider [Error Bags](https://inertiajs.com/docs/v3/the-basics/validation#error-bags).
- An option to exclude directories or files with specific prefixes from the type, similar to [TanStack Router's routeFileIgnorePrefix](https://tanstack.com/router/v1/docs/api/file-based-routing#routefileignoreprefix)

I think it would be fine to add these after removing experimental, but what do you think?

yusukebe:
@ashunar0 @nkfr26 

Thank you! Removing experimental does not mean all features should be implemented. We can develop them after it. So let's merge this.

> I don't have the time to work on these right now, but there are three features I would like to add in the future

Thank you for the feature list. You can implement it at your own pace.

PR #2044, opened 2026-07-21
feat(standard-json-openapi): add package for Standard JSON Schema OpenAPI

### LINKED ISSUE #1763: Standard (json) schema support for @hono/(zod-)openapi

gusanthiago:
Hey everyone, this supersedes #2042 based on the review there, and should close #1763.

## Summary

`@hono/zod-openapi` takes Zod and nothing else today. This new package, `@hono/standard-json-openapi`, takes schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema) — ArkType, Valibot once wrapped with `toStandardJsonSchema()`, and Zod 4, which ships it natively. A route can mix them.

```ts
import { OpenAPIHono, createRoute } from '@hono/standard-json-openapi'

const app = new OpenAPIHono()

app.openapi(
  createRoute({
    method: 'post',
    path: '/users',
    request: {
      body: {
        required: true,
        content: { 'application/json': { schema: type({ name: 'string' }) } },
      },
    },
    responses: { 200: { description: 'Created' } },
  }),
  (c) => c.json({ name: c.req.valid('json').name }, 200)
)
```

## Solution

`@asteasolutions/zod-to-openapi` reads Zod internals, so a foreign schema cannot go to it directly. Non-Zod schemas are hidden inside a Zod carrier (`z.string().openapi(jsonSchema)`) — the generator merges `.openapi()` metadata over whatever it generated, so the output comes entirely from `~standard.jsonSchema`. Zod schemas are never converted and behave exactly as before.

## Test

`pnpm test` in `packages/standard-json-openapi` — 18 pass. They cover ArkType and mixed routes: validation, parameters, input vs output, response headers and sub-apps.

- [x] Add tests
- [x] Run tests


packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+app.doc31('/doc', {
+  openapi: '3.1.0',
+  info: { version: '1.0.0', title: 'My API' },
+})
+```
+
+## Limitations
+
+- **Zod is still a peer dependency.** Document generation uses `@asteasolutions/zod-to-openapi` under the hood; non-Zod schemas are embedded as Zod carrier metadata.
```
KTrain5169: can we try not to have a dependency on Zod-specific stuff? this would make the bundle heavier than necessary

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,139 @@
...
+- **`app.doc()` and OpenAPI 3.0.** ArkType implements only the JSON Schema drafts, so those schemas fall back to `draft-07` and constructs 3.0 lacks pass through unconverted. Prefer `jsonSchemaTargets` when you want to pick the dialect yourself. `app.doc31()` needs no fallback.
+- **Validation hooks.** `result.error` is an array of Standard Schema issues for non-Zod schemas, while `Hook` still calls it a `ZodError`. Widening it would break every hook reading `.issues`, so it stays as-is for now.
```
KTrain5169: I think we'll be fine breaking it somewhat, I'll probably need to check though.

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,139 @@
...
+app.doc31('/doc', {
+  openapi: '3.1.0',
+  info: { version: '1.0.0', title: 'My API' },
+})
+```
+
+## Limitations
+
+- **Zod is still a peer dependency.** Document generation uses `@asteasolutions/zod-to-openapi` under the hood; non-Zod schemas are embedded as Zod carrier metadata.
+- **`app.doc()` and OpenAPI 3.0.** ArkType implements only the JSON Schema drafts, so those schemas fall back to `draft-07` and constructs 3.0 lacks pass through unconverted. Prefer `jsonSchemaTargets` when you want to pick the dialect yourself. `app.doc31()` needs no fallback.
```
KTrain5169: is this line outdated?

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,139 @@
...
+        'application/json': {
+          schema: z.object({ id: z.string() }).openapi('User'),
+        },
+      },
+    },
+  },
+})
+```
+
+Valibot does not implement the interface directly — wrap its schemas with [`toStandardJsonSchema()`](https://valibot.dev/api/toStandardJsonSchema/) from `@valibot/to-json-schema` first.
```
KTrain5169: Also denote this for Zod 4 Mini

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,139 @@
+# Standard JSON OpenAPI Hono
+
+`@hono/zod-openapi` takes Zod and nothing else. This package takes schemas from any library implementing [Standard JSON Schema](https://standardschema.dev/json-schema) — ArkType, Valibot once wrapped with `toStandardJsonSchema()`, and Zod 4, which ships it natively. A route can mix them.
```
KTrain5169: not really a proper README tbh

packages/standard-json-openapi/src/index.ts
```diff
@@ -0,0 +1,936 @@
...
+} from '@asteasolutions/zod-to-openapi'
+import {
+  OpenAPIRegistry,
+  OpenApiGeneratorV3,
+  OpenApiGeneratorV31,
+  extendZodWithOpenApi,
+  getOpenApiMetadata,
+} from '@asteasolutions/zod-to-openapi'
+import { sValidator } from '@hono/standard-validator'
+import { zValidator } from '@hono/zod-validator'
```
KTrain5169: prefer not using @hono/zod-validator

KTrain5169 (changes requested):
Please just don't try and force Zod into the package here

packages/standard-json-openapi/README.md
gusanthiago: I agree, I removed from runtime

packages/standard-json-openapi/src/index.ts
gusanthiago: Dropped. Everything validates through sValidator now

packages/standard-json-openapi/README.md
gusanthiago: Taken — Hook's failure branch is now error: readonly StandardSchemaV1.Issue[] (plus data, matching what sValidator actually passes) instead of ZodError. Nothing to break in practice since the package is unreleased.

packages/standard-json-openapi/README.md
gusanthiago: Not outdated, but badly placed, I adjusted it now

gusanthiago:
Hey @KTrain5169, thanks for the review

* I dropped Zod as a runtime dependency so we can decouple that part;
* Simplified some logic around the module to avoid excessive implementations;
* Adjusted other points about README.md.

Let me know if it is okay now

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,214 @@
...
+        output: (options?: { target?: string }) =>
+          z.toJSONSchema(schema, { io: 'output', target: options?.target as never }),
+      },
+    },
+  })
+
+const User = withJSONSchema(z.object({ name: z.string() }))
+```
+
+A schema that reaches a route without `~standard.jsonSchema` is treated as a literal JSON Schema object and is neither converted nor validated, so wrap before you use it.
```
KTrain5169:
```suggestion
Any validation library listed on the [Standard JSON Schema spec page](https://standardschema.dev/json-schema) is supported here.

Some libraries (notably Valibot and Zod Mini) do **not** expose the `~standard.jsonSchema` interface by default for bundle size reasons. You'll need to follow their instructions on how to expose the interface (in the case of Valibot, this requires you to wrap your schema with the `toStandardJsonSchema` function from `@valibot/to-json-schema`, in Zod Mini's case, you can just wrap your schema with `z.toJSONSchema`).

A schema that reaches a route without `~standard.jsonSchema` is treated as a literal JSON Schema object and is neither converted nor validated, so wrap before you use it.
```

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,214 @@
...
+  return c.json({ id: name }, 200)
+})
+
+app.doc31('/doc', {
+  openapi: '3.1.0',
+  info: { version: '1.0.0', title: 'My API' },
+})
+```
+
+Requests are validated against the same schema that documents them, and `c.req.valid()` is typed from it.
```
KTrain5169: Examples with Zod & Valibot are also needed (this goes for pretty much all examples listed on this README

packages/standard-json-openapi/README.md
```diff
@@ -0,0 +1,214 @@
...
+    }
+  },
+})
+```
+
+A hook passed as the third argument to `app.openapi()` takes precedence over `defaultHook`.
+
+## Limitations
+
+- **Request header arrays are documentation only.** `request.headers` accepts an array of schemas so several can be described, but only a single schema is validated at runtime.
```
KTrain5169: Any reason why this is the case?

packages/standard-json-openapi/CHANGELOG.md
KTrain5169, line 1: File should not exist, should instead be created by the changeset upon first release

packages/standard-json-openapi/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@hono/standard-json-openapi",
+  "version": "0.1.0",
```
KTrain5169:
```suggestion
  "version": "0.0.0",
```

Version will be bumped by changesets

packages/standard-json-openapi/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "description": "A Hono class that generates OpenAPI docs from any Standard JSON Schema library.",
+  "type": "module",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.cts",
```
KTrain5169, line 8: @yusukebe Should middlewares be published as ESM-only?

packages/standard-json-openapi/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "url": "git+https://github.com/honojs/middleware.git",
+    "directory": "packages/standard-json-openapi"
+  },
+  "homepage": "https://github.com/honojs/middleware",
+  "author": "Gustavo Santos Thiago <gusanthiagodv@gmail.com> (https://github.com/gusanthiago)",
+  "peerDependencies": {
+    "@standard-schema/spec": "^1.0.0",
+    "hono": ">=4.10.0"
+  },
+  "devDependencies": {
```
KTrain5169, line 47: seems that valibot is not listed, tests should ideally have a test case for the 3 major schema libraries (Zod, Valibot & ArkType, Zod Mini optional)

.changeset/standard-json-openapi.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hono/standard-json-openapi': minor
+---
+
+Initial release: OpenAPI for Hono using any library that implements [Standard JSON Schema](https://standardschema.dev/json-schema) (ArkType and Zod 4 natively, Valibot via `toStandardJsonSchema()`, Zod Mini via a small wrapper, and others). No schema library is a dependency of this package. Configure JSON Schema dialects with `jsonSchemaTargets` when a library only supports specific targets.
```
KTrain5169:
```suggestion
Initial release: OpenAPI for Hono using any library that implements [Standard JSON Schema](https://standardschema.dev/json-schema) (ArkType and Zod 4 natively, Valibot via `toStandardJsonSchema()`, Zod Mini via `z.toJSONSchema`, and others). No schema library is a dependency of this package. Configure JSON Schema dialects with `jsonSchemaTargets` when a library only supports specific targets.
```

KTrain5169 (changes requested):
I haven't been able to take a proper look at the code myself yet as I don't really have time to experiment to the level required to have a thorough code review

packages/standard-json-openapi/package.json
gusanthiago: Done

packages/standard-json-openapi/CHANGELOG.md
gusanthiago: Done

packages/standard-json-openapi/README.md
gusanthiago: Done

.changeset/standard-json-openapi.md
gusanthiago: Done

packages/standard-json-openapi/package.json
gusanthiago: Added `@validbot` + `@valibot/to-json-schema`, and added a test case to other schema libraries, thanks for bringing this up

packages/standard-json-openapi/README.md
gusanthiago: Added

packages/standard-json-openapi/README.md
gusanthiago: No reason; that was gap — the array never reached a validator, so nothing was validated. Now each schema validates. Only the typing stays limited, since Hono has a single header target. Adjusted the Limitation and added a test.

gusanthiago:
@KTrain5169

I did adjustments mentioned in the PR. Let me know if I need to work on anything more. 

Thanks for the review

packages/standard-json-openapi/src/generator.ts
```diff
@@ -0,0 +1,350 @@
...
+    const inputKey = stableStringify(input)
+
+    const outputName = this.#nameFor(name, output, outputKey)
+    const inputName =
+      inputKey === outputKey ? outputName : this.#nameFor(`${name}Input`, input, inputKey)
+
+    this.#named.set(schema, { input: inputName, output: outputName })
+  }
+
+  schema(schema: RouteSchema, kind: SchemaKind): SchemaObject | ReferenceObject {
```
kamaal111, line 209: The README says a registered schema becomes a `$ref` at *every* use. It does at the top level, but
not when it is nested inside another schema.

```ts
const Address = app.openAPIRegistry.register('Address', z.object({ street: z.string() }))

// route 1 — the registered schema itself
app.openapi(createRoute({
  method: 'get', path: '/addresses',
  responses: { 200: { description: 'ok', content: { 'application/json': { schema: Address } } } },
}), handler)

// route 2 — the same instance, one level down
app.openapi(createRoute({
  method: 'get', path: '/users',
  responses: { 200: { description: 'ok', content: { 'application/json': {
    schema: z.object({ id: z.string(), address: Address }),
  } } } },
}), handler)
```

Generated:

```jsonc
// GET /addresses  ->  a reference, as documented
{ "$ref": "#/components/schemas/Address" }

// GET /users  ->  the same schema, copied out instead of referenced
{
  "type": "object",
  "properties": {
    "id": { "type": "string" },
    "address": {
      "type": "object",
      "properties": { "street": { "type": "string" } },
      "required": ["street"],
      "additionalProperties": false
    }
  },
  "required": ["id", "address"],
  "additionalProperties": false
}
```

packages/standard-json-openapi/src/generator.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    return Object.entries(properties).map(([name, property]) => ({
+      name,
+      required: required.has(name),
+      ...(typeof property.description === 'string' ? { description: property.description } : {}),
+      schema: property as SchemaObject,
+    }))
+  }
+
+  #convert(schema: StandardOpenAPISchema, kind: SchemaKind): JSONSchema {
```
kamaal111, line 261: Registering one plain object emits two components. No routes needed to see it:

```ts
const app = new OpenAPIHono()
app.openAPIRegistry.register('Address', z.object({ street: z.string() }))

app.getOpenAPI31Document({ openapi: '3.1.0', info: { title: 'demo', version: '1' } })
```

```jsonc
"components": {
  "schemas": {
    "Address": {
      "type": "object",
      "properties": { "street": { "type": "string" } },
      "required": ["street"],
      "additionalProperties": false   // <- the only difference
    },
    "AddressInput": {
      "type": "object",
      "properties": { "street": { "type": "string" } },
      "required": ["street"]
    }
  }
}
```

`AddressInput` comes from line 204: `name()` converts the schema twice, once for input and once for
output, compares the two, and creates a second component named `` `${name}Input` `` when they
differ.

`z.object({ street: z.string() })` has no transform in it, so nothing should differ. The two
conversions differ only because Zod adds `additionalProperties: false` on the output side, and the
comparison at line 204 reads that as a real difference.

It shows up on a route that both accepts and returns the schema:

```jsonc
"requestBody": { ... "schema": { "$ref": "#/components/schemas/AddressInput" } },
"responses":   { "200": { ... "schema": { "$ref": "#/components/schemas/Address" } } }
```

so the document says you post an `AddressInput` and get back an `Address`, when they're the same
shape. Used only in a response, `AddressInput` is generated but never referenced by anything.

Deleting that keyword in `#convert` before the comparison gives a single `Address`.

kamaal111 (commented):
@gusanthiago great work on this initiative, I have been thinking about this idea myself too and have created my own middleware that I use in my projects where I have caught several edge-cases and added some conveniences.
https://github.com/kamaal111/hono-standard-openapi

I think your PR is a great first step, but I do think it might be worthwhile to have some tests in this middleware to catch some subtle bugs that may be lying here.

I am happy to branch of your branch and help with adding some test cases if you would like

gusanthiago:
Hello @kamaal111, thanks for the review.

Your library looks great too; I believe we can merge this PR as soon as possible, and we can work on improving it step by step. This PR already has a good size.

I'm adjusting the points that you mentioned, and also happy to include future follow-up PRs.

Thanks :)

gusanthiago:
Done @kamaal111.

Can you take a look again?

We can also create new issues and add some follow-ups PRs to improve the package

packages/standard-json-openapi/eslint-suppressions.json
```diff
@@ -0,0 +1,16 @@
+{
+  "src/index.ts": {
```
kamaal111, line 2: As this is a new middleware, are we sure we want to ship it with ESLint suppressions that can be resolved?

packages/standard-json-openapi/src/standard-schema.test.ts
```diff
@@ -0,0 +1,1013 @@
...
+      '~standard': {
+        ...schema['~standard'],
+        jsonSchema: {
+          input: (options?: { target?: string }) =>
+            zm.toJSONSchema(schema, { io: 'input', target: options?.target as never }),
+          output: (options?: { target?: string }) =>
+            zm.toJSONSchema(schema, { io: 'output', target: options?.target as never }),
+        },
+      },
+    })
```
kamaal111, line 439: This might have been fixed in Zod now, could you verify that wrapping schemas in [z.toJSONSchema()](https://zod.dev/json-schema?id=ztojsonschema#ztojsonschema) would fix the issue you have faced?
I personally haven't faced any issues with Zod Mini

packages/standard-json-openapi/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "exactOptionalPropertyTypes": false,
+    "noImplicitOverride": false,
+    "noImplicitReturns": false,
+    "noPropertyAccessFromIndexSignature": false,
+    "noUncheckedIndexedAccess": false,
+    "noUnusedLocals": false
```
kamaal111, line 9: I think it will be worthwhile to enable these to increase the quality of this new middleware

kamaal111 (commented):
@gusanthiago I think the scope of this PR is way too big, and it looks like [zod-openapi](https://github.com/honojs/middleware/tree/main/packages/zod-openapi) was copied almost "as is" without considering complexities that it brings with it and whether we actually need all of it if we are using [Standard Schema](https://standardschema.dev/schema) and [Standard JSON Schema](https://standardschema.dev/json-schema).

What do you think of decreasing the scope for the initial merge to `OpenAPIHono` + `createRoute` + `openapi()` + `doc/doc31` + `route()`?

Let’s get this reduced scope in without disabling type check rules or disabling eslint rules, and it also makes it easier to review.

gusanthiago:
Makes sense, I trimmed this PR down to the core path: `OpenAPIHono, createRoute, openapi(), doc/doc31`, and `route()`.

Dropped `$, defineOpenAPIRoute`, and `openapiRoutes`. Also removed the ESLint suppressions and turned the strict tsconfig flags back on. 

Happy to add the rest in follow-up PRs @kamaal111

packages/standard-json-openapi/eslint-suppressions.json
gusanthiago: Agreed, the lint issues are fixed rather than suppressed.

packages/standard-json-openapi/src/standard-schema.test.ts
gusanthiago: `z.toJSONSchema()` is enough now, dropped the custom wrapper; the test and `README` use that.

packages/standard-json-openapi/tsconfig.build.json
gusanthiago: Was enabled, the package now inherits the repo’s strict compiler options

packages/standard-json-openapi/src/generator.ts
kamaal111: I have verified that this has been addressed, thank you @gusanthiago

packages/standard-json-openapi/src/generator.ts
kamaal111: I have verified that this has been addressed, thank you

kamaal111 (approved):
Great job on this middleware @gusanthiago, looking forward to expanding its features after it gets merged 😁

gusanthiago:
Thanks @kamaal111, let me know about it, we can do together

PR #2056, opened 2026-07-29
feat(oidc-auth): add optional oidcAuthRefreshErrorHook

pdebie-amazon:
# feat(oidc-auth): add optional `oidcAuthRefreshErrorHook`

## What

Adds an optional `oidcAuthRefreshErrorHook` context variable (mirroring the
existing `oidcClaimsHook`) that `getAuth()` invokes when a refresh-token grant is
rejected by the authorization server, before it swallows the error.

```ts
c.set('oidcAuthRefreshErrorHook', (error, ctx) => {
  // error: oauth2.ResponseBodyError | oauth2.WWWAuthenticateChallengeError
  if (error instanceof ResponseBodyError && error.error === 'invalid_grant') {
    metrics.increment('session.refresh.revoked')
    ctx.header('x-session-dead', '1') // rides the redirect getAuth triggers
  }
})
```

## Why

Today `getAuth()` catches a rejected refresh (`oauth2.ResponseBodyError` /
`oauth2.WWWAuthenticateChallengeError`), deletes the session cookie, and returns
`null` — which drives a redirect to re-authenticate. The **reason** for the
failure is discarded, so an application cannot:

- log or emit metrics on *why* sessions fail to refresh (revoked vs rate-limited
  vs server error), or
- branch on the specific OAuth2 error — e.g. distinguish a permanent
  `invalid_grant` (the refresh token is revoked/expired and will never succeed)
  from a transient `too_many_requests` (the IdP is throttling and a retry would
  work).

## Tests

Two new tests in `packages/oidc-auth/src/index.test.ts`:

1. When the refresh-token grant is rejected (`processRefreshTokenResponse` throws
   `ResponseBodyError`), the hook fires with that error, and a header the hook
   sets on the context rides the resulting redirect; the session cookie is
   cleared.
2. The hook is optional — a rejected refresh with no hook registered still
   redirects (unchanged behaviour).

`pnpm test` (42 pass), `pnpm typecheck`, `pnpm lint`, and `pnpm build` all pass.

### DIFF
--- a/.changeset/oidc-auth-refresh-error-hook.md
+++ b/.changeset/oidc-auth-refresh-error-hook.md
@@ -1,0 +1,5 @@
+---
+"@hono/oidc-auth": minor
+---
+
+Add an optional `oidcAuthRefreshErrorHook`. `getAuth()` otherwise swallows a rejected refresh-token grant (it catches the `oauth4webapi` `ResponseBodyError` / `WWWAuthenticateChallengeError`, deletes the session cookie, and returns `null`), so callers cannot observe *why* a refresh failed. The hook is invoked with the OAuth2 error and the context before the cookie is deleted, letting applications log/meter refresh failures or branch on the specific error (e.g. distinguish a permanent `invalid_grant` from a transient `too_many_requests`), and — since it runs before the redirect and receives the context — optionally set a response header that rides the redirect. The hook is optional and defaults to the current behaviour, so this is fully backward compatible.
--- a/packages/oidc-auth/README.md
+++ b/packages/oidc-auth/README.md
@@ -149,6 +149,28 @@ Note:
 If explicit logout is not required, the logout handler can be omitted.
 If the middleware is applied to the callback URL, the default callback handling in the middleware can be used, so the explicit callback handling is not required.
 
+## Handling session refresh errors
+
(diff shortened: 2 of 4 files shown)


yusukebe (commented):
Can you document `oidcAuthRefreshErrorHook` in the README?

yusukebe (commented):
I pushed a small change. Errors thrown by the hook are now swallowed, so a buggy hook cannot turn the re-login redirect into a 500.

pdebie-amazon:
> I pushed a small change. Errors thrown by the hook are now swallowed, so a buggy hook cannot turn the re-login redirect into a 500.

Hmm are you sure about this change? For me I'd be pretty annoyed by these errors being silently discarded, since then I might not get the behavior I want (changing the response) and have no way of knowing why (there's not even any logging). I'd have to work around this silent catching by manually catching, setting some other context var, then after the OIDC middlewar ran checking if that context var is set and manually throwing a 500. OTOH, if the desired behavior is "continue as normal if there's an error" it's very easy to manually ` catch (e) {}` and hook into my logging lib.

It's a new hook so we don't need "discard errors silently" for backwards compatibility.

(existing `oidcClaimsHook` doesn't silently swallow errors either, and I can work around the silent error discarding if necessary, so just let me know)

yusukebe:
Hi @pdebie-amazon 

Thank you for the comment. I will consider it, but you are right. We can simply wrap it with `catch (e) {}` manually to handle the error. I reverted it. So, one thing: can you update the README to add this feature?

pdebie-amazon:
Added a README section documenting the hook. Thanks for the fast reply!

yusukebe (approved):
LGTM!

yusukebe:
@pdebie-amazon Thank you!! Let's go.

PR #2069, opened 2026-08-06
fix(zod-openapi): export local z binding so OpenAPI patch is not tree-shaken

### LINKED ISSUE #2051: [@hono/zod-openapi] pass-through `z` re-export breaks `.openapi()` under esbuild code splitting
### What version of the middleware are you using?

`@hono/zod-openapi@1.4.0` (also verified present in `1.5.1`)

### What is the problem?

`@hono/zod-openapi` re-exports zod's `z` unchanged after patching it:

```ts
// packages/zod-openapi/src/index.ts
import { z } from 'zod'
// ...
extendZodWithOpenApi(z)
export { extendZodWithOpenApi, z }
```

`.openapi()` exists only as a side effect of that `extendZodWithOpenApi(z)` call, but the `z` export is a **pass-through** — the imported binding is re-exported untouched. So a bundler is free to resolve `import { z } from '@hono/zod-openapi'` directly to zod's own export and drop the import edge to this module, since nothing else from it is used.

With esbuild `splitting: true` that is exactly what happens. Schema-building code lands in a chunk whose only import is zod's chunk, so it evaluates **before** the module that applies the patch, and every schema construction throws:

```
TypeError: external_exports.string(...).optional(...).openapi is not a function
    at dist/chunks/chunk-LBD3OXP2.js:11:47
```

The failure mode is nasty in practice:

- It is **runtime-only**. Types check, unit tests pass (test runners evaluate the unbundled source, where the import edge is real), and the bundle builds successfully.
- It is **per-module non-deterministic**. In our app 7 of 19 entry points threw; which modules break depends on how esbuild happens to group chunks.
- Adding a bare `import '@hono/zod-openapi'` to the affected modules does **not** reliably fix it — esbuild honoured that edge for one of our schema modules and dropped it for another *in the same build*.

For anyone bundling a serverless handler this is a total outage that passes every static gate.

Importing `z` from `@hono/zod-openapi` should guarantee the OpenAPI extension is applied, regardless of how a bundler treats re-exports.

### Steps to reproduce

Any esbuild build with `splitting: true`, multiple entry points, and schemas built from `z` imported from `@hono/zod-openapi`, then `await import()` each emitted entry.

### Proposed fix

Export `z` as a **local binding** instead of a pass-through, so consumers genuinely depend on this module having been evaluated:

```diff
-import { z } from 'zod'
+import { z as zodBase } from 'zod'
 // ...
-extendZodWithOpenApi(z)
-export { extendZodWithOpenApi, z }
+extendZodWithOpenApi(zodBase)
+// Deliberately a local const, not a re-export of zod's `z`: a pass-through
+// re-export lets bundlers resolve `z` straight to zod and drop the edge to this
+// module, skipping the patch above.
+const z = zodBase
+export { extendZodWithOpenApi, z }
```

Behaviourally identical (same object, `extendZodWithOpenApi` is already idempotent) and type-identical, but the binding now only exists here, so the module cannot be elided.

### Verification

Measured on a real 19-entry-point app, not reasoned about:

| configuration | result |
|---|---|
| `splitting: true`, unmodified | **7 of 19 entries throw on import**, identical across 3 clean builds |
| \+ the patch above, nothing else | **19/19 import clean**, across 3 clean builds |

Applied via `pnpm patch` to the published `dist/index.js` (`const __zPatched = z; export { … __zPatched as z }`) with no other workaround in place. Generated OpenAPI document byte-identical either way (43,847 B, 14 paths, 5 component schemas).

Happy to open a PR with the change and a changeset.

edenbuilds:
## Summary
- `@hono/zod-openapi` called `extendZodWithOpenApi(z)` then re-exported the same zod import binding.
- With esbuild `splitting: true`, `import { z } from '@hono/zod-openapi'` can resolve straight to `zod`, dropping this module — so schemas throw `...openapi is not a function` (issue #2051).
- Export a local `const z = zod` (still after the patch) so the side effect is not elidable.
- Follow-up: avoid importing type `z` alongside that value export (TS2395/TS2440), and annotate the export for `--isolatedDeclarations`.

## Test plan
- [x] `pnpm --filter @hono/zod-openapi typecheck`
- [x] `pnpm --filter @hono/zod-openapi build`
- [x] Changeset for `@hono/zod-openapi`

Fixes #2051

### DIFF
--- a/.changeset/zod-openapi-local-z-export.md
+++ b/.changeset/zod-openapi-local-z-export.md
@@ -1,0 +1,13 @@
+---
+"@hono/zod-openapi": patch
+---
+
+fix(zod-openapi): keep the `z` import edge so `.openapi()` survives bundling
+
+Re-exporting zod's `z` as a pass-through let esbuild code-splitting resolve
+`import { z } from '@hono/zod-openapi'` straight to zod and drop the edge to the
+module that runs `extendZodWithOpenApi(z)`, so schemas could be built before the
+patch applied and `.openapi()` was undefined at runtime. Export `z` as an alias
+declaration instead: it compiles to a binding local to the module, which keeps
+that edge, while still carrying zod's type namespace so `z.infer`, `z.ZodType`
+and the rest continue to work. Fixes #2051.
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -41,7 +41,7 @@ import { mergePath } from 'hono/utils/url'
 import type { OpenAPIObject } from 'openapi3-ts/oas30'
 import type { OpenAPIObject as OpenAPIV31bject } from 'openapi3-ts/oas31'
 import type { ZodType, ZodError } from 'zod'
-import { z } from 'zod'
+import * as zodModule from 'zod'
 import { isZod } from './zod-typeguard'
 
 type MaybePromise<T> = Promise<T> | T
@@ -913,8 +913,18 @@ export const createRoute = <P extends string, R extends Omit<RouteConfig, 'path'
   return Object.defineProperty(route, 'getRoutingPath', { enumerable: false })
 }
 
-extendZodWithOpenApi(z)
-export { extendZodWithOpenApi, z }
+// `zodModule.z`, not the `z` alias below: the alias compiles to a binding
+// declared at that point, which this call precedes.
+extendZodWithOpenApi(zodModule.z)
+// An alias declaration, not `export { z }`: re-exporting the imported binding
+// lets a bundler resolve `import { z } from '@hono/zod-openapi'` straight to zod
+// and drop the edge to this module, so the `extendZodWithOpenApi` call above can
+// evaluate after the schemas that need it — `.openapi()` is then undefined at
+// runtime under esbuild code splitting (#2051). This compiles to a binding local
+// to this module, which keeps that edge, while an alias — unlike `const z = zod`
+// — still carries zod's type namespace, so `z.infer` and friends keep working.
+export import z = zodModule.z
+export { extendZodWithOpenApi }
 
 function addBasePathToDocument(document: Record<string, any>, basePath: string) {
   const updatedPaths: Record<string, any> = {}


yusukebe:
This PR is not acceptable.

The exported `z` lost its namespace. So the following throws the error:

```tw
type User = z.infer<typeof Schema> // ✘ TS2503: Cannot find namespace 'z'
```

edenbuilds:
Thanks @yusukebe — you're right, \`export const z\` dropped the TypeScript namespace.

Fixed in the latest commit: keep the local value binding (so esbuild splitting cannot elide the \`extendZodWithOpenApi\` side effect), and merge a namespace so \`z.infer\` / \`z.input\` / \`z.output\` work again:

\`\`\`ts
type User = z.infer<typeof Schema> // OK
\`\`\`

Verified with \`tsc --strict\` against the built \`dist\` types. Happy to adjust if you'd rather take a different approach to #2051.

edenbuilds:
You were right, and the follow-up I pushed after your comment was still wrong — it restored `z.infer` by merging a hand-written namespace, which only covered three aliases. `z.ZodType`, `z.ZodString`, `z.ZodObject` and `z.core` all still broke. Sorry for the noise; I've replaced the whole thing.

The constraint is real: the value has to be a binding local to this module or the import edge disappears, and `export const z` carries only the value side. @salpers reached the same conclusion in #2051 and asked whether there was a `z.infer`-preserving approach nobody had tried. There is one — an alias declaration:

```ts
export import z = zodModule.z
```

An alias is local to the module *and* carries every meaning of its target, so both constraints hold at once. In the built package:

- `dist/index.mjs` → `var z = zodModule.z` — a local binding, so the edge to this module survives
- `dist/index.d.mts` → `import z = zodModule.z` — the namespace survives declaration bundling

The patch call takes `zodModule.z` directly, since it precedes the alias binding.

The whole change is now two lines plus a comment; the internal `z.input<…>` / `z.output<…>` type positions are untouched, because `z` is a full namespace again.

Verified against the built artifacts rather than the source — a consumer resolving `dist/index.d.mts` type-checks all of:

```ts
z.infer<typeof Schema>   z.input<…>   z.output<…>   z.TypeOf<…>
z.ZodType   z.ZodString   z.ZodObject<…>   z.core.$ZodIssue
```

Each of those fails on the `export const z` form, including your exact repro. `z.string().openapi()` is a function when imported from `dist/index.mjs`. 111 package tests and `tsc -b` pass.

If you'd still rather resolve #2051 as a README note about the bundler hazard, I'm happy to send that instead — just say the word.

yusukebe (approved):
LGTM!

yusukebe:
@edenbuilds 

Interesting! Looks good. Thanks!

PR #2070, opened 2026-08-06
feat(structured-logger)!: correct types for hooks and other libraries

BarryThePenguin:

- [x] Add tests
- [x] Run tests

Using this middleware, I noticed a handful of rough edges..

- `BaseLogger` doesn't allow libraries that use a different log signature
- Hooks aren't typed; `onRequest`, `onResponse`, `onError` do not pass `Context<Env>` so types go missing
- The default hooks don't provide much value so are nearly always overridden

After playing around with a few different approaches, I landed on the following breaking changes

- Removing the `BaseLogger` interface, it was only needed because the default hooks used it. Without the default hooks, we no longer need the interface, allowing any kind of logging library to be used.
- The default hooks used properties such as `{ method, path, status, elapsedMs, err,  }`, but I think most consumers would provide their own hooks

With those two changes, the middleware now has the correct types, with the one caveat that consumers now have to provide their own hooks.

The one feature I did add is a `skip` method to not log certain requests. I included all the changes in one PR so others can see the whole thing.

Also @gabry-ts and @jansepke, if you're interested in having a look

### DIFF
--- a/packages/structured-logger/README.md
+++ b/packages/structured-logger/README.md
@@ -28,8 +28,8 @@ app.use(requestId())
 app.use(
   structuredLogger({
     createLogger: (c) => rootLogger.child({ requestId: c.var.requestId }),
-    onRequest: (logger, c) =>
-      logger.info({ method: c.req.method, path: c.req.path }, 'incoming request'),
+    onResponse: (logger, c, elapsedMs) =>
+      logger.info({ method: c.req.method, path: c.req.path, elapsedMs }, 'request completed'),
   })
 )
 
@@ -53,8 +53,8 @@ const app = new Hono()
 app.use(
   structuredLogger({
     createLogger: (c) => rootLogger.child({ requestId: c.var.requestId }),
-    onRequest: (logger, c) =>
-      logger.info({ method: c.req.method, path: c.req.path }, 'incoming request'),
+    onResponse: (logger, c, elapsedMs) =>
+      logger.info({ method: c.req.method, path: c.req.path, elapsedMs }, 'request completed'),
   })
 )
 ```
@@ -70,14 +70,14 @@ const app = new Hono()
 app.use(
   structuredLogger({
     createLogger: () => console,
-    onRequest: (_logger, c) => console.info(c.req.method, c.req.path),
+    onResponse: (_logger, c, elapsedMs) => console.info(c.req.method, c.req.path, elapsedMs),
   })
 )
 ```
 
 ### Skipping routes
 
-Pass a `skip` function to opt specific requests out of logging entirely. Called before `createLogger`, so skipped requests incur no logger allocation:
+Pass a `skip` function to suppress hooks for specific requests. The logger is still created and available on `c.var` — only `onRequest`, `onResponse`, and `onError` are skipped:
 
 ```typescript
 import { Hono } from 'hono'
@@ -89,7 +89,7 @@ app.use(
   structuredLogger({
     createLogger: () => console,
     skip: (c) => c.req.path === '/health',
-    onRequest: (_logger, c) => console.info(c.req.method, c.req.path),
+    onResponse: (_logger, c, elapsedMs) => console.info(c.req.method, c.req.path, elapsedMs),
   })
 )
 ```
@@ -156,8 +156,8 @@ app.use(
   structuredLogger({
     createLogger: () => myLogger,
     contextKey: 'log',
-    onRequest: (logger, c) =>
-      logger.info({ method: c.req.method, path: c.req.path }, 'incoming request'),
+    onResponse: (logger, c, elapsedMs) =>
+      logger.info({ method: c.req.method, path: c.req.path, elapsedMs }, 'request completed'),
   })
 )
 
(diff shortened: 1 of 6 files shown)


gabry-ts:
Thanks for digging into this, and for the ping.

The hook typing problem is real and I should have caught it. Passing `Context<Env>` through to the hooks is clearly right. Same with dropping `BaseLogger`, honestly. I only wrote that interface so the defaults had something to call, and it ended up locking out any library whose signature doesn't look like pino's. Winston being the obvious casualty.

`StructuredLoggerEnv` is a better answer than what's in the README today. I'd just reverted an example that implied you could get `c.var.logger` typed from the middleware alone, which of course you can't, so an exported Env type to declare on the app fixes the thing I was papering over.

Where I'm hesitant is dropping the default hooks. Install the middleware, configure nothing, and you now get a logger on the context and no output at all. No error, no warning, just silence. You're right that most people override the defaults, but the defaults are also what makes this useful in the first five minutes, and "silent unless configured" is a rough default for a package named structured-logger. The README opener still promises automatic response time measurement too, which would need a rewrite.

Could we keep them and let people pass `onResponse: false` to opt out? Same escape hatch, without silence being what you get for free.

`skip` I like, and calling it before `createLogger` is the right order.

Curious what @jansepke thinks, and it's @yusukebe's call on the major bump either way.

BarryThePenguin:
Yeah, agreed. I didn't want to drop the defaults either. Maybe the better change would be to make them required?

If we wanted to keep the defaults, we'd need to reintroduce some `BaseLogger` equivalent to make them work, which was one of the reasons for making the change in the first place. We can't have one without the other.

gabry-ts:
Required works for me, and it's better than the opt-out I suggested. The thing I cared about was silence you don't notice, and required turns that into a compile error instead. Good trade.

Only tweak I'd consider: require `onResponse` and leave the other two optional, so someone who just wants an access log isn't writing two empty functions to get one. And yeah, no reason to drag `BaseLogger` back in.

BarryThePenguin:
Done 👍🏻 making `onResponse` required is a good idea

packages/structured-logger/README.md
```diff
@@ -138,60 +169,69 @@ app.get('/', (c) => {
...
 
-## API
+`c.var.logger` (or the custom key) will then be typed as `pino.Logger` throughout the app.
 
-### `structuredLogger(options)`
+To get typed access to other context variables inside `structuredLogger`, set the type argument with your app's env type:
 
-Returns a Hono `MiddlewareHandler`.
+```typescript
+import type { Context } from 'hono'
```
yusukebe: This line is not needed!

packages/structured-logger/src/index.ts
```diff
@@ -3,99 +3,83 @@
...
-    onResponse = defaultOnResponse,
-    onError = defaultOnError,
-  } = options
+export function structuredLogger<E extends Env = {}, L = unknown, K extends string = DefaultKey>(
+  options: StructuredLoggerOptions<E, L, K>
+): MiddlewareHandler<E> {
+  const { createLogger, contextKey = defaultKey, skip, onRequest, onResponse, onError } = options
 
   return async (c, next) => {
+    if (skip && skip(c)) {
```
yusukebe: When `skip` returns true, the middleware never calls `c.set`, but `StructuredLoggerEnv` types `c.var[contextKey]` as always present. So this type-checks but crashes at runtime:

```ts
// skip: (c) => c.req.path === '/health'
app.get('/health', (c) => {
  c.var.logger.info('ok') // TypeError → 500
})
```

packages/structured-logger/src/index.ts
```diff
@@ -36,7 +36,7 @@ export interface StructuredLoggerOptions<
   skip?: (c: Context<E>) => boolean
 
   /** Called after logger creation, before handler execution. */
-  onRequest?: (logger: L, c: LoggedContext<E, L, K>) => void | Promise<void>
+  onRequest: (logger: L, c: LoggedContext<E, L, K>) => void | Promise<void>
```
yusukebe: `onResponse` should be required instead of `onRequest,` right? As @gabry-ts said: https://github.com/honojs/middleware/pull/2070#issuecomment-5242579401

packages/structured-logger/README.md
```diff
@@ -187,6 +195,8 @@ type AppEnv = { Variables: { tenantId: string } }
 app.use(
   structuredLogger<AppEnv>({
```
yusukebe: This code throws a type error:

```ts
import pino from 'pino'

const rootLogger = pino()

type AppEnv = { Variables: { tenantId: string } }

structuredLogger<AppEnv>({
  createLogger: (c) => rootLogger.child({ tenantId: c.var.tenantId }),
  onRequest: (logger, c) => logger.info({ path: c.req.path }, 'incoming request'),
  // error TS18046: 'logger' is of type 'unknown'
})
```

packages/structured-logger/src/index.ts
BarryThePenguin: Ah, good catch! I've changed it so `c.var[contextKey]` is always set

packages/structured-logger/src/index.ts
BarryThePenguin: 🙈 I've changed that now

packages/structured-logger/README.md
BarryThePenguin: You're right, I've added a test to verify it 👍🏻

BarryThePenguin:
Thank you for the thorough review!

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Great! I'll merge it and release a new version. Thanks!

PR #2075, opened 2026-08-07
fix(mcp): notify onclose when standalone SSE stream aborts

### LINKED ISSUE #1762: StreamableHTTPTransport: SSE stream abort should notify via onclose (or new callback)

edenbuilds:
## Summary
- Call `onclose` when the standalone GET SSE stream (`_GET_stream`) aborts, so apps can detect client disconnects without relying on DELETE.
- Deduplicate notifications so `close()` after an abort does not fire `onclose` twice.

Fixes #1762

## Test plan
- [x] Unit test: cancel SSE body → `onclose` once; subsequent `close()` does not double-fire
- [x] Existing `onsessionclosed` DELETE tests still pass

Made with [Cursor](https://cursor.com)


yusukebe:
Hello @MathurAditya724 ! Can you review this?

packages/mcp/src/streamable-http.ts
```diff
@@ -634,7 +648,9 @@ export class StreamableHTTPTransport implements Transport {
 
     // Clear any pending responses
     this.#requestResponseMap.clear()
-    this.onclose?.()
+    // Standalone SSE abort may already have notified; still notify when there
+    // was no GET stream (e.g. DELETE after JSON-only traffic).
+    this.#notifyClose()
```
MathurAditya724: Thanks for working on this — exposing a dropped standalone SSE connection would be useful.

I’m not sure `onclose` is the right hook here, though. In the MCP Streamable HTTP model, a disconnected SSE stream can be resumed, and the session itself remains valid until the client sends `DELETE` or the server terminates it. The official SDK keeps that distinction: an aborted GET stream only cleans up the stream mapping, while `onclose` is reserved for `transport.close()`.

With this change, a temporary disconnect fires `onclose`, but a later reconnect is still allowed. The `#closedNotified` guard then suppresses the callback when the session is actually closed.

Could we add a separate callback for the standalone SSE stream disconnect instead, and keep `onclose` for terminal transport shutdown? It would also be great to add an abort → reconnect → DELETE test, especially for the resumable/event-store path.

edenbuilds: Implemented in fd6b5fde. The transport now exposes `onsessiondisconnected(sessionId)` for standalone SSE aborts while keeping `onclose` reserved for terminal `close()` / DELETE shutdown. Added coverage for abort → callback → resumable reconnect; with the source fix stashed, the test fails (0 callback calls), and with it restored the focused test passes. Package typecheck and all @hono/mcp tests pass (196/196).

MathurAditya724 (approved):

edenbuilds:
CI follow-up pushed as `67ebe5c8` (edenbuilds).

- Removed the new non-null assertions and updated the package's suppression count.
- Added coverage for errors raised by `onsessiondisconnected`.
- Verified the original regression was red before the behavior fix; current focused tests pass 67/67, full `@hono/mcp` tests pass 197/197, typecheck passes, lint has no errors, and Prettier passes.

The earlier red `autofix`/`build`/Codecov checks were attributable to the PR; this commit addresses their reported failures.

yusukebe (approved):
LGTM!

yusukebe:
@edenbuilds @MathurAditya724 Thanks!

PR #2076, opened 2026-08-07
feat(oauth-providers): add OpenStreetMap OAuth 2.0 provider

### LINKED ISSUE #2074: Add OpenStreetMap OAuth 2.0 provider

@hono/oauth-providers

Add integration of OpenStreetMap's OAuth 2.0 provider into @hono/oauth-providers.
I will try it.

smellman:
Closes #2074.

Adds `openstreetmapAuth`, exported as `@hono/oauth-providers/openstreetmap`, plus a `revokeToken` helper.

```ts
import { Hono } from 'hono'
import { openstreetmapAuth } from '@hono/oauth-providers/openstreetmap'

const app = new Hono()

app.use(
  '/openstreetmap',
  openstreetmapAuth({
    client_id: process.env.OPENSTREETMAP_ID,
    client_secret: process.env.OPENSTREETMAP_SECRET,
    scope: ['read_prefs'],
  })
)

app.get('/openstreetmap', (c) => {
  return c.json({
    token: c.get('token'),
    grantedScopes: c.get('granted-scopes'),
    user: c.get('user-openstreetmap'),
  })
})
```

## Implementation notes

OpenStreetMap runs [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper), and a few of its
choices differ from the other providers in this package. Each decision below is backed by the
[authorization server metadata](https://www.openstreetmap.org/.well-known/oauth-authorization-server)
or by the [openstreetmap-website](https://github.com/openstreetmap/openstreetmap-website) source.

**PKCE is always used, with the `S256` challenge method.** The server advertises
`code_challenge_methods_supported: ["plain", "S256"]`, and Doorkeeper requires a code verifier for
public clients. Sending the challenge unconditionally means the middleware works for both
confidential and public applications.

**No refresh token, and the access token carries no `expires_in`.** OpenStreetMap's Doorkeeper
config sets `access_token_expires_in nil` and leaves `use_refresh_token` disabled, so tokens do not
expire and no refresh token is ever issued. `token` is therefore `{ token: string }` and
`refresh-token` is not set. `revokeToken` is exported instead, since revocation is the only way to
invalidate a token.

**User details come from `GET /api/0.6/user/details.json`.** This carries much more than the OIDC
`userinfo` endpoint, but it needs the `read_prefs` scope — `api_ability.rb` has
`can :details, User if scopes.include?("read_prefs")` — so the README calls that out. The
OpenStreetMap API answers errors with a plain text body rather than JSON, so the response status is
checked before parsing (`authFlow.ts`).

**`skip_authorization` is excluded from `OpenStreetMapScope`,** and `read_email` is documented as
privileged. Both live in `PRIVILEGED_SCOPES` in `lib/oauth.rb` and are only offered to applications
registered by a site administrator, so `user-openstreetmap.email` is normally `undefined`.

## Testing

Unit tests use the existing msw setup: 11 cases covering the redirect URL and its PKCE parameters,
the custom `redirect_uri` and `state` options, CSRF protection (both a mismatched and an omitted
`state`), an invalid code, an API rejection of the token, the success path, and both revocation
outcomes.

Beyond that, the flow was run manually against the live openstreetmap.org authorization server —
authorization, token exchange, user lookup and revocation. The live response then corrected the
types: `contributor_terms.pd` and `messages` are unconditional in `_user.json.jbuilder` for a
request that gets past the `read_prefs` check, so both are required rather than optional.

One trap worth recording for anyone testing this locally: `force_ssl_in_redirect_uri` in
`doorkeeper.rb` only permits plain http for the hosts `127.0.0.1` and `::1`. A redirect URI of
`http://localhost:3000/...` is rejected at application registration time.

---

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/lucky-pandas-tickle.md
+++ b/.changeset/lucky-pandas-tickle.md
@@ -1,0 +1,7 @@
+---
+'@hono/oauth-providers': minor
+---
+
+Add an OpenStreetMap OAuth 2.0 provider, available as `@hono/oauth-providers/openstreetmap`. The
+flow always uses PKCE with the `S256` challenge method, and `revokeToken` is exported to invalidate
+tokens, which OpenStreetMap never expires on its own.
--- a/packages/oauth-providers/README.md
+++ b/packages/oauth-providers/README.md
@@ -1209,6 +1209,145 @@ app.get('/msentra/refresh', (c, next) => {
 })
 ```
 
+### OpenStreetMap
+
+```ts
+import { Hono } from 'hono'
+import { openstreetmapAuth } from '@hono/oauth-providers/openstreetmap'
+
+const app = new Hono()
+
+app.use(
+  '/openstreetmap',
+  openstreetmapAuth({
+    client_id: process.env.OPENSTREETMAP_ID,
+    client_secret: process.env.OPENSTREETMAP_SECRET,
+    scope: ['read_prefs'],
+  })
+)
+
+export default app
+```
+
+Register your application on the [OAuth 2 applications page](https://www.openstreetmap.org/oauth2/applications)
+of your OpenStreetMap account to obtain the client ID and secret.
+
+#### Parameters
+
+- `client_id`:
+  - Type: `string`.
+  - `Required`.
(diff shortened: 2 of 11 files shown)


smellman:
Test application: https://gist.github.com/smellman/536b56922ef3033e80e73284bf38fe7c

yusukebe (approved):
LGTM!

yusukebe:
Hey hey @smellman !

Looks good. Thank you for the new feature!

PR #2084, opened 2026-08-13
revert(zod-openapi): export local z binding

### LINKED ISSUE #2078: @hono/zod-openapi@1.5.2: generated declarations reference zodModule without importing it

kamaal111:
Fixes https://github.com/honojs/middleware/issues/2078

This reverts commit https://github.com/kamaal111/hono-middleware/commit/fc0377c5595bde3147d181e5013a3fc710175e40.

Original commit title that has been reverted:
fix(zod-openapi): export local z binding so OpenAPI patch is not tree-shaken (https://github.com/honojs/middleware/pull/2069)

This commit is being reverted because its use-case is not yet supported
by `rolldown-plugin-dts` which is a subdependency of `tsdown` in this
project. Its impact is that the zod types have decraded and is resulting
in types resulting in `any` which causes the typescript compiler to
error in strict mode

- [x]  ~~Add tests~~ not relevant
- [x] Run tests

### DIFF
--- a/.changeset/few-cooks-mix.md
+++ b/.changeset/few-cooks-mix.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Revert the local z binding export because it causes zod types to degrade to any when built with tsdown.
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -41,7 +41,7 @@ import { mergePath } from 'hono/utils/url'
 import type { OpenAPIObject } from 'openapi3-ts/oas30'
 import type { OpenAPIObject as OpenAPIV31bject } from 'openapi3-ts/oas31'
 import type { ZodType, ZodError } from 'zod'
-import * as zodModule from 'zod'
+import { z } from 'zod'
 import { isZod } from './zod-typeguard'
 
 type MaybePromise<T> = Promise<T> | T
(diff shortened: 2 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@kamaal111 Thank you for the quick fix!

PR #2087, opened 2026-08-13
fix(zod-openapi): return 415 for request bodies with undeclared content type

### LINKED ISSUE #891: [zod-openapi][security] Unexpected validation bypass when omitting content-type
Hey there,
starting from `"@hono/zod-openapi": "0.15.2"` the body validation is skipped if the content-type is not matched - [ref PR](https://github.com/honojs/middleware/pull/686). 

Personally I was very surprised by this as this means you can effectively bypass validations and invoke the handler as long as you just omit the content-type.
I was also not able to find any mentioning of this in the docs. I think this can be a significant security risk as it causes unexpected behaviour in the handler which might expect a validated payload.

I would argue that the mention PR should be reverted and the handler should fail if the content-type is not matched, but at least this should be clearly documented and have been marked as a breaking change.

<details>
  <summary>Reproduction</summary>
  
```js
import { createRoute, z, OpenAPIHono } from '@hono/zod-openapi';

const route = createRoute({
  method: 'POST',
  path: '/book',
  request: {
    // This would mitigate the issue as we enforce the usage of `application/json` content-type
    //headers: z.object({
    //  'content-type': z.literal('application/json'),
    //}),
    body: {
      content: {
        'application/json': {
          schema: z.object({
            title: z.string().startsWith('[TITLE]').min(10).max(30),
          }),
        },
      },
    },
  },
});

const app = new OpenAPIHono();

app.openapi(route, (c) => {
  const validatedBody = c.req.valid('json');

  console.log('This will only be logged when the request body passed validation, right?!', validatedBody);

  return c.json({ processed: true });
});



console.log("Case I: sending a valid request with content-type `application/json`")
/**
 * Output:
 * 
 * This will only be logged when the request body passed validation, right?! { title: '[TITLE] Hello World' }
 * { status: 200, response: { processed: true } }
 */
await app.request('/book', {
  method: 'POST',
  headers: new Headers({ 'Content-Type': 'application/json' }),
  body: JSON.stringify({ title: '[TITLE] Hello World' }),
}).then(async response => console.log({ status: response.status, response: await response.json() }));


console.log("\n\nCase II: sending a malformed request with content-type `application/json`")
/**
 * Output:
 * 
 * {
 *    status: 400,
 *    response: { success: false, error: { issues: [Array], name: 'ZodError' } }
 * }
 */
await app.request('/book', {
  method: 'POST',
  headers: new Headers({ 'Content-Type': 'application/json' }),
  body: JSON.stringify({ title: 'not valid' }),
}).then(async response => console.log({ status: response.status, response: await response.json() }));

console.log("\n\nCase III: sending a malformed request without any content-type")
/**
 * Output:
 * 
 * Case III: sending a malformed request without any content-type
 * This will only be logged when the request body passed validation, right?! {}
 * { status: 200, response: { processed: true } }
 */
await app.request('/book', {
  method: 'POST',
  body: JSON.stringify({ title: 'not valid' }),
}).then(async response => console.log({ status: response.status, response: await response.json() }));
```
</details>

With `"@hono/zod-openapi": "0.15.1"` you got the following behaviour - which I would expect:

```
Case I: sending a valid request with content-type `application/json`
This will only be logged when the request body passed validation, right?! { title: '[TITLE] Hello World' }
{ status: 200, response: { processed: true } }


Case II: sending a malformed request with content-type `application/json`
{
  status: 400,
  response: { success: false, error: { issues: [Array], name: 'ZodError' } }
}


Case III: sending a malformed request without any content-type
{
  status: 400,
  response: { success: false, error: { issues: [Array], name: 'ZodError' } }
}
``` 

starting from `"@hono/zod-openapi": "0.15.2"` the handler is invoked if the content-type is omitted:

```
Case I: sending a valid request with content-type `application/json`
This will only be logged when the request body passed validation, right?! { title: '[TITLE] Hello World' }
{ status: 200, response: { processed: true } }


Case II: sending a malformed request with content-type `application/json`
{
  status: 400,
  response: { success: false, error: { issues: [Array], name: 'ZodError' } }
}


// ‼️ This is the problem ‼️
Case III: sending a malformed request without any content-type
This will only be logged when the request body passed validation, right?! {}
{ status: 200, response: { processed: true } }
```

---

as a mitigation the content-type can be enforced by validating the incoming `content-type` header.

Thanks!

WolfieLeader:
Fixes #891.

Heads up: this could also have been fixed by defaulting `request.body.required` to `true`. I went the other way because that turns genuinely optional bodies into 400s and changes the emitted OpenAPI document. Close this if you'd rather have the simpler default.

## The bug

For an optional body, the generated middleware validates only when the `Content-Type` header matches a declared media type. Any other Content-Type skips validation and injects `{}` as the validated data. The client controls the header, so it can bypass validation on any route with an optional body.

This route is enough to reproduce it:

```ts
const route = createRoute({
  method: 'post',
  path: '/posts',
  request: {
    body: { content: { 'application/json': { schema: z.object({ id: z.number() }) } } }, // required defaults to false
  },
  responses: { 200: { description: 'ok' } },
})

const app = new OpenAPIHono()

app.openapi(route, (c) => {
  const data = c.req.valid('json') // typed as { id: number }
  return c.json({ data })
})
```

Sending the same body under different Content-Types. The `After` column is this PR:

| Body sent | `Content-Type` | Before | After |
|---|---|---|---|
| `{"id":7}` | `application/json` | 200 · `{"id":7}` | 200 · `{"id":7}` |
| `{"id":7}` | `application/vnd.api+json` | 200 · `{"id":7}` | 200 · `{"id":7}` |
| `{"id":7}` | `text/plain` | 200 · `{}` | 415 |
| `{"id":7}` | *(omitted)* | 200 · `{}` | 415 |
| `<post><id>7</id></post>` | `application/xml` | 200 · `{}` | 415 |
| `<post><id>7</id></post>` | `text/xml` | 200 · `{}` | 415 |
| *(none)* | *(omitted)* | 200 · `{}` | 200 · `{}` |
| `{"id":7}` *(required: true)* | `text/plain` | 400 | 415 |

`data` is typed `{ id: number }`, so `data.id` is a `number` to the compiler while the runtime value is `undefined`.
> XML was tested for the love of the game

That last row is the only behavior change beyond the fix itself: `required: true` with a mismatched Content-Type moves from 400 to 415. Hono's `validator()` already substitutes `{}` on a mismatch, so the schema was failing on an empty object rather than on the real problem.

## The fix

A gate runs before the body validators:

- Content-Type matches a declared media type -> validate as today.
- Content-Type matches no declared media type -> 415.
- No Content-Type, but a body was sent -> 415.
- No Content-Type and no body -> unchanged: `{}` passes through for optional bodies, and `required: true` still fails the schema with 400.

The issue thread proposed 400 for this case. This PR returns 415 because [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#section-15.5.16) defines it for exactly this situation. Switching to 400 is a one-line change if you prefer.

One gap: only JSON and form media types get a validator, so a route declaring `application/json` **and** `application/xml` still gets `{}` when sent XML. Rejecting that would break XML-only routes, which are legitimate, so it needs a call.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -875,13 +875,13 @@ function addBasePathToDocument(document: Record<string, any>, basePath: string) 
 }
 
 function isJSONContentType(contentType: string) {
-  return /^application\/([a-z-\.]+\+)?json/.test(contentType)
+  return /^application\/([a-z-\.]+\+)?json/i.test(contentType)
 }
 
 // Mirrors hono/validator's checks. Anything looser lets Hono skip parsing and validate {}.
-const jsonRequestRegex = /^application\/([a-z-\.]+\+)?json(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/
-const multipartRequestRegex = /^multipart\/form-data(;\s?boundary=[a-zA-Z0-9'"()+_,\-./:=?]+)?$/
-const urlencodedRequestRegex = /^application\/x-www-form-urlencoded(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/
+const jsonRequestRegex = /^application\/([a-z-\.]+\+)?json(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/i
+const multipartRequestRegex = /^multipart\/form-data(;\s?boundary=[a-zA-Z0-9'"()+_,\-./:=?]+)?$/i
+const urlencodedRequestRegex = /^application\/x-www-form-urlencoded(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/i
 
 function matchesMediaType(contentType: string, mediaType: string) {
   if (mediaType === '*/*') {
@@ -930,8 +930,8 @@ async function hasRequestBody(c: Context): Promise<boolean> {
 }
 
 function isFormContentType(contentType: string) {
+  const lower = contentType.toLowerCase()
   return (
-    contentType.startsWith('multipart/form-data') ||
-    contentType.startsWith('application/x-www-form-urlencoded')
+    lower.startsWith('multipart/form-data') || lower.startsWith('application/x-www-form-urlencoded')
   )
 }
--- a/packages/zod-openapi/src/index.ts
+++ b/packages/zod-openapi/src/index.ts
@@ -27,6 +27,7 @@ import type {
   TypedResponse,
   ValidationTargets,
 } from 'hono'
+import { HTTPException } from 'hono/http-exception'
 import type { H, MergePath, MergeSchemaPath } from 'hono/types'
 import type {
   ClientErrorStatusCode,
@@ -632,6 +633,21 @@ export class OpenAPIHono<
     const bodyContent = route.request?.body?.content
 
     if (bodyContent) {
+      const declaredMediaTypes = Object.keys(bodyContent)
+      const mediaTypeGate: MiddlewareHandler = async (c, next) => {
+        const contentType = c.req.header('content-type')
+        if (contentType) {
+          if (!declaredMediaTypes.some((mediaType) => matchesMediaType(contentType, mediaType))) {
+            throw new HTTPException(415, { message: 'Unsupported Media Type' })
+          }
+        } else if (c.req.raw.body !== null && (await c.req.arrayBuffer()).byteLength > 0) {
+          // Server adapters may attach an empty body stream to bodyless
+          // requests, so a non-null stream alone does not prove a body was sent.
+          throw new HTTPException(415, { message: 'Unsupported Media Type' })
+        }
+        await next()
+      }
+      validators.push(mediaTypeGate)
       for (const mediaType of Object.keys(bodyContent)) {
         if (!bodyContent[mediaType]) {
           continue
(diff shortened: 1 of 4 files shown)


packages/zod-openapi/src/index.ts
```diff
@@ -632,6 +633,21 @@ export class OpenAPIHono<
...
+            throw new HTTPException(415, { message: 'Unsupported Media Type' })
+          }
+        } else if (c.req.raw.body !== null && (await c.req.arrayBuffer()).byteLength > 0) {
+          // Server adapters may attach an empty body stream to bodyless
+          // requests, so a non-null stream alone does not prove a body was sent.
+          throw new HTTPException(415, { message: 'Unsupported Media Type' })
+        }
+        await next()
+      }
+      validators.push(mediaTypeGate)
```
yusukebe: This gate also runs for routes that only declare media types we don't validate, like `application/xml`. Those routes have nothing to bypass, so I want to keep their behavior. Can we push the gate only when the route has a JSON or form validator? Like this:

```ts
if (declaredMediaTypes.some((m) => isJSONContentType(m) || isFormContentType(m))) {
  validators.push(mediaTypeGate)
}
```

yusukebe:
Hi @WolfieLeader 

Sorry to be late!

I added a comment. Plus, the README describes the old behavior. Its example sends a body without Content-Type and shows {}, but that request returns 415 after this change. Can you update that section?

packages/zod-openapi/src/index.ts
WolfieLeader: Will check

WolfieLeader:
@yusukebe Yeah once I have everything done

WolfieLeader:
KK

WolfieLeader:
@yusukebe You're the best💪🏽💯

yusukebe (approved):
LGTM!

yusukebe:
Thank you @WolfieLeader !

PR #2103, opened 2026-08-20
fix(inertia): normalize empty props in PageProps

nkfr26:
## What

Normalize the props resolved by `PageProps`:

- Renders without props resolve to `{}`.
- When multiple handlers render the same page, each handler's prop shape becomes its own member of the resulting union.
- When the page is rendered both with and without props, the props-less variant shares the other renders' keys as optional `never`, so absent props must be accessed with `?.`.

## Problem

- Props-less renders resolved to `Record<string, never>`: accessing nonexistent props produces no error.
- When the same page is rendered with and without props by different handlers, the merged props type is `Record<string, never> | { errors: Record<string, string> }`. Because `Record<string, never>` has an index signature, `props.errors` type-checks as always present, but the prop can be `undefined` at runtime when the props-less handler runs.

## Fix

Apply an internal `NormalizeProps` helper in `src/page-props.ts`:

- renders without props → `{}`
- mixed renders (same page, with and without props) → the props-less variant is rewritten to `{ [K in keyof OtherVariants]?: never }`, keeping existing property accesses valid while forcing `?.` for values that may be absent
- renders with different prop shapes keep their own union members

The normalized result is flattened with `Simplify`.
The helper is not exported, so the public d.ts surface is unchanged.

## Changes

- `src/page-props.ts` — normalization, `Simplify` flattening, `MethodOutput` distribution
- `src/page-props.test.ts` — type tests via `PageProps` + `AppRegistry`
- `.changeset/fast-kings-reply.md` — patch changeset

Type-only change.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/fast-kings-reply.md
+++ b/.changeset/fast-kings-reply.md
@@ -1,0 +1,5 @@
+---
+'@hono/inertia': patch
+---
+
+Normalize the props resolved by `PageProps`: renders without props now resolve to `{}`, and when the same page is rendered with and without props by different handlers, the props are made optional.
--- a/.changeset/fast-kings-reply.md
+++ b/.changeset/fast-kings-reply.md
@@ -2,4 +2,4 @@
 '@hono/inertia': patch
 ---
 
-Normalize the props resolved by `PageProps`: renders without props now resolve to `{}`, and when the same page is rendered with and without props by different handlers, the props are made optional.
+Normalize the props resolved by `PageProps`: renders without props now resolve to `{}`, and when the same page is rendered with and without props by different handlers, the props-less variant shares the other renders' keys as optional `never`, so absent props must be accessed with `?.`.
--- a/packages/inertia/src/page-props.test.ts
+++ b/packages/inertia/src/page-props.test.ts
@@ -5,17 +5,13 @@ import { inertia } from './index'
 
 const _app = new Hono()
   .use(inertia())
-  .get('/with-props', (c) => c.render('WithProps', { lazy: () => Promise.resolve({ id: 0 }) }))
+  .get('/lazy-props', (c) => c.render('LazyProps', { lazy: () => Promise.resolve({ id: 0 }) }))
   .get('/without-props', (c) => c.render('WithoutProps'))
-  .post(
-    '/optional-props',
-    (c) => c.render('OptionalProps', { errors: { email: 'invalid' } }),
-    (c) => c.render('OptionalProps')
-  )
   .get(
     '/union-props',
-    (c) => c.render('UnionProps', { a: 0 }),
-    (c) => c.render('UnionProps', { b: 'string' })
+    (c) => c.render('UnionProps'),
+    (c) => c.render('UnionProps', { kind: 'ok' as const, value: 0 }),
+    (c) => c.render('UnionProps', { kind: 'ng' as const, error: 'message' })
   )
 
 declare module '@hono/inertia' {
@@ -25,19 +21,19 @@ declare module '@hono/inertia' {
 }
 
 describe('PageProps', () => {
-  it('resolves lazy prop values', () => {
-    expectTypeOf<PageProps<'WithProps'>>().toEqualTypeOf<{ lazy: { id: number } }>()
+  it('resolves lazy props', () => {
+    expectTypeOf<PageProps<'LazyProps'>>().toEqualTypeOf<{ lazy: { id: number } }>()
   })
 
-  it('collapses the props of a render without props to {}', () => {
+  it('normalizes renders without props to {}', () => {
     expectTypeOf<PageProps<'WithoutProps'>>().toEqualTypeOf<{}>()
   })
 
-  it('makes the props optional when multiple handlers render the same page with and without props', () => {
-    expectTypeOf<PageProps<'OptionalProps'>>().toEqualTypeOf<{ errors?: { email: string } }>()
-  })
-
-  it('keeps non-empty props from multiple handlers', () => {
-    expectTypeOf<PageProps<'UnionProps'>>().toEqualTypeOf<{ a: number } | { b: string }>()
+  it('normalizes union types', () => {
+    expectTypeOf<PageProps<'UnionProps'>>().toEqualTypeOf<
+      | { kind: 'ok'; value: number }
+      | { kind: 'ng'; error: string }
+      | { kind?: never; value?: never; error?: never }
+    >()
   })
 })
--- a/packages/inertia/src/page-props.test.ts
+++ b/packages/inertia/src/page-props.test.ts
@@ -8,10 +8,15 @@ const _app = new Hono()
   .get('/lazy-props', (c) => c.render('LazyProps', { lazy: () => Promise.resolve({ id: 0 }) }))
   .get('/without-props', (c) => c.render('WithoutProps'))
   .get(
-    '/union-props',
-    (c) => c.render('UnionProps'),
-    (c) => c.render('UnionProps', { kind: 'ok' as const, value: 0 }),
-    (c) => c.render('UnionProps', { kind: 'ng' as const, error: 'message' })
+    '/union-with-empty-props',
+    (c) => c.render('UnionWithEmptyProps'),
+    (c) => c.render('UnionWithEmptyProps', { kind: 'ok' as const, value: 0 }),
+    (c) => c.render('UnionWithEmptyProps', { kind: 'ng' as const, error: 'message' })
+  )
+  .get(
+    '/union-without-empty-props',
+    (c) => c.render('UnionWithoutEmptyProps', { a: 0 }),
+    (c) => c.render('UnionWithoutEmptyProps', { b: 'string' })
   )
 
 declare module '@hono/inertia' {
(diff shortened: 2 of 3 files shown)


yusukebe:
@nkfr26 Thanks!

Hey @ashunar0 ! Can you review this?

nkfr26:
@yusukebe @ashunar0  
Please let me know if anything is unclear! In particular, when the presence or absence of props is mixed, making it Partial feels destructive, or rather, tricky to handle. It does improve convenience, but I don't think it's entirely correct. Especially when there are two or more props, I feel this is undesirable. I'd love to hear your thoughts.

nkfr26:
After some thought, I've come to believe that the Union pattern is superior. I've included the implementation and comparison below. If you prefer this one, I'll make the revisions again.

### Implementation

```ts
type NormalizeProps<T> = T extends Record<string, never> ? {} : T
```

### Comparison

```ts
const _app = new Hono()
  .use(inertia())
  .get('/without-props', (c) => c.render('WithoutProps'))
  .get('/optional-props',
    (c) => c.render('OptionalProps'),
    (c) => c.render('OptionalProps', { errors: { email: 'invalid' } })
  )
```

#### **`WithoutProps`** (single handler, props-less only)

| Pattern | Resolved type |
|---|---|
| **Before the fix** | `Record<string, never>` |
| **Partial / Union** | `{}` |

Accessing a prop that does not exist:

```ts
// Before the fix: no error — the index signature of Record<string, never> resolves it to never
props.foo // compiles silently

// Partial / Union: compile error
props.foo // TS2339: Property 'foo' does not exist on type '{}'
```

#### **`OptionalProps`** (two handlers, with and without props)

| Pattern | Resolved type |
|---|---|
| **Before the fix** | `{ errors: { email: string } } \| Record<string, never>` |
| **Partial** | `{ errors?: { email: string } }` |
| **Union** | `{ errors: { email: string } } \| {}` |

Accessing `errors`:

```ts
// Before the fix: type-checks as always present, but undefined at runtime
props.errors.email // compiles, crashes when rendered without props

// Partial: direct access, handle undefined normally
props.errors?.email

// Union: narrow first, then access
'errors' in props && props.errors.email
```

Note: a prop that does not exist (e.g. `props.foo`) is already a TS2339 error in all three versions — the `{ errors }` member blocks it even before the fix.

### Discriminated union style usage

If you expect to use the props like a discriminated union, the Union pattern is the more correct choice.
With the `Partial` pattern, the same code cannot guarantee `props.error` — it remains `string | undefined`.

```ts
const _app = new Hono()
  .use(inertia())
  .get(
    '/union-props',
    (c) => c.render('UnionProps'),
    (c) => c.render('UnionProps', { kind: 'ok' as const, value: 0 }),
    (c) => c.render('UnionProps', { kind: 'ng' as const, error: 'message' })
  )
```

```ts
// When `props.kind` is 'ng', `props.error` is guaranteed to be a string
if ('kind' in props && props.kind === 'ng') { props.error }
```

### By the way

I used the `errors` prop as an example, but since in Inertia it's generally common to obtain it from `Form` or `useForm`, this may not have been a very appropriate example. https://inertiajs.com/docs/v3/the-basics/forms#slot-props

ashunar0:
@nkfr26 Sorry for the slow review. I pulled the branch locally and checked how the types actually behave in a few cases.

First, thanks for catching these two problems. The second one in particular — `props.errors.email` type-checking when a page is rendered both with and without props — leads straight to a runtime crash, so it definitely needs fixing. I fully agree with the direction.

## Some context first

There is no counterpart to this in inertia-laravel. Laravel has no mechanism to derive prop types from the server; you hand-write `interface Props` on the frontend. The same bug exists there — it's just not caught by types at all.

So "match the original" isn't available to us here. We have to decide what to prioritize ourselves. My two criteria:

1. The type must not lie (`props.errors.email` compiling and then crashing at runtime has to go).
2. The ergonomics of the code users actually write must not regress.

## Your concern about `Partial` was right

> Especially when there are two or more props, I feel this is undesirable.

I measured this. Here's a page rendered by one handler with three props and by another with none:

```ts
.get(
  '/three',
  (c) => c.render('Three'),
  (c) => c.render('Three', { a: 0, b: 'x', c: true })
)
```

The `Partial` version resolves to `{ a?: number; b?: string; c?: boolean }`. But in reality, if `a` is there, `b` and `c` are there too — the same `c.render` call passes all three at once. `Partial` destroys that correlation:

```ts
if (props.a !== undefined) {
  props.b // still string | undefined
}
```

Checking `a` doesn't settle `b`, so users end up writing `?.` that can never actually be undefined. The same happens with discriminated unions: after narrowing on `props.kind === 'ng'`, `props.error` stays `string | undefined`.

## On the union pattern

It's the more accurate type, but it breaks destructuring, which is what concerns me:

```tsx
function Signup({ errors }: PageProps<'Signup'>) {}
//               ~~~~~~ TS2339: Property 'errors' does not exist on type '{ errors: {...} } | {}'
```

Practically every Inertia page component is written this way, including the examples in the official docs. If this starts requiring `'errors' in props`, we end up with a type that is correct but that nobody actually uses.

## Proposal: fill the empty member with optional `never` instead of `{}`

I think the root problem isn't "the value may be undefined" — it's that the union members don't share the same set of keys. `{}` has no `errors` key at all, so TypeScript refuses the access.

So instead of collapsing the empty member to `{}`, we could give it the same keys as the other members, made optional and typed `never`:

```ts
type AllKeys<T> = T extends unknown ? keyof T : never

type NormalizeProps<T, All = T> = T extends Record<string, never>
  ? { [K in AllKeys<Exclude<All, Record<string, never>>>]?: never }
  : T
```

Which yields:

```ts
{ errors: { email: string } } | { errors?: never }
```

- The keys line up, so `props.errors` is accessible and destructuring works.
- The value can still be undefined, so `props.errors?.email` is enforced.
- The non-empty members are left completely untouched, so required stays required and narrowing keeps working.

That third point is the decisive difference from `Partial`. `Partial` rewrites `{ kind: 'ok' }` into `{ kind?: 'ok' }` and loses discrimination; this only inflates the empty member.

## Measured comparison

| | Before | Partial | Union | Proposal |
|---|---|---|---|---|
| Nonexistent prop is an error | ❌ | ✅ | ✅ | ✅ |
| `props.errors?.email` | type lies | ✅ | ❌ | ✅ |
| `function Page({ errors })` | ✅ | ✅ | ❌ | ✅ |
| With 3 props, checking one settles the rest | — | ❌ | ✅ | ✅ |
| `error: string` after `kind === 'ng'` | — | ❌ | ✅ | ✅ |

I also verified:

- Page rendered without props only → `{}` (since `AllKeys` resolves to `never`), which is what this PR intends.
- Props containing `defer` → `{ items: number[] } | { items?: never }`. Normalization runs after prop resolution, so there's no interference.
- Existing unions with no empty member (`{a} | {b}`) → the condition doesn't match, so behavior is unchanged.

## The trade-off

There's one way this loses to `Partial`: the type shown in the editor gets longer.

```ts
// Partial
{ errors?: { email: string } }

// Proposal
{ errors: { email: string } } | { errors?: never }
```

`?: never` isn't self-explanatory on first sight, so the implementation should carry a comment. But that's a one-time cost to understand, whereas the `?.` noise is something users pay on every line. I'd rather prioritize how the code reads over how the type displays.

## Other notes

- Extracting `MethodOutput`, introducing `Simplify`, and distributing over `C extends unknown` all look good to me. The distribution in particular is the right move — it makes bare `PageProps` normalize per page.
- Very minor: `.changeset/fast-kings-reply.md` is missing a trailing newline.

## Summary

I'm on board with the direction. For the implementation I'd like to push for the optional `never` approach above rather than `Partial` or the union — it keeps both the ergonomics and the accuracy, and it leaves the existing union behavior alone.

nkfr26:
@ashunar0 
Thank you for the review!
I think it's a wonderful way to compensate for each other's shortcomings! I adopted it as-is. Since it only affects empty objects, and the rule is consistent between the no-props case and the mixed case, I removed one test.
The description has also been updated. Please take a look when you have a chance!

ashunar0:
@nkfr26

Thanks for the update! I checked out the branch and ran typecheck / lint / tests, plus a few cases that aren't in the test file, and everything behaved as expected. 🎉

### One thing: the removed test

I'd like to ask for `keeps non-empty props from multiple handlers` (`{ a: number } | { b: string }`) to come back.

I understand the reasoning — the rule is now consistent, so the case felt redundant. But `NormalizeProps` is a conditional type with two branches, and right now only one of them is covered:

- true branch (fill the empty member with the other keys) → covered by the `UnionProps` test
- false branch (leave existing members alone) → no longer covered

The false branch is exactly the "don't change how existing unions behave" property that this PR must not break, so I'd like a test guarding it.

Something like this:

```ts
  .get(
    '/plain-union',
    (c) => c.render('PlainUnion', { a: 0 }),
    (c) => c.render('PlainUnion', { b: 'string' })
  )
```

```ts
  it('keeps unions without an empty member untouched', () => {
    expectTypeOf<PageProps<'PlainUnion'>>().toEqualTypeOf<{ a: number } | { b: string }>()
  })
```

If I've misread your intent behind the removal, let me know!

Other than that this looks good to me. LGTM 👍

nkfr26:
@ashunar0 
Sorry for the late reply.

I don't think it's quite accurate to say that only one branch is covered. The existing `UnionWithEmptyProps` test includes both empty and non-empty props, so it already covers both branches through a mixed union.
That said, I added the test back because it makes the case with no empty props explicit. I would have kept this test if it had existed before the fix. I also updated the component names and test descriptions to make the two cases match.

Please have a look!

yusukebe (approved):

yusukebe:
@nkfr26 @ashunar0 

Thanks! Looks good. I'll merge this now.

PR #2109, opened 2026-08-29
fix(inertia): return 303 for redirects from PUT/PATCH/DELETE

ashunar0:
## Summary

Redirects issued from `PUT`, `PATCH`, and `DELETE` handlers are returned as `302`. Per the [Fetch spec](https://fetch.spec.whatwg.org/#http-redirect-fetch), a `302` only downgrades `POST` to `GET` — for the other methods the client replays the original method against the redirect target. So `c.redirect('/users')` inside `app.put('/users/:id')` makes the Inertia client issue `PUT /users` instead of `GET /users`.

The Inertia protocol [requires `303` here](https://inertiajs.com/redirects), and `inertia-laravel` performs the same rewrite in its middleware.

Reproduced with a plain Node server and `fetch`, no Inertia involved:

```
PUT  → 302:  PUT /update-302   →  PUT /target    💥
PUT  → 303:  PUT /update-303   →  GET /target    ✅
POST → 302:  POST /update-302  →  GET /target    (already correct)
```

## Protocol behaviour

| Request | Handler returns | Before | After |
| --- | --- | --- | --- |
| `PUT` / `PATCH` / `DELETE` + `X-Inertia` | `302` | `302` | **`303`** |
| `POST` + `X-Inertia` | `302` | `302` | `302` |
| any method, no `X-Inertia` | `302` | `302` | `302` |
| `PUT` + `X-Inertia` | `307` / `308` | unchanged | unchanged |
| `PUT` + `X-Inertia` | `c.render(...)` | `200` | `200` |

## Implementation

The middleware now awaits `next()` and rewrites the status on the way out.

| Scope | |
| --- | --- |
| `302` → `303` for `PUT` / `PATCH` / `DELETE` on Inertia requests | ✅ |
| Non-Inertia requests left untouched | ✅ |
| `POST` left untouched (a `302` already becomes a `GET`) | ✅ |
| Other statuses (`301`, `307`, `308`, `200`) left untouched | ✅ |
| `Vary` on responses that don't go through `c.render` | ❌ separate gap, follow-up |

The gating matches `inertia-laravel`'s `Middleware::handle`: the Inertia header must be present, and only `302` is rewritten.

`return next()` became `await next()` + `return c.res`. The explicit return keeps `noImplicitReturns` satisfied now that the version-mismatch branch is no longer the only path returning a value.

## Test plan

- [x] `302` → `303` for `PUT`, `PATCH`, `DELETE`
- [x] `POST` stays `302`
- [x] non-Inertia `PUT` stays `302`
- [x] `307` stays `307`
- [x] a rendered `PUT` response stays `200`
- [x] full suite (71 tests), typecheck, lint, format

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/tidy-pugs-repeat.md
+++ b/.changeset/tidy-pugs-repeat.md
@@ -1,0 +1,5 @@
+---
+'@hono/inertia': patch
+---
+
+Rewrite `302` redirects to `303` for `PUT`, `PATCH`, and `DELETE` Inertia requests, so the client follows them with a `GET` instead of replaying the original method against the redirect target.
--- a/packages/inertia/README.md
+++ b/packages/inertia/README.md
@@ -191,6 +191,21 @@ When the `Referer` header is unavailable (e.g. under a `no-referrer` Referrer-Po
 app.post('/users', (c) => c.render('Users/New', {}, { url: '/users/new' }))
 ```
 
+## Redirects
+
+Redirect the way you normally would in Hono:
+
+```ts
+app.put('/users/:id', async (c) => {
+  await updateUser(c.req.param('id'), await c.req.parseBody())
+  return c.redirect('/users')
+})
+```
+
+The Inertia protocol requires redirects issued from `PUT`, `PATCH`, and `DELETE` requests to use [`303 See Other`](https://inertiajs.com/redirects): with a `302` the client replays the original method against the redirect target instead of following it with a `GET`. The middleware rewrites `302` to `303` for those methods on Inertia requests, so `c.redirect` needs no special casing.
+
+Non-Inertia requests, `POST` redirects, and explicit statuses such as `307` are left untouched.
+
 ## Partial reloads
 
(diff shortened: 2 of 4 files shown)


yusukebe:
@ashunar0 Thanks!

Hey @nkfr26 Can you review this? It seems good to me.

nkfr26:
@ashunar0 

I checked the inertia-laravel implementation and confirmed that it handles this the same way. LGTM!

By the way, separate from this issue, there are two other cases around this logic that are handled in inertia-laravel but not in @hono/inertia:

https://github.com/inertiajs/inertia-laravel/blob/3.x/src/Middleware.php#L154-L179

```php
if ($response->isOk() && empty($response->getContent())) {
    $response = $this->onEmptyResponse($request, $response);
}
```

```php
if ($isRedirect && $this->redirectHasFragment($response) && ! $request->prefetch()) {
    $response = $this->onRedirectWithFragment($request, $response);
}
```

The first one seems to serve as a fallback for empty responses, while the second handles redirects containing URL fragments via an Inertia location response.

When implementing these cases, I think it would be better to change the `X-Inertia` header check to an early return, as inertia-laravel does.

nkfr26 (approved):

yusukebe (approved):

yusukebe:
@nkfr26 Thanks for reviewing!

> By the way, separate from this issue, there are two other cases around this logic that are handled in inertia-laravel but not in @hono/inertia:

For this issue, can one of you create another PR?

Anyway, I'll merge this immediately. Thanks @ashunar0

PR #2111, opened 2026-08-29
perf(zod): upgrade to Zod 4.5 and optimize header validation

T4ko0522:
## Summary

- upgrade Zod 4 support to Zod 4.5.2
- require Zod `^4.5.2` in `@hono/zod-openapi`
- require Zod `^3.25.0 || ^4.5.2` in `@hono/zod-validator`
- cache the case-insensitive header key mapping when creating validator middleware instead of rebuilding it for every request
- add regression tests for both Zod 3 and Zod 4

Zod 4.5.2 was released less than 24 hours ago and is currently being rejected by the repository's supply chain policy check, so this PR remains in draft status.

Once 24 hours have passed since the release, the following steps will be taken:

- Regenerate `pnpm-lock.yaml`
- Rerun package tests and policy checks
- Mark this PR as ready for review

## Benchmark

Benchmarks use 20-field schemas and report median throughput.

### Header validation

The end-to-end Hono benchmark uses mixed-case schema keys and lowercase request headers.

| Implementation | Throughput |
| --- | ---: |
| Before, Zod 4.5.2 | 78,743 ops/s |
| After, Zod 4.5.2 | 104,317 ops/s |
| Improvement | **+32.5%** |

The improvement comes from building the case-insensitive key mapping once when the middleware is created instead of inspecting the schema on every request.

### Invalid query validation

A separate-process paired benchmark was used to check whether moving the header metadata initialization affected non-header validation.

| Implementation | Throughput |
| --- | ---: |
| Before, Zod 4.5.2 | 65,231 ops/s |
| After, Zod 4.5.2 | 64,970 ops/s |
| Difference | **-0.4%** |

### Schema creation and retained memory

The following results compare Zod 4.4.3 and Zod 4.5.2 while creating 1,000 object schemas with 20 fields.

| Metric | Zod 4.4.3 | Zod 4.5.2 | Difference |
| --- | ---: | ---: | ---: |
| Schema creation | 1,676 schemas/s | 5,709 schemas/s | **3.41x faster** |
| Retained heap per schema | 326,820 bytes | 61,198 bytes | **81.3% less** |

- [x] Add tests
- [x] Run tests
- [ ] `pnpm changeset` at the top of this repo and push the changeset

### DIFF
--- a/.changeset/zod-performance.md
+++ b/.changeset/zod-performance.md
@@ -1,0 +1,7 @@
+---
+'@hono/zod-openapi': minor
+'@hono/zod-validator': minor
+---
+
+Updated Zod 4 support to version 4.5.  
+Cached case-insensitive header schema metadata when creating validator middleware.
--- a/packages/zod-openapi/package.json
+++ b/packages/zod-openapi/package.json
@@ -41,7 +41,7 @@
   "homepage": "https://github.com/honojs/middleware",
   "peerDependencies": {
     "hono": ">=4.10.0",
-    "zod": "^4.0.0"
+    "zod": "^4.5.2"
   },
   "devDependencies": {
     "hono": "^4.11.5",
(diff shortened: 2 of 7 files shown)


T4ko0522:
I closed this because I thought it would be better to handle it as separate PRs.

PR #2112, opened 2026-08-29
perf(zod-validator): optimize header validation

T4ko0522:
## Summary

- cache the case-insensitive header key mapping when creating validator middleware instead of rebuilding it for every request
- avoid repeatedly inspecting the header schema shape during validation
- add regression tests for both Zod 3 and Zod 4

## Benchmark

Benchmarks use Zod 4.4.3 with 20-field schemas and report median throughput.

The before and after implementations were measured in separate Node.js processes across 10 paired runs, with five samples per process.

### Header validation

The end-to-end Hono benchmark uses mixed-case schema keys and lowercase request headers.

| Implementation |    Throughput |
| -------------- | ------------: |
| Before         |  86,699 ops/s |
| After          | 123,357 ops/s |
| Improvement    |    **+42.3%** |

The improvement comes from building the case-insensitive key mapping once when the middleware is created instead of inspecting the schema and rebuilding the mapping for every request.

The 95% confidence interval was `+35.2%` to `+48.6%`.

### Invalid query validation

A separate-process paired benchmark was used as a control to check whether moving the header metadata initialization affected non-header validation.

| Implementation |   Throughput |
| -------------- | -----------: |
| Before         | 38,695 ops/s |
| After          | 39,704 ops/s |
| Difference     |    **+2.6%** |

The 95% confidence interval was `-1.8%` to `+7.8%`, indicating no meaningful performance change outside the header-validation path.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.changeset/spicy-weeks-know.md
+++ b/.changeset/spicy-weeks-know.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-validator': patch
+---
+
+Cached case-insensitive header schema metadata when creating validator middleware.
--- a/packages/zod-validator/src/index.ts
+++ b/packages/zod-validator/src/index.ts
@@ -136,20 +136,24 @@ function zValidatorFunction<
 ):
   | MiddlewareHandler<E, P, V, ZodValidatorFailureResponse<T>>
   | MiddlewareHandler<E, P, V, ExtractValidationResponse<HookFn>> {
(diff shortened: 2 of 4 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@T4ko0522  Good approach! Looks good. Thanks!

PR #2115, opened 2026-08-31
perf(zod-openapi): Flatten batched route schemas

scttcper:
`openapiRoutes()` currently builds its RPC schema by recursively peeling the route tuple. The expensive part comes from this type shape, shortened to show the recursion:

```ts
type SchemaFromRoutes<Routes, BasePath> =
  Routes extends readonly [infer Head, ...infer Tail]
    ? SchemaFromRoute<Head, BasePath> &
        SchemaFromRoutes<Tail, BasePath>
    : {}
```

That produces a right-nested intersection like `Route1Schema & (Route2Schema & (Route3Schema & ...))`. This change projects every route independently, intersects the resulting union once, and normalizes only the top-level path map:

```ts
type SchemaFromRoutes<Routes, BasePath> =
  number extends Routes['length']
    ? {}
    : Routes extends readonly []
      ? {}
      : NormalizeSchema<
          UnionToIntersection<SchemaFromRoute<Routes[number], BasePath>>
        >
```

## TypeScript 6 and 7 benchmarks

These use consumers of each checkout's built declarations. Check times are medians from nine interleaved fresh compiler processes on Node 26.8.1, using TypeScript 6.0.2 and 7.0.2.

### TypeScript 6

| Routes | Before inst. | After inst. | Before check | After check |
|---:|---:|---:|---:|---:|
| 10 | 319,281 | 312,435 | 340 ms | 340 ms |
| 25 | 403,360 | 381,932 | 410 ms | 390 ms |
| 100 | TS2589 | 731,507 | TS2589 | 690 ms |

### TypeScript 7

| Routes | Before inst. | After inst. | Before check | After check |
|---:|---:|---:|---:|---:|
| 10 | 307,726 | 300,880 | 136 ms | 129 ms |
| 25 | 391,805 | 370,377 | 157 ms | 148 ms |
| 100 | TS2589 | 719,952 | TS2589 | 259 ms |

At 25 routes this removes 5.3-5.5% of compiler instantiations and improves median check time by 4.9-5.7%. The 100-route cases fail with TS2589 before the change and complete successfully after it on both compiler generations.

### DIFF
--- a/.changeset/flat-routes-scale.md
+++ b/.changeset/flat-routes-scale.md
@@ -1,0 +1,5 @@
+---
+'@hono/zod-openapi': patch
+---
+
+Flatten batched route type inference so large `openapiRoutes` tuples do not hit TypeScript's recursive instantiation limit.
--- a/packages/zod-openapi/eslint-suppressions.json
+++ b/packages/zod-openapi/eslint-suppressions.json
@@ -26,7 +26,7 @@
       "count": 1
     },
     "@typescript-eslint/no-unsafe-member-access": {
-      "count": 4
+      "count": 7
     },
     "@typescript-eslint/no-unsafe-return": {
       "count": 1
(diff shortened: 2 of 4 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@scttcper 

This is also good! Nice job.

PR #2117, opened 2026-09-01
fix(inertia): send redirects with a URL fragment as an Inertia location

ashunar0:
## Summary

Implements the two cases from `inertia-laravel`'s `Middleware::handle` that `@hono/inertia` was missing. Follow-up to @nkfr26's review on #2109.

**① Redirects containing a URL fragment**

A browser follows a redirect without ever sending the fragment to the server, so the `#profile` in `c.redirect('/users#profile')` is lost. `inertia-laravel` detects this and returns a `409` with `X-Inertia-Redirect`, letting the client navigate in JavaScript, where the fragment survives.

Prefetch requests are exempt: a prefetch only warms the cache for a later visit and must not navigate on the spot.

**② `Vary: X-Inertia` missing on most responses**

The same URL yields a JSON page object for Inertia requests and a full HTML document otherwise, so every response varies on `X-Inertia`. Today the header is only set on responses that went through `c.render` — redirects and errors never reach the renderer. A proxy caching one of those could serve an Inertia response to a plain browser navigation.

`inertia-laravel` sets it right after `$next($request)`, before the early return, on every response. This is the gap I noted as "separate gap, follow-up" in #2109.

## Protocol behaviour

| Request | Handler returns | Before | After |
| --- | --- | --- | --- |
| with `X-Inertia` | `302` + `Location: /users#profile` | `302` | **`409` + `X-Inertia-Redirect`** |
| with `X-Inertia` + `Purpose: prefetch` | same | `302` | `302` |
| without `X-Inertia` | same | `302` | `302` |
| with `X-Inertia` | `302` + `Location: /users` | `302` | `302` |
| any | redirect / error | no `Vary` | **`Vary: X-Inertia`** |
| any | `c.render(...)` | `Vary: Accept, X-Inertia` | unchanged |

The statuses considered are `201` / `301` / `302` / `303` / `307` / `308`, matching Symfony's `Response::isRedirect()`, which is what `inertia-laravel` gates the fragment check on.

Prefetch detection mirrors Laravel's `Request::prefetch()`: `X-Purpose: preview`, `Purpose: prefetch`, or `Sec-Purpose: prefetch`.

## Implementation

As @nkfr26 suggested, the `X-Inertia` header check became an early return, so additional branches don't nest. The order now matches `inertia-laravel`:

```
await next()
  ↓ Vary on every response
  ↓ no X-Inertia → early return
  ↓ 302 → 303 (PUT/PATCH/DELETE)   ← #2109
  ↓ redirect with a fragment → 409
```

As a result the `302` → `303` condition no longer needs `c.req.header('X-Inertia') &&`.

The `409` is built by assigning `c.res` rather than returning it: `compose` ignores a middleware's return value once the response is finalized (the same reason the `303` in #2109 was an assignment). The `c.res` setter also copies the previous headers onto the new response, so `Location` is deleted beforehand — `Vary`, on the other hand, is carried over by that same mechanism.

The `Vary` helper does not clobber the `Accept, X-Inertia` the renderer already set: it appends `X-Inertia` only when absent. The version-mismatch `409` returns before `next()`, so it sets the header on its own.

## Out of scope

The other case @nkfr26 raised — the empty-response fallback:

```php
if ($response->isOk() && empty($response->getContent())) {
    $response = $this->onEmptyResponse($request, $response);   // Redirect::back()
}
```

I left this out. `Redirect::back()` resolves the previous URL from the session, with `Referer` only as a fallback. `@hono/inertia` has no session, so it would have to rely on `Referer` alone — which does not match Laravel's behaviour and puts a user-controlled value straight into `Location`. I'd rather track it as a separate issue.

## Test plan

- [x] redirect with a fragment → `409` + `X-Inertia-Redirect`, `Location` dropped, `Vary` kept
- [x] applies after the `303` rewrite (`PUT` + fragment → `409`)
- [x] fragment-less redirects untouched
- [x] non-Inertia requests untouched
- [x] all three prefetch headers (`Purpose` / `Sec-Purpose` / `X-Purpose`) untouched
- [x] `Vary: X-Inertia` on redirects and errors
- [x] the renderer's `Accept, X-Inertia` preserved
- [x] `Vary` on the version-mismatch `409`
- [x] full suite (81 tests), typecheck, lint, format

- [x] Add tests
- [x] Run tests

### DIFF
--- a/packages/inertia/src/index.ts
+++ b/packages/inertia/src/index.ts
@@ -469,6 +469,7 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
       const current = version ?? ''
       if (requested !== current) {
         c.header('X-Inertia-Location', c.req.url)
+        c.header('Vary', 'X-Inertia')
         return c.body(null, 409)
       }
     }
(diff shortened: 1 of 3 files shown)


yusukebe:
@nkfr26 Can you review this?

packages/inertia/src/index.ts
```diff
@@ -657,22 +658,79 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
...
+  }
+  c.header('Vary', current ? `${current}, X-Inertia` : 'X-Inertia')
+}
+
+/**
+ * Whether the request is a prefetch. Mirrors Laravel's `Request::prefetch`,
+ * which `inertia-laravel` uses to skip the fragment redirect.
+ */
+const isPrefetch = (c: Context): boolean =>
+  c.req.header('X-Purpose')?.toLowerCase() === 'preview' ||
```
nkfr26, line 730: Isn't it `X-Moz` ? Laravel does not support `X-Purpose: preview` .
https://github.com/illuminate/http/blob/master/Request.php#L314-L324

packages/inertia/src/index.ts
```diff
@@ -469,6 +469,7 @@ export const inertia = (options: InertiaOptions = {}): MiddlewareHandler => {
       const current = version ?? ''
       if (requested !== current) {
         c.header('X-Inertia-Location', c.req.url)
+        c.header('Vary', 'X-Inertia')
         return c.body(null, 409)
       }
     }
```
nkfr26, line 475: Would it make sense to align the processing order with inertia-laravel while we're here? This would also make the Vary header handling consistent and improve maintainability.

nkfr26 (commented):
@ashunar0 
Thanks for the fix!
I don't see any major issues, but I left a few comments. I also wanted to comment on the point currently marked as out of scope.

Laravel's `Redirect::back()` first uses the `Referer` header and falls back to the previous URL stored in the session when the `Referer` is unavailable.
https://github.com/illuminate/routing/blob/master/UrlGenerator.php#L162-L175
However, using the `Referer` directly as the `Location` header has also been discussed as a potential security concern in Laravel.
https://github.com/laravel/framework/issues/14642
I think it would be better to leave this out of the current PR and consider addressing it separately.

PR #2122, opened 2026-09-03
chore: migrate to @cloudflare/vitest-plugin

BarryThePenguin:
Today I learned about [`@cloudflare/vitest-plugin`](https://developers.cloudflare.com/workers/testing/vitest-integration/migration-guides/migrate-to-vitest-plugin/)

### DIFF
--- a/packages/firebase-auth/package.json
+++ b/packages/firebase-auth/package.json
@@ -48,7 +48,7 @@
     "hono": ">=4.0.0"
   },
   "devDependencies": {
-    "@cloudflare/vitest-pool-workers": "^0.22.0",
+    "@cloudflare/vitest-plugin": "^1.0.0",
     "firebase-tools": "^15.26.0",
     "hono": "^4.11.5",
     "tsdown": "^0.22.3",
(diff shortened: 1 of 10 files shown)


packages/firebase-auth/package.json
```diff
@@ -48,7 +48,7 @@
     "hono": ">=4.0.0"
   },
   "devDependencies": {
-    "@cloudflare/vitest-pool-workers": "^0.22.0",
+    "@cloudflare/vitest-plugin": "^1.0.0",
```
yusukebe, line 51: Is it better to set it to the latest version, `1.1.4`?

https://www.npmjs.com/package/@cloudflare/vitest-plugin

BarryThePenguin: The lockfile is already using `1.1.3`. While `1.1.4` was only released today so it fails the default pnpm [`minimumReleaseAge`](https://pnpm.io/settings/dependency-resolution#minimumreleaseage) of 24 hours. Dependabot should update at somepoint

yusukebe: Ah, I see!

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin Thanks!

Issue #2110, opened 2026-08-29
Proposal: support shared data (always / share / page.sharedProps) in @hono/inertia


ashunar0:
## Summary

`@hono/inertia` has no **shared data** — props that are automatically included with every page. Today, things every page needs (`auth`, `flash`, `errors`) have to be passed by hand from every handler, so the boilerplate grows with the route count and a single omission silently breaks one page.

```ts
app.get('/dashboard', (c) => c.render('Dashboard', { stats, auth: getUser(c), flash: getFlash(c) }))
app.get('/users',     (c) => c.render('Users/Index', { users, auth: getUser(c), flash: getFlash(c) }))
```

Reading through `inertia-laravel`, this turns out not to be a single feature but a **set of three**, so I'd like to land them in order.

## Background

Three things stood out in `inertia-laravel`.

**1. Shared props are filtered by partial reloads like any other prop**

`PropsResolver::shouldInclude()` doesn't special-case shared props. If a shared prop doesn't match `only` / `except`, it's dropped.

```php
if (! $this->isPartial || $prop instanceof AlwaysProp || $parentWasResolved) {
    return true;
}
```

So it isn't "shared, therefore always present" — it's "wrapped in `always()`, therefore always present". **Without `always()`, shared data doesn't behave correctly.**

**2. Laravel shares `errors` by default**

```php
public function share(Request $request)
{
    return ['errors' => Inertia::always($this->resolveValidationErrors($request))];
}
```

`errors` being present on every page is a consequence of shared + `always()`. Since `@hono/inertia` has neither, users pass `errors` per handler instead — and a handler that forgets makes `PageProps` resolve to `{ errors: X } | {}` (the case discussed in #2103).

**3. `sharedProps` is emitted on the page object, so it's part of the protocol**

```php
protected function buildMetadata(): array
{
    return array_filter([
        'sharedProps'   => $this->sharedPropKeys,
        'mergeProps'    => $this->mergeProps,
        // ...
    ], fn ($v) => count($v) > 0);
}
```

The client is told which keys are shared. That makes merging two objects server-side insufficient for conformance.

## Proposed plan

These depend on each other, so I'd send them as separate PRs as usual.

```
① always()                          ← foundation, useful on its own
   └─ ② share() + page.sharedProps   ← depends on ①
         └─ ③ SharedProps type registry  ← depends on ②
```

**① `always()`**

[...]

PR #2125, opened 2026-09-05
feat(inertia): add always() for props that bypass partial reload filtering

ashunar0:
## Overview

Adds `always()`, a prop wrapper that bypasses partial reload filtering. This is item ① from #2110, and the foundation the shared data work (②) builds on.

## Motivation

Partial reloads narrow the response down to the props listed in `X-Inertia-Partial-Data` (or everything except the ones in `X-Inertia-Partial-Except`). Cross-cutting props — validation errors, flash messages, the authenticated user — are still needed by the page even when the client asked for a single prop, and there is currently no way to express that.

```ts
app.get('/users', (c) =>
  c.render('Users/Index', {
    users: () => db.users.all(),
    errors: always(getErrors(c)), // present even on a partial reload of `users`
  })
)
```

## Parity with inertia-laravel

`AlwaysProp` short circuits on the first condition of `shouldIncludeInPartialResponse`, so it bypasses **both** the `only` and the `except` header. This PR matches that.

```php
if (! $this->isPartial || $prop instanceof AlwaysProp || $parentWasResolved) {
    return true;
}
```

There is no `always` entry in `buildMetadata()`, so nothing is added to the page object.

## Implementation

Same marker approach as `defer()` / `merge()` / `scroll()`. The renderer unwraps the marker at the top of the prop loop, skips the filter, and lets the inner value fall through to the existing resolution path. As a result `always(() => ...)` still resolves lazily and `always(merge(...))` still emits its merge metadata. Wrapping a plain value keeps the synchronous fast path intact.

`always(defer(...))` is not supported: `AlwaysProp` is not `Deferrable` in inertia-laravel either, and "always present on the initial response" and "never present on the initial response" cannot both hold. This is documented in the README.

## Related

Item ① of #2110. Item ② (`share`) is being worked on by @nkfr26 in #2124 — shared data needs this to behave correctly under partial reloads, which is exactly why Laravel wraps `errors` in `always()`.

- [x] Add tests
- [x] Run tests
- [x] Follow [the contribution guide](https://github.com/hono/middleware?tab=readme-ov-file#how-to-contribute)

### DIFF
--- a/.changeset/tall-onions-repeat.md
+++ b/.changeset/tall-onions-repeat.md
@@ -1,0 +1,5 @@
+---
+'@hono/inertia': minor
+---
+
+Add `always()` to mark a prop as always included. Props wrapped with it bypass both the `X-Inertia-Partial-Data` and `X-Inertia-Partial-Except` filters, so cross-cutting props such as validation errors or flash messages survive a partial reload. Mirrors `Inertia::always()` in `inertia-laravel`.
--- a/packages/inertia/README.md
+++ b/packages/inertia/README.md
@@ -223,6 +223,25 @@ Function props (`() => T | Promise<T>`) are evaluated lazily, so heavy data fetc
 
 A request is treated as a partial reload only when `X-Inertia-Partial-Component` matches the rendered component **and** at least one of `X-Inertia-Partial-Data` / `X-Inertia-Partial-Except` is present. Otherwise the request is processed as a normal Inertia visit and every prop — including function props — is resolved.
 
+### Always included props
+
+Some props are needed by the page no matter what the client asked for — validation errors, flash messages, the authenticated user. Wrap them with `always()` and they bypass both `X-Inertia-Partial-Data` and `X-Inertia-Partial-Except`:
+
+```ts
+import { always } from '@hono/inertia'
+
+app.get('/users', (c) =>
+  c.render('Users/Index', {
+    users: () => db.users.all(),
+    errors: always(getErrors(c)), // present even on a partial reload of `users`
+  })
+)
+```
+
+The wrapped value follows the same rules as a plain prop, so `always(() => getFlash(c))` defers the work until the prop is rendered.
+
+`always()` is a filter escape hatch rather than a prop type — wrap plain values or functions with it. Combining it with `defer()` is not supported, since a deferred prop is by definition absent from the initial response.
+
 ## Deferred props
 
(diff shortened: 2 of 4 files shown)


nkfr26:
For direct compositions such as `always(merge(...))` and `always(defer(...))`, the current behavior appears to differ from `inertia-laravel` 3.x.

| Case | This PR | inertia-laravel |
| --- | --- | --- |
| `always(merge(...))` | `always` is unwrapped first, so the inner `merge` marker is processed and merge metadata is emitted | The outer `AlwaysProp` is used for filtering and metadata collection. Resolving it also resolves the inner `MergeProp` to its value, so merge metadata is not emitted |
| `always(defer(...))` | `always` is unwrapped first, so the inner `defer` marker is processed and the prop is deferred on the initial response | Initial exclusion sees the outer `AlwaysProp`, so the prop is not excluded. Resolving the `AlwaysProp` also resolves the inner `DeferProp`, so the value is included in the initial response |

I noticed this while looking at these parts of the PR description:

> `always(merge(...))` still emits its merge metadata.

> `always(defer(...))` is not supported: `AlwaysProp` is not `Deferrable` in inertia-laravel either.

In this PR, `const value = forced ? raw.data : raw` unwraps `always` before the existing prop-type checks, so the inner marker gets its normal behavior.

In `inertia-laravel`, the outer `AlwaysProp` remains the prop considered for partial filtering, initial exclusion, and metadata collection before resolution. Since resolving an `AlwaysProp` also resolves callable inner prop objects, direct compositions end up behaving differently. The same applies to `always(scroll(...))`.

Is this wrapper-composition behavior intentional? I'm wondering whether Hono wants to define these semantics independently, or whether direct compositions should behave closer to `inertia-laravel`.

If supporting nested prop-marker composition is out of scope for this PR, another option might be to reject it explicitly instead of implicitly applying the inner prop type's behavior.

For example:

```ts
export const always = <T>(data: T): T => {
  if (isDeferred(data) || isMerge(data) || isScroll(data)) {
    throw new TypeError('always() cannot wrap another prop type')
  }

  const marker: AlwaysProp<T> = {
    [ALWAYS_MARKER]: true,
    data,
  }
  return marker as unknown as T
}
```

Since `always(defer(...))` is already documented as unsupported, explicitly rejecting nested prop markers may make the API behavior clearer and avoid introducing composition semantics that differ from `inertia-laravel`.

PR #2138, opened 2026-09-11
feat(x-trust): add X-Trust human presence middleware

htl-syterme:
## What this adds
A new middleware package `@hono/x-trust` that validates the X-Trust header — an open standard proving human presence via HMAC-SHA256 signed token.

## Why
Hono-based APIs face bot abuse on free tiers. X-Trust annotates requests with a human score (0..1) before they reach handlers — zero KYC, zero PII, zero friction.

## Usage
```ts
import { xTrust } from '@hono/x-trust'

app.use('*', xTrust({ secret: process.env.HTL_SECRET, minScore: 0.5 }))

app.get('/', (c) => {
  const { trusted, score } = c.get('xTrust')
  return c.json({ trusted, score })
})


htl-syterme:
Happy to add a changeset — will push it shortly. This is a new package addition so it will be a minor bump for @hono/x-trust.

htl-syterme:
Also available on JSR for Deno/Node users:
`deno add jsr:@htl-syterme/htl-core`
https://jsr.io/@htl-syterme/htl-core

htl-syterme:
X-Trust differs from Cloudflare Turnstile fundamentally:

- Turnstile: one verification at entry, widget friction
- X-Trust: signs every HTTP request, 120s TTL, zero friction

Continuous human attestation per request — not a checkpoint.
This is the gap the IETF Web Bot Auth charter explicitly leaves out.

yusukebe:
This middleware isn't commonly requested, and users can implement it themselves without much effort. We should not add this to our official monorepo. Closing. Thanks.

